PR #8, opened 2017-12-08
Optimize avr port

Heiko-L:


feilipu:
I've found that [code generated with the decoration `__attribute__ (flatten)`](https://github.com/feilipu/miniAVRfreeRTOS/blob/master/port.c#L514) achieves the required outcome, without changing the standard calls.

I'd also note that the `port.c` file contained in the `/portable/GCC/ATMega323/` directory hasn't been updated since 2011, and gcc has changed many of its decorations, so things like `__attribute__ (signal)` are now depreciated for `__attribute__ (interrupt)`, for example.

So, there's lot's of other stuff that could also be fixed, if anyone is still interested in 8-bit AVRs. :smile:

Heiko-L:
> I've found that code generated with the decoration __attribute__ (flatten) achieves the required outcome, without changing the standard calls.

flatten does, like declaring a function inline, not guarantee that the function gets inlined.

> so things like __attribute__ (signal) are now depreciated for ISR_NAKED, for example.

from /avr/interrupt.h:
```
#if (__GNUC__ == 4 && __GNUC_MINOR__ >= 1) || (__GNUC__ > 4)
#  define __INTR_ATTRS used, externally_visible
#else /* GCC < 4.1 */
#  define __INTR_ATTRS used
#endif
#  define ISR(vector, ...)            \
    void vector (void) __attribute__ ((signal,__INTR_ATTRS)) __VA_ARGS__; \
    void vector (void)
#  define ISR_NAKED      __attribute__((naked))
```

Despite used and externally_visible it is the same ;) (I think they're useful, so using /avr/interrupt.h macros seems to be better and more readable.

Heiko-L:
@feilipu I saved another cycle for "scheduler preemption"- what do you think about that?

feilipu:
I only comment, because you asked. Generally, I don't think the level of interest is high enough to invest time into changes for ATmega at AWS FreeRTOS. 8 bit machines are not imho their focus. The interest here is 32 bit IOT connectivity to AWS services.  It is probably better use of your time to make improvement suggestions on Richard's freeRTOS.org web site.

Having said that, the mechanism you're using is damaging a reliable construct that assembly programmers are used to. Effectively it is 1. save the interrupt status (from `__SREG__`), 2. clear the interrupt enable, 3. do stuff, 4. restore previous interrupt status. You've damaged the construct by putting a `cli` BEFORE saving the `__SREG__` so you can never know what the interrupt status was before you saved it.

Have a look at the macros for [`util/atomic.h`](http://www.nongnu.org/avr-libc/user-manual/group__util__atomic.html) to see what the gurus do.

In the particular case here it doesn't matter, but the two macros are now incomplete and can't be used as a elsewhere anymore. This is because the interrupt can't be guaranteed to be off as the context is being written to the stack. That's a subtle bug waiting just to happen.

feilipu:
> In the particular case here it doesn't matter, but the two macros (`portSAVE_CONTEXT` and `portRESTORE_CONTEXT`) are now incomplete and can't be used elsewhere anymore. This is because the interrupt can't be guaranteed to be off as the context is being written to the stack. That's a subtle bug just waiting to happen.

Whilst the above is true in the situation where these two macros are only used locally, as is the case, there is no issue. In fact, the [`cli`](https://github.com/aws/amazon-freertos/blob/master/lib/FreeRTOS/portable/GCC/ATMega323/port.c#L85) doesn't need to be there at all, as [it is generated by the ISR preamble code](https://github.com/aws/amazon-freertos/blob/master/lib/FreeRTOS/portable/GCC/ATMega323/port.c#L78).

dcgaws:
Heiko-L, thank you for taking the time to prepare these commits. We don't at this time have the resources to test and support ATmega 8-bit changes in Amazon FreeRTOS.

PR #21, opened 2018-03-01
fix Amazon copyright statements

MarkAtwood:
The Amazon copyright statement was incorrect in a few places, including in the top level LICENSE file.


mradulan:
Thank you for posting this out, it will get fixed.

PR #28, opened 2018-05-14
[port][avr] Added kernel port for ATMega328P

shchers:
In scope of this patch was implemented port of
FreeRTOS kernel for AVR ATMega328P. Also, added
demo sample with basic LEDs blinking for Arduino
Nano board.

Signed-off-by: Sergey Shcherbakov <shchers@gmail.com>

### DIFF
--- a/demos/avr/arduino_nano/FreeRTOSConfig.h
+++ b/demos/avr/arduino_nano/FreeRTOSConfig.h
@@ -1,0 +1,54 @@
+/*
+ * Basic FreeRTOS configuration properties
+ * Author: Sergey Shcherbakov <shchers@gmail.com>
+ *
+ * See http://www.freertos.org/a00110.html.
+ */
+
+
+#ifndef FREERTOS_CONFIG_H
+#define FREERTOS_CONFIG_H
+
+#include <avr/io.h>
+
+/* Ensure stdint is only used by the compiler, and not the assembler. */
+#if defined( __ICCARM__ ) || defined( __ARMCC_VERSION ) || defined( __GNUC__ )
+    #include <stdint.h>
+    extern uint32_t SystemCoreClock;
+       extern int DbgConsole_Printf( const char *fmt_s, ... );
+#endif
+
+
+#define configSUPPORT_STATIC_ALLOCATION              0
+
+#define configUSE_PREEMPTION                         1
+#define configUSE_IDLE_HOOK                          0
+#define configUSE_TICK_HOOK                          0
+#define configUSE_TICKLESS_IDLE                      0
+#define configUSE_DAEMON_TASK_STARTUP_HOOK           0
+#define configCPU_CLOCK_HZ                           ( ( unsigned long ) 16000000 )
+#define configTICK_RATE_HZ                           ( ( TickType_t ) 1000 )
+#define configMAX_PRIORITIES                         ( 4 )
+#define configMINIMAL_STACK_SIZE                     ( ( unsigned short ) 128 )
+#define configTOTAL_HEAP_SIZE                        ( (size_t ) ( 512 ) )
+#define configMAX_TASK_NAME_LEN                      ( 8 )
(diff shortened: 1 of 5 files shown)


rzr:
What about 2560 ? shouldnt it be very similar ?

feilipu:
They are quite similar, but there are many additional features for the 2560.

If you're interested in comparing implementations for the two (and 1284p, and 16u2), this [file](https://github.com/feilipu/avrfreertos/blob/master/freeRTOS10xx/include/FreeRTOSBoardDefs.h) provides some overview. The [`port.c`](https://github.com/feilipu/avrfreertos/blob/master/freeRTOS10xx/portable/port.c) shows how the different register capabilities of the two devices should be handled.

Here's the repositories for [avrfreertos](https://github.com/feilipu/avrfreertos) and [Arduino FreeRTOS Library](https://github.com/feilipu/Arduino_FreeRTOS_Library).
Both of which cover the differences, and similarities across the two MCU.

IMHO the 2560 could potentially be integrated into amazon-freertos. But, there's no potential for the 328p to be integrated. There is not enough RAM in the 328p to support both an IP stack and a SD Card file system. Both would need an external Ethernet, and probably the W5500 would be most suited for this purpose.

shchers:
Hi @rzr ,
Sure, it's very similar and even more powerful, but it should be integrated as a different patch.

Hi @feilipu ,
2560 really good and even better than 328p, but why thinking that 328p don'y have potential to be integrated?
It's cost effective and powerful enough to collect data from sensors. 
IP and FAT can be used on 328p, of course with major limitations, anyway such limitations will hit 2560, because there's not so big difference in ROM/RAM.

Few more comments:
  - IP - a lot of modems have embedded TCP/IP and UDP stack, for radio-links you don't need IP at all.
  - SDCard not really need for most cases and even in case if tons of data need to be saved, you always can manage without FAT.

With best regards,
SErgey

feilipu:
> But why thinking that 328p don't have potential to be integrated? It's cost effective and powerful enough to collect data from sensors.

@shchers, yes, the 328p is sufficient for collecting data from sensors, but this AWS implementation requires substantial additional library support to inter-work with AWS services that is (IMHO) beyond the capabilities of the 328p. There are things like [encryption](https://github.com/aws/amazon-freertos/tree/master/lib/crypto), authentication, and [MQTT ](https://github.com/aws/amazon-freertos/tree/master/lib/mqtt)protcols that require RAM and processor resources that are not available on the 328p.

> IP and FAT can be used on 328p, of course with major limitations, anyway such limitations will hit 2560, because there's not so big difference in ROM/RAM.

The 2560 has 8kB of RAM, which is **4x** that of the 328p. It can also support [external static ram](https://feilipu.me/2013/02/17/ardusat-xramfs-prototyping) up to 512kB if needed (though this is unlikely to be needed). 8kB allows IPv4 packets (1300bytes) to be managed (generated and parsed) within the main memory. Unfortunately with only 2kB of main memory in total the 328p is substantially constrained in handling IP networking, which is the essence of this AWS variety of FreeRTOS.

FreeRTOS for the 328p is (IMHO) a great tool for learning about schedulers, and the language and functionality of a RTOS. But, it is not really a workable solution for integration into the AWS toolset. And just to note that all the above is just my opinion, and I have no say in what AWS do, or don't do.

RichardBarry:
Thanks for providing your input.  Is this code in a git repo anywhere?  If so we can add a link to your code from the FreeRTOS Interactive site (https://interactive.freertos.org) and then link to it from one of the older AVR pages on the FreeRTOS.org site.  Unfortunately the FreeRTOS interactive forum host changed their forum software recently and no longer allow the code to be attached directly.

shchers:
Hello @RichardBarry ,

My repo is here, on Github - https://github.com/shchers/amazon-freertos
branch: m328p

With best regards,
Sergey

RichardBarry:
Thanks - sorry forgot to close.

PR #49, opened 2018-08-28
Python tweaks for quick start

jrr:
A couple small tweaks to smooth out the quick start process:

- Add shebang line and chmod +x so that users can `./SetupAWS.py`
- Git ignore *.pyc


sarenameas:
Thank you for taking the time to make these changes. These will be fixed.

dcgaws:
jrr, you did the right thing vis-a-vis the PR. Thank you. We took the shebang change, but not the .gitignore one for now. Regarding the latter, we aren't currently including those files in our release process.

PR #61, opened 2018-09-20
Fix spelling mistake of comments

bnyf:
Delete a redundant ‘an’ in the comments about the function that creates a task.


WRansohoff:
Thanks for reporting this; we will put the change into an upcoming release.

PR #64, opened 2018-09-28
FreeRTOS_DNS: Repair LLMNR and FreeRTOS_gethostbyname_a function

htibosch:
With these changes, the non-blocking function FreeRTOS_gethostbyname_a() will work properly again.
A DNS ID is 16 bits, and not 32-bits ( see xIdentifier = ipconfigRAND32 in FreeRTOS_DNS ).
ulDNSHandlePacket() must be called for incoming DNS replies ( in FreeRTOS_UDP_IP ).
And also LLMNR requests will be responded to correctly.

Signed-off-by: Hein Tibosch <hein@htibosch.net>

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
+++ b/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
@@ -789,7 +789,6 @@ BaseType_t xIsCallingFromIPTask( void );
 typedef struct xSOCKET_SET
 {
 	EventGroupHandle_t xSelectGroup;
-	BaseType_t bApiCalled;	/* True if the API was calling  the private vSocketSelect */
 	FreeRTOS_Socket_t *pxSocket;
 } SocketSelect_t;
 
--- a/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
+++ b/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
@@ -789,7 +789,6 @@ BaseType_t xIsCallingFromIPTask( void );
 typedef struct xSOCKET_SET
 {
 	EventGroupHandle_t xSelectGroup;
-	FreeRTOS_Socket_t *pxSocket;
 } SocketSelect_t;
 
 extern void vSocketSelect( SocketSelect_t *pxSocketSelect );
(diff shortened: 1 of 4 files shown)


dcgaws:
Thank you for proposing these patches; we're reviewing.

htibosch:
Hm, I think that I should have created a new branch for the above commit?
'SocketSelect_t' has nothing to do with the earlier subject DNS.

htibosch:
Remove another unused field 'pxSocket' from the type 'SocketSelect_t':
I think that both 'bApiCalled' and 'pxSocket' are remains from earlier testing.
All that remains in the declaration is:

`typedef struct xSOCKET_SET
{
	EventGroupHandle_t xSelectGroup;
} SocketSelect_t;
`

htibosch:
This test in ulDNSHandlePacket() :

`if( pxNetworkBuffer->xDataLength > sizeof( UDPPacket_t ) )`

is misplaced.

In the function "prvProcessIPPacket()", it was proven that the received UDP-packet has 3 headers and that it has a valid UDP payload.

The length of the 3 headers ( Ethernet, IP, and UDP ) were already subtracted:

`pxNetworkBuffer->xDataLength -= sizeof( UDPPacket_t );`

The only test that should remain here is this one:

`	if ( xPlayloadBufferLength < sizeof( DNSMessage_t ) )
	{
		return pdFAIL;
	}
`

htibosch:
At this point, there are only 2 changes left.
I'll describe them here in C comments:

Change-1:
	/* DNS identifiers are 16-bit. */
	xIdentifier = ( TickType_t )( ipconfigRAND32( ) & 0xffff );
	// As for the lookup call-back routine, a matching will take place
	// between the identifier received and the identifiers stored.
	// An "identifier received" is always 16-bits.
	// As long as 32-bit identifiers are issued here, a match is never found
	// and as a result of that, "FreeRTOS_gethostbyname_a()" can not succeed.

Change-2:
	// The following test is misplaced: comparing a payload buffer
	// length with the size of a UDS packet of 42 bytes long:
	// This test will never pass, unless it's an extremely long
	// DNS/LLMNR packet:
	if( pxNetworkBuffer->xDataLength > sizeof( UDPPacket_t ) )
	// "sizeof( UDPPacket_t )" has already been subtracted
	// from xDataLength
	{
		prvParseDNSReply( pucUDPPayloadBuffer,
			xPlayloadBufferLength,
			( uint32_t )pxDNSMessageHeader->usIdentifier );
	}

I'm sorry, but at this moment, I can not find the time to write tests for these 2 changes.
Yes they are essential to get "FreeRTOS_gethostbyname_a()" working again.

Thanks, Hein

htibosch:
Closing this pull-request for now.
I will open a new clean pull request for "Repairing the DNS asynchronous lookup"

PR #76, opened 2018-10-31
Fixed error in return type of xTaskCreateStatic

chanibal:
The xTaskCreateStatic claimed that it return pdPASS or an error value depending on success, but what they it returns is an TaskHandle_t.

This error is also in official FreeRTOS.


ravibhagavandas:
Thanks for reporting. We will fix this in upcoming releases.

sarenameas:
This has been completed: https://github.com/aws/amazon-freertos/blob/master/lib/include/task.h#L386

PR #79, opened 2018-11-06
Implement Access Point mode for ESP32 MCUs

morganatkinspa:
Implemented missing functions for third party ESP drivers.

### DIFF
--- a/lib/third_party/mcu_vendor/espressif/esp-idf/components/esp32/include/esp_event.h
+++ b/lib/third_party/mcu_vendor/espressif/esp-idf/components/esp32/include/esp_event.h
@@ -44,6 +44,7 @@ typedef enum {
     SYSTEM_EVENT_AP_STOP,                  /**< ESP32 soft-AP stop */
     SYSTEM_EVENT_AP_STACONNECTED,          /**< a station connected to ESP32 soft-AP */
     SYSTEM_EVENT_AP_STADISCONNECTED,       /**< a station disconnected from ESP32 soft-AP */
+    SYSTEM_EVENT_AP_STAIPASSIGNED,         /**< ESP32 soft-AP assign an IP to a connected station */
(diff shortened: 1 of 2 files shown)


alexa-noxon:
Thanks for sharing your ESP32 WiFi function implementations.   The Amazon FreeRTOS team is currently in the process of transitioning to using Github as a primary development repository and unfortunately, there is still a little work to be done in determining the internal procedure for accepting public contributions.

yeti7:
@morganatkinspa,
Thanks for the mods. Were you able to also get DHCP running?

morganatkinspa:
@yeti7 Kind of...
We imported the LWIP stack (which I don't think is supported) to run only DHCP and web server. This was for the initial setup of the device, once the device was configured it rebooted - dropping the LWIP stack to just using the provided IP stack.
It wasn't the easiest thing to get working...

yeti7:
@morganatkinspa,
LWIP is where I ended up looking too. Is your solution available on somewhere github or can be shared?

morganatkinspa:
@yeti7 Unfortantly we can't plublish that code, it's quite tied into the rest of our codebase.
Best of luck!

qiutongs:
We apologize that it takes so long to provide feedback. Our code base has changed a lot and now we accept external Pull Request.  I am sorry that I will close this Pull Request but you are welcomed to open a new one if needed.

For your reference, here are latest code:
https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/esp-idf/components/esp32/include/esp_event.h#L47

https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c

ashwiniverma:
@morganatkinspa @qiutongs  I am trying to implement an AP for Esp32 DevKitC. The AP is successfully created but when I try to connect by laptop to the created AP , it get a max connection deauth error (log: wifi: max connection, deauth!). 

I know this might not be the right place for this question, but just wanted to bring that up here too.

PR #92, opened 2018-11-15
fix: Use symbolic links for git hooks

tgsong:
Description
-----------

Git hooks were set up with hard link before. However, git checkout will remove local file and replace it with a new file, thus breaking the hard link. We should use symbolic links instead.

Checklist:
----------

- [ ] ~~I have tested my changes. No regression in existing tests.~~
- [ ] ~~My code is Linted.~~


tools/git/hooks/install_hooks.ps1
```diff
@@ -1,2 +1,3 @@
-New-Item -ItemType HardLink -Name ../../../.git/hooks/pre-commit -Value src/pre_commit.py
-New-Item -ItemType HardLink -Name ../../../.git/hooks/commit-msg -Value src/commit_msg.py
+Remove-Item -Recurse -Force ..\..\..\.git\hooks\*
```
alexa-noxon: Would we want to consider only deleting the 2 files instead of the whole folder?  Is there any chance someone will have other stuff hanging out in there that they might want to keep?

Issue #93, opened 2018-11-15
Missing include in aws_logging_task.h


Hadatko:
This header file is using data types which are not defined. Missing include file. Likely portmacro.h
https://github.com/aws/amazon-freertos/blob/master/demos/common/include/aws_logging_task.h

RichardBarry:
Thank you for taking the time to point this out.  In this case it is intended that FreeRTOS.h be included in your C file before aws_logging_task.h.  It is our preference to avoid including one header from another if possible - for example if you want to use a FreeRTOS kernel queue API then you would include FreeRTOS.h first, and queue.h (the header file that has contains the prototypes of the kernel's queue API) second.

In the case you are reporting here it is noted that a #error message in the header file informing of the need to include FreeRTOS.h first will make the code more user friendly so I will leave this issue open for now.

Just as general information - portmacro.h and FreeRTOSConfig.h should never be included in a C file directly.  Instead just include FreeRTOS.h as FreeRTOS.h ensures the other header files are brought in in the correct order.  That is one of the cases where it is necessary to have one header include another.....

Hadatko:
Well this is nonsense for me to request include some file in app in right order, instead of including it by default. For example i am using this header in c file where i don't want any another rtos related include (because why????). This is just bothering me. 
Ou it  is not working for you? it is because you forgot to include this file. We can do it for you but we will not doit.... 
You can add directly related include file there, or your common header file FreeRTOS.h...

PR #105, opened 2018-11-21
Add FreeRTOS.h include warning to library headers

alexa-noxon:
In order to guarantee correct include order, FreeRTOS.h must be
included before FreeRTOS library headers that rely on FreeRTOS
kernel types and prototypes.  These changes add an error with
instructions to include the required header.

Resolves issue #93

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/common/include/aws_logging_task.h
+++ b/demos/common/include/aws_logging_task.h
@@ -26,11 +26,17 @@
 #ifndef AWS_LOGGING_TASK_H
 #define AWS_LOGGING_TASK_H
 
+#ifndef INC_FREERTOS_H
+    #error "include FreeRTOS.h must appear in source files before include aws_logging_task.h"
+#endif
+
 /*
  * Called once to create the logging task and queue.  Must be called before any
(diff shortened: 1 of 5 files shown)


alexa-noxon:
Hi Hadatko,

While that solution would work, in general, the policy for FreeRTOS has been to try to avoid including headers inside of headers when possible.  I agree that there are definitely pros and cons to just including the headers vs adding warnings.   

Sorry for the inconvenience this may have caused and thanks for pointing it out - hopefully adding these warnings can help others in the future narrow in on compile problems much faster!

Hadatko:
Thanks for explanation. But policy should change if that make sense. I didn't hear exaplanation why this policy is good. There are many ways how to solve this. But you prefer bothering your customers :/ Why is including header file bad. Only reason i see is laziness to do it properly.

MarekNovakNXP:
If a header file relies on being included in some specific order with other headers, even if this could be avoided, I don't understand how this can be possibly a good practice.... Just an opinion.

PR #109, opened 2018-11-24
Resolve warnings in TI's test project

tgsong:
Description
-----------

2 Warnings resolved in this commit:
- FreeRTOS.h redefine UART_PRINT from uart_term.h
- No effect statement in network_if.c

Previously UART_PRINT was defined to nothing so that test output is not
clobbered. This commit moves the redefine from uart_term.h to network_if.c
and redefine it to do {} while(0) to resolve 2 warnings introduced above.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tests/ti/cc3220_launchpad/common/application_code/ti_code/network_if.c
+++ b/tests/ti/cc3220_launchpad/common/application_code/ti_code/network_if.c
@@ -58,6 +58,12 @@
 #include "network_if.h"
 #include "uart_term.h"
 
+/* Redefine UART_PRINT to no-op so that test output is not clobbered. */
+#ifdef UART_PRINT
+#undef UART_PRINT
+#define UART_PRINT(X, ...) do {} while(0)
+#endif
+
 //*****************************************************************************
 //                          LOCAL DEFINES
 //*****************************************************************************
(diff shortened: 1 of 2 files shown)


alexa-noxon:
are we concerned about these changes being blown away if we update SDKs some time in the future?

tests/ti/cc3220_launchpad/common/application_code/ti_code/network_if.c
```diff
@@ -58,6 +58,12 @@
 #include "network_if.h"
 #include "uart_term.h"
 
+/* Redefine UART_PRINT to no-op so that test output is not clobbered. */
```
dcgaws: You're changing the UART_PRINT macro in uart_term.h, which is included by this module, as part of this PR. So why add these lines here?

Also, if anybody falls into this block of code with the device prompting the developer to type in a new SSID, they're going to be confused because there won't be any prompt text.

tgsong: We **changed** the `UART_PRINT` print macro in uart_term.h before, this commit reverts them. If you compare the file with the one in demos code, they look the same now.

As to the second point, this is test code, which is only for vendor who are qualifying a new board and running our tests. That's why we disabled the `UART_PRINT` macro before because it's causing problems with our test automation. This commit only fixed the warning problems we introduced before.

alexa-noxon:
Closing as this is out of date with the directory re-org

PR #117, opened 2018-11-28
Modify style checker for PKCS #11 types

alexa-noxon:
PKCS #11 types may follow the format CK_TYPE_PTR or CK_TYPE_PTR_PTR.
This change updates the style checker to recognize _PTR and _PTR_PTR
line endings as pointer types.

This change also allows variables to start with a number.

Checklist:
----------

- [ ] ~I have tested my changes. No regression in existing tests.~
- [ ] ~My code is Linted.~

### DIFF
--- a/tools/checks/style/hn_check/src/hn_check.py
+++ b/tools/checks/style/hn_check/src/hn_check.py
@@ -233,8 +233,7 @@ def get_base_prefix(type_name):
 def get_identifier_prefix(identifier):
     direct_identifier = re.sub(r"^p+", "", identifier)
     indirection = "p" * (len(identifier) - len(direct_identifier))
-    for key in TYPE_PREFIXES:
-        prefix = TYPE_PREFIXES[key]
+    for prefix in TYPE_PREFIXES.values():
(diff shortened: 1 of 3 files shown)


tools/checks/style/hn_check/src/hn_check.py
```diff
@@ -224,7 +235,7 @@ def get_identifier_prefix(identifier):
     indirection = "p" * (len(identifier) - len(direct_identifier))
     for key in TYPE_PREFIXES:
```
tgsong: Though this line didn't change, looks like we can just do `for prefix in TYPE_PREFIXES.values():`, `key` is only used to get values from this dictionary.

tools/checks/style/hn_check/src/hn_check.py
```diff
@@ -247,4 +258,4 @@ def main():  # pragma: no cover
 
 
 if __name__ == "__main__":  # pragma: no cover
-    main()
+    main()
```
tgsong: Minor thing, missing newline here.

alexa-noxon: good catch.  I got confused by the whitespace checker output.

PR #144, opened 2018-12-07
Fix/ble large obj transfer

ravibhagavandas:
Fixes bugs within WiFi provisioning service and Large Object Transfer

Description
-----------
1. Fixes a bug in WiFi provisioning service, where the background task to connect to saved networks blocks a list network request.
 - Added priorities for each task and changed lock to a mutex.
 - Introduced delay between re-connection to the saved networks.
2. Fixes a possible deadlock in closing Mqtt Connection while a large object transfer is ongoing.
3. Renamed demo network connection APIs and added more docs.

Checklist:
----------
- [ ] My code is Linted.

### DIFF
--- a/lib/bluetooth_low_energy/services/mqtt_ble/aws_iot_mqtt_ble.c
+++ b/lib/bluetooth_low_energy/services/mqtt_ble/aws_iot_mqtt_ble.c
@@ -129,14 +129,13 @@ void vToggleMQTTService( BLEAttribute_t * pxAttribute,
                        BLEAttributeEvent_t * pxEventParam );
 
 /*
- * @brief Resets the send and received buffer for the given MQTT connection.
- * MQTT connection should be closed or the service should be disabled
- * before resetting the buffer.
+ * @brief Resets the send and receive buffer for the given MQTT Service.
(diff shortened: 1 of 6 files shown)


lib/bluetooth_low_energy/services/mqtt_ble/aws_iot_mqtt_ble.c
```diff
@@ -634,11 +671,23 @@ void vRXLargeMesgCharCallback( BLEAttribute_t * pxAttribute,
 
         			if( pxWriteParam->xLength < mqttBLETRANSFER_LEN( usBLEConnMTU ) )
         			{
+                                        for ( x =0; x < pxService->xConnection.xRecvOffset; x++ )
+                                        {
+                                            if( pxService->xConnection.pRecvBuffer[x] == '\\' )
+                                            {
+                                                pBuf = ( pxService->xConnection.pRecvBuffer ) + x - 100;
+
```
huguesBouvier: it is a magic number. Maybe do a #define

ravibhagavandas: Its fixed in second commit.

PR #146, opened 2018-12-09
Repairing the DNS asynchronous lookup

htibosch:
The API `FreeRTOS_gethostbyname_a()` is the asynchronous version of `FreeRTOS_gethostbyname()`. It makes a call-back as soon as an answer from a DNS server is received, or after a time-out was reached.

-----------

Due to a regression ( commit v1.3.2 on august 23, 2018 ), `FreeRTOS_gethostbyname_a()` stopped working.

In `xProcessReceivedUDPPacket()`, `ulDNSHandlePacket()` must be called to process the ( asynchronous ) DNS-reply.

The function `ulDNSHandlePacket()` now has two packet-length checks that are misplaced. These checks would be OK if the field `xDataLength` contains the total packet length. However, in line 1615 in FreeRTPS_IP.c :

    pxNetworkBuffer->xDataLength -= sizeof( UDPPacket_t );

Within `ulDNSHandlePacket()`, `xDataLength` represents the length of the UDP payload. The only test needed is this one:

    if ( xPlayloadBufferLength < sizeof( DNSMessage_t ) )
    {
        return pdFAIL;
    }

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
@@ -817,10 +817,6 @@ DNSMessage_t *pxDNSMessageHeader;
 	/* In this context, 'pxNetworkBuffer->xDataLength' is the
 	length of the UDP payload buffer. */
 	xPlayloadBufferLength = pxNetworkBuffer->xDataLength;
-	if ( xPlayloadBufferLength < sizeof( DNSMessage_t ) )
-	{
-		return pdFAIL;
-	}
 
 	pucUDPPayloadBuffer = pxNetworkBuffer->pucEthernetBuffer + sizeof( UDPPacket_t );
 	pxDNSMessageHeader = ( DNSMessage_t * ) pucUDPPayloadBuffer;
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
@@ -437,7 +437,7 @@ TickType_t xIdentifier = 0;
 			}
 			else
 			{
-				/* prvGetHostByName will be called to start a DNS lookup */
+				/* prvGetHostByName will be called to start a DNS lookup. */
 			}
 		}
 	}
@@ -446,7 +446,8 @@ TickType_t xIdentifier = 0;
 	/* Generate a unique identifier. */
 	if( 0 == ulIPAddress )
 	{
-		xIdentifier = ( TickType_t )ipconfigRAND32( );
+		/* DNS identifiers are 16-bit. */
+		xIdentifier = ( TickType_t )( ipconfigRAND32( ) & 0xffff );
 	}
 
 	#if( ipconfigDNS_USE_CALLBACKS != 0 )
@@ -567,7 +568,7 @@ TickType_t xWriteTimeOut_ms = ipconfigSOCK_DEFAULT_SEND_BLOCK_TIME;
 					if( lBytes > 0 )
 					{
 						/* The reply was received.  Process it. */
-						ulIPAddress = prvParseDNSReply( pucUDPPayloadBuffer, lBytes, xIdentifier );
+						ulIPAddress = prvParseDNSReply( pucUDPPayloadBuffer, ( size_t )lBytes, xIdentifier );
 
 						/* Finished with the buffer.  The zero copy interface
 						is being used, so the buffer must be freed by the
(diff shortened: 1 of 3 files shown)


htibosch:
There is also this change in FreeRTOS_DNS.c:

`-	xIdentifier = ( TickType_t )ipconfigRAND32( );`
`+	/* DNS identifiers are 16-bit. */`
`+	xIdentifier = ( TickType_t )( ipconfigRAND32( ) & 0xffff );`

If `xIdentifier` receives a 32-bit random number, the call-back will never be made. A DNS identifier is, by definition, only 16 bits wide.

Here is the call-back, in which `usIdentifier` is 16-bits:

`	vDNSDoCallback( ( TickType_t ) pxDNSMessageHeader->usIdentifier, pcName, ulIPAddress );`

This worked well until commit v1.3.2 on august 23, 2018.

htibosch:
Dan, thanks for the explanation. I was already wondering why this PR didn't receive any attention.

I want to emphasise however, that the proposed changes are a simple reparation of a regression which was caused the commit v1.3.2. After that commit, both LLMNR and asynchronous DNS lookups stopped working. People on the FreeRTOS forum are complaining about this.

Part of my earlier PR has been accepted and merged, a few changes, described here above, are still pending.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -446,7 +446,8 @@ TickType_t xIdentifier = 0;
 	/* Generate a unique identifier. */
 	if( 0 == ulIPAddress )
 	{
-		xIdentifier = ( TickType_t )ipconfigRAND32( );
+		/* DNS identifiers are 16-bit. */
+		xIdentifier = ( TickType_t )( ipconfigRAND32( ) & 0xffff );
```
dcgaws: Yes, but in the context of this function, the identifier is a TickType_t, so the mask seems risky. An alternative would be to change the type of xIdentifier, as well as the prototype of the routine below that it gets passed to.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -567,7 +568,7 @@ TickType_t xWriteTimeOut_ms = ipconfigSOCK_DEFAULT_SEND_BLOCK_TIME;
 					if( lBytes > 0 )
 					{
 						/* The reply was received.  Process it. */
-						ulIPAddress = prvParseDNSReply( pucUDPPayloadBuffer, lBytes, xIdentifier );
+						ulIPAddress = prvParseDNSReply( pucUDPPayloadBuffer, ( size_t )lBytes, xIdentifier );
```
dcgaws: Since the receive routine returns int32_t, would it make more sense to change prvParseDNSReply to take that type instead of size_t?

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -813,21 +814,16 @@ uint8_t *pucUDPPayloadBuffer;
 size_t xPlayloadBufferLength;
 DNSMessage_t *pxDNSMessageHeader;
 
-	xPlayloadBufferLength = pxNetworkBuffer->xDataLength - sizeof( UDPPacket_t );
-	if ( xPlayloadBufferLength < sizeof( DNSMessage_t ) )
-	{
-		return pdFAIL;
-	}
+	/* In this context, 'pxNetworkBuffer->xDataLength' is the
```
dcgaws: Agreed, although I don't think we should remove the check for minimal size of a DNS message.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_UDP_IP.c
```diff
@@ -347,6 +347,19 @@ UDPPacket_t *pxUDPPacket = (UDPPacket_t *) pxNetworkBuffer->pucEthernetBuffer;
 		/* There is no socket listening to the target port, but still it might
 		be for this node. */
 
+		#if( ipconfigUSE_DNS == 1 ) && ( ipconfigDNS_USE_CALLBACKS == 1 )
```
dcgaws: Wouldn't this proposed code make it easier for the device to be spoofed? A rogue device on the network can send DNS packets to the device at any time and the device will add those entries to its cache.

htibosch:
I will close the PR and merge the propose changes with my new [PR 293](https://github.com/aws/amazon-freertos/pull/293).
Hein

PR #157, opened 2018-12-14
initial commit of new defender library header

qiutongs:
Description
-----------

This is initial commit of new defender library header. It has defined data types and primary functions faced by external users.

Checklist:
----------

- [NO ] I have tested my changes. No regression in existing tests.
- [X] My code is Linted.

### DIFF
--- a/lib/include/aws_iot_defender.h
+++ b/lib/include/aws_iot_defender.h
@@ -1,0 +1,142 @@
+/*
+ * Copyright (C) 2018 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file aws_iot_defender.h
+ * @brief
+ */
+
+#ifndef _AWS_IOT_DEFENDER_H_
+#define _AWS_IOT_DEFENDER_H_
+
+/* Standard includes. */
+#include <stdint.h>
+
+/**
+ * This enum type is the ID of metrics. It can be specified in AwsIotDefender_SetMetrics function.
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_METRIC_TCP_ESTABLISHED_TOTAL,
+} AwsIotDefenderMetric_t;
+
+/**
+ * Return codes of defender functions.
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_SUCCESS = 0,
+    AWS_IOT_DEFENDER_ALREADY_STARTED,
+    AWS_IOT_DEFENDER_PERIOD_TOO_SHORT,
+    AWS_IOT_DEFENDER_OUT_OF_MEMORY,
+
+    AWS_IOT_DEFENDER_INTERNAL_FAILURE
+} AwsIotDefenderError_t;
+
+/**
+ * Root cause code of failure, used in callback
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_FAILURE_NONE = 0,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CREATION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CONNECTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_SUBSCRIPTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_PUBLISH,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_DISCONNECTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_DELETION,
+    AWS_IOT_DEFENDER_FAILURE_METRICS_REJECTION
+} AwsIotDefenderFailureRootCause_t;
+
+/**
+ * Callback parameters.
+ */
+typedef struct AwsIotDefenderCallbackInfo
+{
+    const void * pPayload; /**< The received message if there is. */
+    size_t payloadLength;  /**< Length of the message. */
+    AwsIotDefenderFailureRootCause_t failureRootCause;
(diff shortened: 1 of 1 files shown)


lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Return a string that describes AwsIotDefenderError_t
+ */
+const char * AwsIotDefender_StrError( AwsIotDefenderError_t error );
+
+/**
+ * Return a string that describes AwsIotDefenderFailureRootCause_t
+ */
+const char * AwsIotDefender_StrFailureRootCause( AwsIotDefenderFailureRootCause_t failureRootCause );
```
lt72: Name is not consistent with AwsIotDefender_StrError.

qiutongs: Sorry, are you saying I should combine AwsIotDefender_StrError and AwsIotDefender_StrFailureRootCause to a single function?

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Return codes of defender functions.
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_SUCCESS = 0,
+    AWS_IOT_DEFENDER_ALREADY_STARTED,
+    AWS_IOT_DEFENDER_PERIOD_TOO_SHORT,
+    AWS_IOT_DEFENDER_OUT_OF_MEMORY,
```
gordonwang0: The other libraries call this `NO_MEMORY`, so `AWS_IOT_DEFENDER_NO_MEMORY`.

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+
+    AWS_IOT_DEFENDER_INTERNAL_FAILURE
```
gordonwang0: I think we should avoid "generic failure" error codes...

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+ * Root cause code of failure, used in callback
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_FAILURE_NONE = 0,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CREATION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CONNECTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_SUBSCRIPTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_PUBLISH,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_DISCONNECTION,
```
gordonwang0: In MQTT v4, disconnect never fails.

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+    AWS_IOT_DEFENDER_INTERNAL_FAILURE
+} AwsIotDefenderError_t;
+
+/**
+ * Root cause code of failure, used in callback
+ */
+typedef enum
+{
+    AWS_IOT_DEFENDER_FAILURE_NONE = 0,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CREATION,
```
gordonwang0: This option should be removed for MQTT v4

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+    AWS_IOT_DEFENDER_FAILURE_MQTT_CONNECTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_SUBSCRIPTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_PUBLISH,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_DISCONNECTION,
+    AWS_IOT_DEFENDER_FAILURE_MQTT_DELETION,
```
gordonwang0: Also remove for MQTT v4.

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+    AWS_IOT_DEFENDER_FAILURE_METRICS_REJECTION
+} AwsIotDefenderFailureRootCause_t;
+
+/**
+ * Callback parameters.
+ */
+typedef struct AwsIotDefenderCallbackInfo
+{
+    const void * pPayload; /**< The received message if there is. */
```
gordonwang0: why void*?

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Stop the defender
+ */
+AwsIotDefenderError_t AwsIotDefender_Stop( void );
+
+/**
+ * Set period in seconds.
+ */
+AwsIotDefenderError_t AwsIotDefender_SetPeriod( uint32_t periodSeconds );
```
gordonwang0: Should use `uint64_t` for time.

lib/include/aws_iot_defender.h
gordonwang0: Maybe call this one AwsIotDefender_FailureRootCause?

lib/include/aws_iot_defender.h
```diff
@@ -0,0 +1,142 @@
...
+                                                 size_t metricsCount );
+
+/**
+ * Start the defender */
+AwsIotDefenderError_t AwsIotDefender_Start( AwsIotDefenderStartInfo_t * pStartInfo );
+
+/**
+ * Stop the defender
+ */
+AwsIotDefenderError_t AwsIotDefender_Stop( void );
```
gordonwang0: Does this ever fail?  If not, make void return type.

lib/include/aws_iot_defender.h
qiutongs: ok

lib/include/aws_iot_defender.h
qiutongs: ok, I will remove

lib/include/aws_iot_defender.h
qiutongs: great, I will remove

lib/include/aws_iot_defender.h
qiutongs: But there are cases we can't exactly know the problem. It could be an internal bug, e.g. timer creation failure

lib/include/aws_iot_defender.h
qiutongs: ok

lib/include/aws_iot_defender.h
qiutongs: because MQTT defines AwsIotMqttPublishInfo_t:  const void * pPayload

lib/include/aws_iot_defender.h
gordonwang0: I see.  In that case, it's fine.

lib/include/aws_iot_defender.h
qiutongs: timer delete can fail? Even we don't have failure case today, isn't it future-proof to return error?

lib/include/aws_iot_defender.h
qiutongs: You are saying uint32_t is too small?

lib/include/aws_iot_defender.h
qiutongs: ok

lib/include/aws_iot_defender.h
gordonwang0: Yes, makes sense.

lib/include/aws_iot_defender.h
gordonwang0: In the other libraries, time is in milliseconds, and for that uint32_t was too small.

I realize in Defender it must be in seconds... but just for consistency, I think we should make is uint64_t.

lib/include/aws_iot_defender.h
gordonwang0: What format is the payload in?  If it's UTF-8 or binary, maybe make the pointer type char*/uint8_t*.

lib/include/aws_iot_defender.h
qiutongs: ok

lib/include/aws_iot_defender.h
qiutongs: I see. MQTT is more general, thus defining void *. For defender, uint8_t * should be the case. Good catch.

PR #172, opened 2018-12-21
update implementation for defender's public APIs

qiutongs:
Description
-----------

- separate implementation source files, adding internal functions in internal header
    - api functions
    - mqtt/network management (not included in this PR)
    - metrics collector (not included in this PR)
    - specific metrics reporter (not included in this PR)

Checklist:
----------

- [X] My code is Linted.

### DIFF
--- a/lib/defender/aws_iot_defender.c
+++ b/lib/defender/aws_iot_defender.c
@@ -19,41 +19,23 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#include <stdbool.h>
-
 /* Defender internal include. */
 #include "private/aws_iot_defender_internal.h"
 
-#include "platform/aws_iot_network.h"
-
+/* Platform clock include. */
 #include "platform/aws_iot_clock.h"
 
-/* helper macros to convert seconds to milliseconds and vice versa. */
-#define _toMilliseconds( secondValue )    ( secondValue ) * 1000
+/*
+ * Small time value to kick off timer immediately.
+ */
+#define _DEFENDER_SHORT_RELATIVE_MILLISECONDS    100
 
-#define _toSeconds( millisecondValue )    ( millisecondValue ) / 1000
-
-/* Period between reports in milliseconds. */
-static uint64_t _periodMilliSecond = _toMilliseconds( AWS_IOT_DEFENDER_DEFAULT_PERIOD_SECONDS );
-
-/* the timer that connect, subscribe, publish metrics report */
-static AwsIotTimer_t _metricsPublishTimer;
-
-/* the timer that disconnect */
-static AwsIotTimer_t _guardTimer;
-
-/* defender internally manages network and mqtt connection */
-static AwsIotNetworkConnection_t _networkConnection = AWS_IOT_NETWORK_CONNECTION_INITIALIZER;
-
-static AwsIotMqttConnection_t _mqttConnection = AWS_IOT_MQTT_CONNECTION_INITIALIZER;
-
-static AwsIotDefenderStartInfo_t _startInfo = { 0 };
-
-static char * _pPublishTopic = NULL;
-
-static char * _pAcceptTopic = NULL;
-
-static char * _pRejectTopic = NULL;
+/*
+ * This mode is supposed to be used by test, demo or debug.
+ */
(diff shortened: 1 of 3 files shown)


lib/include/aws_iot_defender.h
```diff
@@ -21,45 +21,84 @@
...
 /**
- * This enum type is the ID of metrics. It can be specified in AwsIotDefender_SetMetrics function.
+ * @brief Metrics group options for AwsIotDefender_SetMetrics() function.
  */
 typedef enum
 {
-    AWS_IOT_DEFENDER_METRIC_TCP_ESTABLISHED_TOTAL,
-} AwsIotDefenderMetric_t;
+    AWS_IOT_DEFENDER_METRICS_TCP_CONN,
+    AWS_IOT_DEFENDER_METRICS_LIS_TCP,
```
lt72: What does LIS stands for? Not very intuitive...

lt72: LISTENING? Better long names...  aldo make CONN into CONNECTION

lib/defender/aws_iot_defender.c
```diff
@@ -251,41 +268,56 @@ static void _metricsPublishTimerExpirationRoutine( void * pArgument )
 {
```
lt72: This routing *could* (it won't, but could...) fire while the previous one is executing... you should atomically (if possible) set a flag that indicates that metrics collection and upload is in progress and therefore if it fires agaon it should do nothing and let the previous one terminate. Just so that the code is formally correct and an occasional reviewer that does not know we have a 5 minutes minimum interval will not be surprised.

lib/include/private/aws_iot_defender_internal.h
```diff
@@ -51,45 +131,95 @@
...
+/**
+ * Listening TCP ports metrics is added to the encoder object.
+ */
+AwsIotSerializerError_t AwsIotDefenderInternal_GetLisTcpMetrics( AwsIotSerializerEncoderObject_t * pEncoderObject );
+
+/**
+ * Build three topics names used by defender library.
+ */
+bool AwsIotDefenderInternal_BuildTopicsNames( const char * pThingName,
+                                              uint16_t thingNameLength,
```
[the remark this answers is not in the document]
qiutongs: In MQTT connect, the client identifier length is defined as uint16_t. So I made it the same.

lib/defender/aws_iot_defender.c
```diff
@@ -19,98 +19,112 @@
...
-
-static AwsIotDefenderStartInfo_t _startInfo = { 0 };
-
-static char * _pPublishTopic = NULL;
-
-static char * _pAcceptTopic = NULL;
+/*
+ * Small time value to kick off timer immediately.
+ */
+#define _DEFENDER_SHORT_RELATIVE_MILLISECONDS    100
```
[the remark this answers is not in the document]
qiutongs: Sure. But I am curious how you reasoned about that only one thread will go before?

lib/include/private/aws_iot_defender_internal.h
```diff
@@ -51,45 +131,95 @@
...
+ * Set the network connection to callback MQTT.
+ */
+bool AwsIotDefenderInternal_SetMqttCallback( AwsIotDefenderEventType_t * pEventType );
 
-#define _TOPIC_SUFFIX_REJECTED          _TOPIC_SUFFIX_PUBLISH "/rejected"
+/**
+ * Connect to AWS with MQTT.
+ */
+bool AwsIotDefenderInternal_MqttConnect( const char * pThingName,
+                                         uint16_t thingNameLength,
```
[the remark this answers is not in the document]
qiutongs: Yeah, AwsIotMqttConnectInfo_t defines clientIdentifierLength as uint16_t

lib/include/aws_iot_defender.h
qiutongs: I like long time. I was just concerned if the macro name too long will cause problem..

qiutongs: I like long *name*

lib/defender/aws_iot_defender.c
qiutongs: I agree totally. I will fix it.

qiutongs: A second thought: let's deal with it after we have the plan for moving to task pool.

PR #177, opened 2018-12-27
implementation of sockets reporter which serialize sockets metrics da…

qiutongs:
…ta to report

Description
-----------

- serialize sockets metrics data to report based on user's input

Checklist:
----------

- [X I have tested my changes. No regression in existing tests.
- [X My code is Linted.

### DIFF
--- a/lib/defender/aws_iot_defender_reporter_socket.c
+++ b/lib/defender/aws_iot_defender_reporter_socket.c
@@ -1,0 +1,255 @@
+/*
+ * Copyright (C) 2018 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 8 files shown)


lib/defender/aws_iot_defender_reporter_socket.c
```diff
@@ -0,0 +1,255 @@
...
+    uint8_t hasTotal = ( tcpConnFlag & AWS_IOT_DEFENDER_METRICS_TCP_CONNECTIONS_ESTABLISHED_TOTAL ) > 0;
+    uint8_t hasLocalPort = ( tcpConnFlag & AWS_IOT_DEFENDER_METRICS_TCP_CONNECTIONS_ESTABLISHED_LOCAL_PORT ) > 0;
+    uint8_t hasRemoteAddr = ( tcpConnFlag & AWS_IOT_DEFENDER_METRICS_TCP_CONNECTIONS_ESTABLISHED_REMOTE_ADDR ) > 0;
+
+    uint8_t total = AwsIotDefenderInternal_GetEstablishedConnectionsCount();
+
+    _defenderMetricsConnection_t * pConnections = NULL;
+
+    /* 15 for IP + 1 for ":" + 5 for port + 1 terminator */
+    char remoteAddr[ 15 + 1 + 5 + 1 ] = "";
```
lt72: I would declare a macro at the top of the file to track the value 15 + 1 + 5 + 1

lib/defender/aws_iot_defender_reporter_socket.c
```diff
@@ -0,0 +1,255 @@
...
+    {
+        error = _AwsIotDefenderEncoder.closeContainer( pEncoderObject, &tcpConnectionMap );
+    }
+
+    return error;
+}
+
+/*-----------------------------------------------------------*/
+
+AwsIotSerializerError_t AwsIotDefenderInternal_GetLisTcpMetrics( AwsIotSerializerEncoderObject_t * pEncoderObject )
```
lt72: If this function is only called in this file then it should be called
_GetLisTcpMetrics...

what does LIs means? it is not obvious, so consider changing name...

lib/defender/aws_iot_defender_reporter_socket.c
qiutongs: ok, I will do.

lib/defender/aws_iot_defender_reporter_socket.c
qiutongs: This function is implemented in this file but used by "collector" to aggregate report.

It should be "Listening TCP Ports". I will change the name.

PR #188, opened 2019-01-04
Fix ESP32 compilation errors when BLE is disabled

copercini:
Description
-----------

Using the branch feature/ble-beta and disabling ESP32 BLE, still are some includes and operations related to BLE causing compilation errors, this PR has the intent to fix it adding some extra ifs

Checklist:
----------

- [ ] My code is Linted.


abhidixi11:
The Amazon FreeRTOS team is currently in the process of transitioning to using Github as a primary development repository and unfortunately, there is still a little work to be done in determining the internal procedure for accepting public contributions.
In this particular case, we already have commited the fix to BLE branch.

PR #189, opened 2019-01-04
Qiutongs/defender tmp

qiutongs:
Description
-----------

- not destroy resources in start function if it fails due to "already started"
- not destroy resources in stop function if it has not started yet
- use semaphore to have stop function wait for timers to finish
    - wait for potential metrics timer callback to finish before destroying semaphore
- ignore small buffer in collector(not consider "buffer too small" as a serialization failure"
- use mutex to protect metrics flags array(between SetMetrics functions and metrics timer callback)

Checklist:
----------

- [X] My code is Linted.


lib/defender/aws_iot_defender.c
```diff
@@ -74,6 +79,8 @@ static void _guardTimerExpirationRoutine( void * );
 
 uint32_t _AwsIotDefenderMetricsFlag[ _DEFENDER_METRICS_GROUP_COUNT ];
 
+AwsIotMutex_t _AwsIotDefenderMetricsMutex;
```
lt72: How about placing all of this global types in a struct? Would it increase readability?

lib/defender/aws_iot_defender.c
```diff
@@ -86,19 +93,26 @@ static AwsIotTimer_t _metricsPublishTimer;
 /* the timer that disconnect */
 static AwsIotTimer_t _guardTimer;
 
+/*
+ * This semaphore is to synchronize AwsIotDefender_Stop and metrics timer callback + MQTT callback + guard timer callback.
+ * AwsIotDefender_Stop waits on this semaphore.
+ * Metrics timer callback tries to wait on this semaphore.
+ */
+static AwsIotSemaphore_t _timerSem;
```
lt72: Do we actually need a semaphore and a mutex?

lib/defender/aws_iot_defender.c
qiutongs: make sense.  Let me change it.

lib/defender/aws_iot_defender.c
qiutongs: I think the semaphore is necessary because you need to implement this logic: destroy resources only after the timer callback completes execution.

The mutex for metrics can be left out if we don't need to support modifying metrics after defender is started.

PR #196, opened 2019-01-07
performance counter -- driver for st.

yuhui-zheng:
Implementation of HAL performance counter APIs for STM32 discovery. 

Description
-----------
This PR is purely "device driver", which includes configuration for HW peripheral timer, IRQ, compilation, and etc. I would like to get code structure and generic steps reviewed. 

Known todos:
- FreeRTOSConfig.h, when configUSE_HAL_PERF_COUNTER is set to 0, perf counter APIs and memory allocation should be completely switched off.
- Code formatting. My debugging env is on MacOS && Eclipse, I need to format everything with our uncrustify cfg. 

Things you may find interesting -- the way perf counter resolution is configured:
In FreeRTOSConfig.h, user seems to be able to define whatever counter resolution they want with ```configHAL_PERF_COUNTER_FREQ```. However, to minimize the need for kernel to server counter overflow IRQ, timer period is always set to the maximum it allows. (for 32-bit counter, 0xffffffff) And period absolute duration is adjusted with clock prescaler, which only takes integer. 

e.g.1 APB1 bus frequency max is 80MHz, and our test project use 80MHz as APB1 frequency. When user sets ```configHAL_PERF_COUNTER_FREQ = 50MHz```, ```prescaler = (uint32_t) 80Mhz/50Mhz = 1```. In this case, timer source clock is still 80MHz, and  ```aws_hal_perfcounter_get_frequency_hz()``` returns 80MHz.

e.g.2 ```configHAL_PERF_COUNTER_FREQ = 40MHz```, ```prescaler = 2```. So, ```aws_hal_perfcounter_get_frequency_hz()``` returns 40MHz.

See more in code comment.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/hal/include/aws_hal_perf_cntr.h
+++ b/lib/hal/include/aws_hal_perf_cntr.h
@@ -1,0 +1,76 @@
+/*
+* Amazon FreeRTOS V1.x.x
+* Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included in
+* all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*
+* http://aws.amazon.com/freertos
+* http://www.FreeRTOS.org
+*/
+
+/**
+* @file aws_hal_perf_cntr.h
+* @brief File for the APIs of performance counter called by application layer.
+*
+* Performance counter uses a hardware peripheral timer to track time elapsed since
+* the start of the counter. The interface implementation is MCU specific, and counter
+* resolution is configurable via FreeRTOSConfig.h.
+*
+* To use the interfaces, application code needs to:
+* - Initialize counter by calling aws_hal_perfcounter_open().
+* - Get counter value for as many times as you want in whatever thread context
+*   by calling aws_hal_perfcounter_get_value().
+* - Time elapsed can be derived by ( counter value / counter frequency ),
+*   where counter frequency can be acquired by calling aws_hal_perfcounter_get_frequency_hz().
+* - Once completely done with performance measurement,
+*   free resources by calling aws_hal_perfcounter_close().
+*
+* @warning It's not recommended to repurpose the hardware timer, nor have timer vector interrupt
+*          priority lower than other peripheral vector interrupt priorities. Please refer to implementation
+*          specific details.
+*/
+
+#ifndef _AWS_HAL_PERF_CNTR_H_
+#define _AWS_HAL_PERF_CNTR_H_
+
+/**
+ * @brief Initialize a hardware timer which is to be used as performance counter.
+ */
+void aws_hal_perfcounter_open(void);
+
+/**
+ * @brief Deinitialize the hardware timer.
+ */
+void aws_hal_perfcounter_close( void);
+
+/**
+ * @brief Get current count from the performance counter.
+ *
+ * @return Total count since counter started.
+ */
+uint64_t aws_hal_perfcounter_get_value(void);
+
+/**
+ * @brief Get configured frequency of the performance counter.
+ *
+ * @return Frequency which the counter is configured to run at.
+ */
+uint32_t aws_hal_perfcounter_get_frequency_hz(void);
+
+#endif /* _AWS_HAL_PERF_CNTR_H_ */
--- a/lib/hal/portable/st/aws_hal_perf_cntr.c
+++ b/lib/hal/portable/st/aws_hal_perf_cntr.c
@@ -1,0 +1,233 @@
+/*
+* Amazon FreeRTOS V1.x.x
+* Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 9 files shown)


lib/hal/include/aws_hal_perf_cntr.h
```diff
@@ -0,0 +1,76 @@
+/*
+* Amazon FreeRTOS V1.x.x
```
yuhui-zheng: I'll update FreeRTOS version before merging to master.

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+    xTimerHandle.Init.RepetitionCounter = 0x00;
+    xTimerHandle.Init.AutoReloadPreload = TIM_AUTORELOAD_PRELOAD_ENABLE;
+
+    xTimerHandle.Channel = HAL_TIM_ACTIVE_CHANNEL_CLEARED;
+
+    xTimerHandle.Lock = HAL_UNLOCKED;
+
+    xTimerHandle.State = HAL_TIM_STATE_RESET;
+
+    HAL_TIM_Base_Init( &xTimerHandle );
```
yuhui-zheng: Error_Handler() upon issue.

mradulan: should this be HAL_TIM_Base_MspInit instead? I do not see HAL_TIM_Base_MspInit being called anywhere.

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+    HAL_TIM_Base_Start_IT( &xTimerHandle );
+}
+
+/*-----------------------------------------------------------*/
+
+void aws_hal_perfcounter_close( void)
+{
+    HAL_TIM_Base_Stop_IT( &xTimerHandle );
+
+    HAL_TIM_Base_DeInit( &xTimerHandle );
```
mradulan: Should this be HAL_TIM_Base_MspDeInit?

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+
+    /* To minimize the need for serving overflow IRQ,
+     * .Period is always set to the max value the counter allows.
+     * Performance counter resolution is adjusted with .Prescaler.
+     */
+    xTimerHandle.Init.Prescaler = ulTimClock / HAL_PERF_COUNTER_FREQ;
+    xTimerHandle.Init.CounterMode = TIM_COUNTERMODE_UP;
+    xTimerHandle.Init.Period = HW_TIMER_32_CONST_PERIOD;
+    xTimerHandle.Init.ClockDivision = TIM_CLOCKDIVISION_DIV1;
+    xTimerHandle.Init.RepetitionCounter = 0x00;
```
mradulan: can you assign a variable instead of hardcoding 0x00

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+* @file aws_hal_perf_cntr.c
+* @brief File of the implementation for performance counter APIs calling STM drivers.
+*/
+
+#include <stdint.h>
+
+/* ST HAL APIs to be called in this file. */
+#include "stm32l4xx_hal.h"
```
mradulan: should this header be included stm32l4xx_hal_tim.h?

yuhui-zheng: I actually tried stm32l4xx_hal_tim.h first, which failed to compile. (That was when I was trying to implement error handling path for _open(), and HAL_StatusTypeDef was missing from *_tim.h header.) I'll meditate over this. If all we have to do is ignoring errors, then *_tim.h would do the work.

lib/hal/portable/st/aws_hal_perf_cntr.c
yuhui-zheng: Will make a macro in next PR.

lib/hal/portable/st/aws_hal_perf_cntr.c
yuhui-zheng: HAL_TIM_Base_{Init, DeInit} call HAL_TIM_Base_Msp{Init, DeInit} internally. From ST HAL programming manual, *_Msp*() is pretty much an extension point where user can define additional routine for initialization and deinitialization. So, in our case, besides regular timer initialization routine, we are turning on interrupt for TIM5.

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+/*-----------------------------------------------------------*/
+
+uint32_t aws_hal_perfcounter_get_frequency_hz(void)
+{
+    uint32_t ulTimClock = 0;
+
+    /* Get timer clock frequency. */
+    ulTimClock = HAL_RCC_GetPCLK1Freq();
+
+    return ( ulTimClock / ( uint32_t )( ulTimClock / HAL_PERF_COUNTER_FREQ ) );
```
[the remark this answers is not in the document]
yuhui-zheng: ulTimClock in STM32L4xx's context is the frequency of PCLK1 which comes from APB1 (Advanced Peripheral Bus). In order to make ```HAL_RCC_GetPCLK1Freq()``` return zero, from ST clock tree diagram it requires the entire bus to be shut down, which impacts TIM/RTC/UART/I2C/...... Correct me here, I think the sane way is to keep this bus clock on always, and turn on/off individual components on demand, which makes ```HAL_RCC_GetPCLK1Freq()``` not zero... 

As for "Isn't it just HAL_PERF_COUNTER_FREQ?" It may be easier for people to read if whatever user input for ```HAL_PERF_COUNTER_FREQ ``` in FreeRTOSConfig.h actually reflects the counter frequency. Right not ```HAL_PERF_COUNTER_FREQ``` does take whatever value user inputs, but .prescaler is always set to the closest integer value. (line 177 in the same file.) The trade off made here is that timer doesn't always get the resolution user requested. (e.g. Would 12.5ns vs. 14ns per clock cycle really makes a difference? Probably not... ) But calculation is simpler and the need to serve IRQ is minimized with .period always set to maximum value.

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/*-----------------------------------------------------------*/
+/**
+ * @brief Clean up
+ */
+
+void HAL_TIM_Base_MspDeInit( TIM_HandleTypeDef * pxTimerHandle )
+{
+    if (TIM5 == pxTimerHandle->Instance )
```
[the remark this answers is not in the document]
yuhui-zheng: Will update in next PR.

yuhui-zheng:
For uint64_t type value validation, it seems fine using plain assertion. (TEST_ASSERT_MESSAGE and TEST_ASSERT) So that I'm not flipping on UNITY_SUPPORT_64 macro.

lib/hal/portable/st/aws_hal_perf_cntr.c
```diff
@@ -0,0 +1,237 @@
...
+ */
+#define HW_TIMER_32_WIDTH           ( sizeof( uint32_t ) * 8 )
+#define HW_TIMER_32_CONST_PERIOD    ( UINT32_MAX )
+#define HW_TIMER_32_LOADING_VALUE   ( 0x0UL )
+
+
+/*-----------------------------------------------------------*/
+
+static TIM_HandleTypeDef xTimerHandle = { 0 };
+static uint64_t ullTimerOverflow = 0;
```
lt72: This should just be 32 bit, and then you cast it before shifting it in get_value.

PR #197, opened 2019-01-08
fix: pthread_mutex_unlock update owner atomically

abhidixi11:
Added critical section protection to update owner and unlock atomically

Description
-----------

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_pthread_mutex.c
+++ b/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_pthread_mutex.c
@@ -297,6 +297,10 @@ int pthread_mutex_unlock( pthread_mutex_t * mutex )
 
     if( iStatus == 0 )
     {
+        /* Suspend the sheduler so that
+         * mutex is unlocked AND owner is updated atomically */
+        vTaskSuspendAll();
+
         /* Call the correct FreeRTOS mutex unlock function based on mutex type. */
         if( pxMutex->xAttr.iType == PTHREAD_MUTEX_RECURSIVE )
         {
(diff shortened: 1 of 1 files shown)


lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_pthread_mutex.c
```diff
@@ -297,6 +297,10 @@ int pthread_mutex_unlock( pthread_mutex_t * mutex )
 
     if( iStatus == 0 )
     {
+        /* Enter the critical section so that
+         * mutex is unlocked AND owner is updated atomically */
+        taskENTER_CRITICAL();
+
         /* Call the correct FreeRTOS mutex unlock function based on mutex type. */
         if( pxMutex->xAttr.iType == PTHREAD_MUTEX_RECURSIVE )
         {
```
gordonwang0: > FreeRTOS API functions must not be called from within a critical section. 

See: https://www.freertos.org/taskENTER_CRITICAL_taskEXIT_CRITICAL.html

lt72: Just move the critical section around the update of xTaskOwner.

lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_pthread_mutex.c
```diff
@@ -297,6 +297,10 @@ int pthread_mutex_unlock( pthread_mutex_t * mutex )
 
     if( iStatus == 0 )
     {
+        /* Suspend the sheduler so that
+         * mutex is unlocked AND owner is updated atomically */
+        vTaskSuspendAll();
```
huguesBouvier: Could you try to put a mutex here instead? suspendall is violent

lt72: vTaskSuspendAll( ) is less invasive than disabling interrupts and also is the only other native way in FreeRTOS to use a critical section. Adding a mutex to guard the state would be more expensive in terms of RAM, so I would prefer to go ahead with this fix for now, and reconsider later, eventually, after applying the other fixes for POSIX layer.

PR #205, opened 2019-01-10
fix: Add printing of FS security alert count to the main.c of the TI CC3220SF

sarenameas:
* This is to help with debugging possible security alerts caused by testing that are not cleared. 
* Through experimentation the alert count would print only when the routine is placed after the WiFi connection is complete.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/ti/cc3220_launchpad/common/application_code/main.c
+++ b/tests/ti/cc3220_launchpad/common/application_code/main.c
@@ -50,6 +50,7 @@
 /* TI-Driver includes. */
 #include <ti/drivers/GPIO.h>
 #include <ti/drivers/SPI.h>
+#include <ti/drivers/net/wifi/simplelink.h>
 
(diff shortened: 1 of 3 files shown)


tests/ti/cc3220_launchpad/common/application_code/main.c
```diff
@@ -383,22 +390,22 @@ void vApplicationGetTimerTaskMemory( StaticTask_t ** ppxTimerTaskTCBBuffer,
 /*-----------------------------------------------------------*/
 
 /**
- * @brief Socket event handler function.
```
sarenameas: This is no longer needed. It was deleted in the demo main.c, but forgotten in the test main.c.

alexa-noxon: whoops! good catch :)

PR #212, opened 2019-01-14
Issue template files for new incoming issues.

mradulan:
Description
-----------
Standard issue templates help debug the issue efficiently. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,8 @@
 ---
 name: Bug report
-about: Create a bug report for the project Amazon FreeRTOS
+about: Create a bug report for Amazon FreeRTOS
 labels: bug
+
 ---
 
 **Describe the bug**
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,31 +8,28 @@ labels: bug
 A clear and concise description of what the bug is.
 
 **System Information**
-- Hardware board used (Please provide the name of the exact model you are using)
+- Which hardware board or part numbers?
 - IDE used
 - Operating System [Windows|Linux|MacOS]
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
...
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**System Information**
+- Hardware board used
+- Operating System [Windows|Linux|MacOS]
+- Version of Amazon FreeRTOS (run ``git describe --tags`` to find it)
+- Demo ran [MQTT|Shadow|Greengrass|OTA|Device Defender|TCP Echo|Sub pub Demo| Simple TCP Server]
```
dcgaws: We still want to hear about Amazon FreeRTOS bugs involving hardware other than our qualified boards, and code other than our Demo and Test projects. In those cases, we may not be able to reproduce the problem, but let's consider what data we'd need in order to best be able to take action in those cases.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots or console output**
+If appropriate, please cut-and-paste the console output to explain the issue
+or add a screenshot.
```
dcgaws: We should explicitly ask for a call stack.

mradulan: Call stack would be appropriate with an IDE, although it will be difficult to provide a call stack with ESP32

.github/ISSUE_TEMPLATE/bug_report.md
mradulan: I will add a `Custom User Application` option for the user to mention the project they are using, in which case they can add the code in the section `Code to reproduce the behavior`

PR #218, opened 2019-01-15
clean cbor serializer code

qiutongs:
Description
-----------

clean cbor serializer code

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/lib/include/aws_iot_serializer.h
+++ b/lib/include/aws_iot_serializer.h
@@ -36,25 +36,29 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#define AWS_IOT_SERIALIZER_INDEFINITE_LENGTH                      0xffffffff
+#define AWS_IOT_SERIALIZER_INDEFINITE_LENGTH                       0xffffffff
 
-#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_STREAM   { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_STREAM }
+#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_STREAM    { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_STREAM }
 
-#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_MAP      { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_MAP }
+#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_MAP       { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_MAP }
 
-#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_ARRAY    { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_ARRAY }
+#define AWS_IOT_SERIALIZER_ENCODER_CONTAINER_INITIALIZER_ARRAY     { .pHandle = NULL, .type = AWS_IOT_SERIALIZER_CONTAINER_ARRAY }
+
+#define AWS_IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER              { .type = AWS_IOT_SERIALIZER_UNDEFINED, .pHandle = NULL, .value = 0 }
+
+#define AWS_IOT_SERIALIZER_DECODER_ITERATOR_INITIALIZER            NULL
 
 /* helper macro to create scalar data */
 #define AwsIotSerializer_ScalarSignedInt( signedIntValue )                                                                          \
     ( AwsIotSerializerScalarData_t ) { .value = { .signedInt = ( signedIntValue ) }, .type = AWS_IOT_SERIALIZER_SCALAR_SIGNED_INT } \
 
-#define AwsIotSerializer_ScalarTextString( pTextStringValue )                                                                                           \
+#define AwsIotSerializer_ScalarTextString( pTextStringValue )                                                                                                                                  \
     ( AwsIotSerializerScalarData_t ) { .value = { .pString = ( ( uint8_t * ) pTextStringValue ), .stringLength = strlen( pTextStringValue ) }, .type = AWS_IOT_SERIALIZER_SCALAR_TEXT_STRING } \
 
 #define AwsIotSerializer_ScalarByteString( pByteStringValue, length )                                                                                            \
     ( AwsIotSerializerScalarData_t ) { .value = { .pString = ( pByteStringValue ), .stringLength = ( length ) }, .type = AWS_IOT_SERIALIZER_SCALAR_BYTE_STRING } \
 
-/* Determin if the data type is a container. */
+/* Determine if the data type is a container. */
 #define AwsIotSerializer_IsContainer( dataType )                                                                                                                      \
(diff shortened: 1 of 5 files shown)


lib/include/aws_iot_serializer.h
```diff
@@ -49,11 +49,15 @@
     ( AwsIotSerializerScalarData_t ) { .value = { .signedInt = ( signedIntValue ) }, .type = AWS_IOT_SERIALIZER_SCALAR_SIGNED_INT } \
 
 #define AwsIotSerializer_ScalarTextString( pTextStringValue )                                                                                           \
-    ( AwsIotSerializerScalarData_t ) { .value = { .pString = ( ( uint8_t * ) pTextStringValue ), .stringLength = 0 }, .type = AWS_IOT_SERIALIZER_SCALAR_TEXT_STRING } \
+    ( AwsIotSerializerScalarData_t ) { .value = { .pString = ( ( uint8_t * ) pTextStringValue ), .stringLength = strlen( pTextStringValue ) }, .type = AWS_IOT_SERIALIZER_SCALAR_TEXT_STRING } \
 
 #define AwsIotSerializer_ScalarByteString( pByteStringValue, length )                                                                                            \
     ( AwsIotSerializerScalarData_t ) { .value = { .pString = ( pByteStringValue ), .stringLength = ( length ) }, .type = AWS_IOT_SERIALIZER_SCALAR_BYTE_STRING } \
 
+/* Determin if the data type is a container. */
```
ravibhagavandas: Type "Determine"

lib/serializer/cbor/aws_iot_serializer_tinycbor_encoder.c
```diff
@@ -145,19 +177,23 @@ static AwsIotSerializerError_t _openContainer( AwsIotSerializerEncoderObject_t *
...
-    AwsIotSerializerError_t returnError = AWS_IOT_SERIALIZER_INTERNAL_FAILURE;
-    CborError cborError = CborUnknownError;
+    /* New object must be a container. */
+    if( !AwsIotSerializer_IsContainer( pNewEncoderObject->type ) )
+    {
+        return AWS_IOT_SERIALIZER_INVALID_INPUT;
+    }
 
-    CborEncoder * pOuterEncoder = _castEncoderObjectToCborEncoder( pEncoderObject );
+    AwsIotSerializerError_t returnedError = AWS_IOT_SERIALIZER_SUCCESS;
```
ravibhagavandas: Declarations are at the beginning of the function as per the coding standard.

lib/serializer/cbor/aws_iot_serializer_tinycbor_encoder.c
```diff
@@ -169,25 +205,23 @@ static AwsIotSerializerError_t _openContainer( AwsIotSerializerEncoderObject_t *
                 break;
 
             case AWS_IOT_SERIALIZER_CONTAINER_STREAM:
-                /* no action is required. */
+                /* TODO: figure out what it means to open stream container. */
```
ravibhagavandas: As per CBOR standard, data can be either map array or a sequence of scalar values/containers ( which is a stream ). So in my opinion we shouldn't support opening a stream container within another container. Thoughts ?

lib/include/aws_iot_serializer.h
qiutongs: fixed

lib/serializer/cbor/aws_iot_serializer_tinycbor_encoder.c
qiutongs: I wonder if this standard can be loosen a little bit. I found it more readable in some cases...

lib/serializer/cbor/aws_iot_serializer_tinycbor_encoder.c
qiutongs: I agree. Let me enforce it.

lib/include/aws_iot_serializer.h
```diff
@@ -303,34 +314,34 @@ typedef struct AwsIotSerializerDecodeInterface
...
-   * Pretty format and print the buffer.
-   * @param[in] pDataBuffer Pointer to the buffer
-   * @param[in] dataSize Size of the buffer.
-   */
+    /**
+     * Pretty format and print the buffer.
+     * @param[in] pDataBuffer Pointer to the buffer
+     * @param[in] dataSize Size of the buffer.
+     */
     void ( * print )( uint8_t * pDataBuffer,
```
ravibhagavandas: I guess we are no longer using this function. Could we remove it ?

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
```diff
@@ -23,382 +23,453 @@
...
+            case AWS_IOT_SERIALIZER_SCALAR_SIGNED_INT:
+               {
+                   int i = 0;
+                   cborError = cbor_value_get_int( pCborValue, &i );
+
+                   /* TODO: assert no error on _createDecoderObject */
+
+                   pDecoderObject->value = ( AwsIotSerializerScalarValue_t )
+                   {
+                       .signedInt = i
```
ravibhagavandas: We could just do - pDecoderObject->value.signedInt = i

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
```diff
@@ -23,382 +23,453 @@
...
-        pCborDecoder->isOutermost = true;
-        _createDecoderObject(pCborDecoder, pDecoderObject);
+        pCborValueWrapper->isOutermost = true;
+
+        returnedError = _createDecoderObject( pCborValueWrapper, pDecoderObject );
     }
-    else
+
+    /* If there is any error or decoder object is a scalar type, free the cbor resources. */
+    if( cborError || returnedError || !_isArrayOrMap( pDecoderObject->type ) )
```
ravibhagavandas: CBOR buffer can be a single data item as well ( for example a CBOR integer ). Here we limit to either array or MAP
https://en.wikipedia.org/wiki/CBOR

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
```diff
@@ -23,382 +23,453 @@
...
+    _translateErrorCode( cborError, &returnedError );
+
+    return returnedError;
 }
 
 /*-----------------------------------------------------------*/
 
-static AwsIotSerializerError_t _get(AwsIotSerializerDecoderIterator_t iterator,
-    AwsIotSerializerDecoderObject_t * pValueObject)
+static AwsIotSerializerError_t _destroy( AwsIotSerializerDecoderObject_t * pDecoderObject )
```
ravibhagavandas: As discussed lets change to void return type?

lib/include/aws_iot_serializer.h
qiutongs: yes, let me remove

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
qiutongs: Yes, that is cleaner

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
qiutongs: yes

lib/serializer/cbor/aws_iot_serializer_tinycbor_decoder.c
qiutongs: No. I free the memory of pointer of CborValue for a different reason. When it is a scalar data, the pointer of CborValue is not saved to the "handle", thus not useful. If they are not freed here, it will be memory leak.

PR #223, opened 2019-01-18
fix: POSIX unit tests for time_t and cond var

abhidixi11:
* The tests were failing after posix timer and cond var changes
* espressif does not implement PTHREAD_COND_INITIALIZER
* so unit tests need to use pthread_cond_init()

Description
-----------

Checklist:
----------

- [ ] My code is Linted.


gordonwang0:
I think the formatting for the `#if/#endif` are incorrect, please run code formatting tool.

PR #235, opened 2019-01-22
Updates to the Quick Start script given documention feedback

sarenameas:
- For ease of use add the relative path to afr_source_dir in configure.json.
- Add support for the ~ relative home path in Unix systems. 

Tested ../.., ~, and absolute paths in Linux and Windows.

----------

- [n/a] My code is Linted.


mradulan:
Thank you for doing this. Works on MacOS too.

PR #243, opened 2019-01-24
Fix batch of fix for Nordic tests

huguesBouvier:
Fix OTA PAL test on Nordic

Description
-----------
Fix OTA PAL test on Nordic

Checklist:
----------

- [ ] My code is Linted.


lib/ota/portable/nordic/nrf52840-dk/aws_ota_pal.c
```diff
@@ -515,24 +516,50 @@ static ret_code_t prvPAL_WriteImageDescriptor( OTA_FileContext_t * const C )
...
+      {
+        ul32BitBlocksTosend = otapalSERIALIZING_ARRAY_SIZE;
+      }
+
+      if(ul32BitBlocksTosend > NRF_FICR->CODEPAGESIZE)
+      {
+        ul32BitBlocksTosend = NRF_FICR->CODEPAGESIZE;
+      }
+
+      memcpy(pulSerializingArray, pacData, ulBlockSize);
```
ravibhagavandas: Should this be done outside while loop ?

tests/nordic/nrf52840-dk/common/application_code/main.c
```diff
@@ -23,72 +23,89 @@
...
+#include "bt_hal_manager.h"
+#include "bt_hal_gatt_server.h"
+#include "aws_ble_services_init.h"
+#include "app_uart.h"
+
+/* MQTT v4 include. */
+#include "aws_iot_mqtt.h"
+
+#include <aws_ble.h>
+#include "aws_iot_network_manager.h"
```
ravibhagavandas: This header file can be removed.

tests/nordic/nrf52840-dk/common/application_code/main.c
```diff
@@ -23,72 +23,89 @@
...
 /* clang-format off */
+#define mainTEST_RUNNER_TASK_STACK_SIZE     ( configMINIMAL_STACK_SIZE * 12 )
 
 /* Logging Task Defines. */
 #define mainLOGGING_MESSAGE_QUEUE_LENGTH    ( 15 )
-#define mainLOGGING_TASK_STACK_SIZE         ( configMINIMAL_STACK_SIZE * 8 )
+#define mainLOGGING_TASK_STACK_SIZE         ( configMINIMAL_STACK_SIZE * 4 )
+
+/* BLE Lib defines. */
+#define mainBLE_DEVICE_NAME                 "nRF52840-dk"                               /**< Name of device. Will be included in the advertising data. */
```
ravibhagavandas: This could be replaced by the BLE configs defined in aws_ble_config.h

lib/ota/portable/nordic/nrf52840-dk/aws_ota_pal.c
huguesBouvier: Yes on those there is not choice. We can only copy as much as a page size or what is available in the buffer. The chunk is then copied in flash. The next chunk is copied in the next loop cycle

tests/nordic/nrf52840-dk/common/application_code/main.c
huguesBouvier: Fixed

tests/nordic/nrf52840-dk/common/application_code/main.c
huguesBouvier: good catch

PR #258, opened 2019-01-29
improve defender unit tests

qiutongs:
Description
-----------

- not do test assert in MQTT callback; instead, copy the data and assert outside the callback
- remove "retry if reject" logic; instead, simply increase the sleep time

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tests/common/defender/aws_iot_tests_defender_api.c
+++ b/tests/common/defender/aws_iot_tests_defender_api.c
@@ -43,9 +43,8 @@
 #define _WAIT_STATE_TOTAL_SECONDS             5
 
 /* Time interval for defender agent to publish metrics. It will be throttled if too frequent.
- * TODO: add "retry on throttle logic"
  */
-#define _DEFENDER_PUBLISH_INTERVAL_SECONDS    8
+#define _DEFENDER_PUBLISH_INTERVAL_SECONDS    20
 
 /* Define a decoder based on chosen format. */
 #if AWS_IOT_DEFENDER_FORMAT == AWS_IOT_DEFENDER_FORMAT_CBOR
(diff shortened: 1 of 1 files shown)


tests/common/defender/aws_iot_tests_defender_api.c
```diff
@@ -130,10 +133,13 @@ TEST_TEAR_DOWN( Full_DEFENDER )
 {
     AwsIotDefender_Stop();
 
-    if( _reportAccepted || _reportRejected )
+    /* Actually get defender callback. */
+    if( _callbackInfo.eventType != -1 )
```
yuhui-zheng: Chatted offline. It seems in this implementation, we leave throttling handling to user. To be specific, application is increasing _DEFENDER_PUBLISH_INTERVAL_SECONDS, and parsing _callbackInfo.eventType to determine what to do next. 

This may not be the best solution. Since:
1. This creates unreliable test. Throttling is usually per region per account, a fix number with single attempt would never reliably work. 
2. Inconsistent with other AWS clients. Where usually in other clients, users can input retry scheme, and max retry. (e.g. typical retry scheme -- exponential backoff delay. typical max retry -- 3.)


This review comment is more about design instead of this specific PR. Apologies.

tests/common/defender/aws_iot_tests_defender_api.c
```diff
@@ -471,16 +491,30 @@ TEST( Full_DEFENDER, SetPeriod_after_started )
...
+    if( pCallbackInfo != NULL )
+    {
+        _callbackInfo.eventType = pCallbackInfo->eventType;
+        _callbackInfo.metricsReportLength = pCallbackInfo->metricsReportLength;
+        _callbackInfo.payloadLength = pCallbackInfo->payloadLength;
 
-    _reportRejected = pCallbackInfo->eventType == AWS_IOT_DEFENDER_METRICS_REJECTED;
+        if( _callbackInfo.payloadLength > 0 )
+        {
+            _callbackInfo.pPayload = AwsIotTest_Malloc( _callbackInfo.payloadLength );
```
yuhui-zheng: line 503, 512, 533, and 534. Symmetry is beauty. Better to put allocate/free both in test setup/teardown. Or both in local test.

tests/common/defender/aws_iot_tests_defender_api.c
qiutongs: I agree with #1. And I don't have a good solution now.

#2 will not be a problem for actual user. Defender is a daemon application that will send metrics as slow as 5 minutes (unless user keeps start, stop, start, stop...)

So this problem is scoped to our testing.

tests/common/defender/aws_iot_tests_defender_api.c
qiutongs: This makes sense. I can move the allocate to setup

PR #259, opened 2019-01-29
prvCheckOptions refactoring

karkhaz:
Description
-----------
Prior to these commits, the function `prvCheckOptions` was rather long and contained two nested loops that were difficult to follow. This PR pulls each of those loops out into separate functions, `prvCheckOptions_innerLoop` and `prvCheckOptions_outerLoop`. The variables manipulated by those loops are passed in by const reference.

This makes the code easier to follow, and is in preparation for a commit introducing automated proofs of code correctness for the two loops and the overall function.

Much of this PR was originally contributed by Mark R. Tuttle.

Checklist:
----------

  - Switched on `testrunnerFULL_TCP_ENABLED` and ran in Windows Simulator, all tests passed
- [ ] My code is Linted.
  - No. I ran uncrustify and it changed the entire file, so I decided to keep consistency with the existing code. Let me know if anything needs tidying :)

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
@@ -1272,8 +1272,8 @@ static BaseType_t prvSingleStepTCPHeaderOptions( const unsigned char ** const pp
 	{
 		/* All other options have a length field, so that we easily
 		can skip past them. */
-		unsigned char len = ( *ppucPtr )[ 1 ];
-		if( ( len < 2 ) || ( len > xRemainingOptionsBytes ) )
+		unsigned char ucLen = ( *ppucPtr )[ 1 ];
+		if( ( ucLen < 2 ) || ( ucLen > xRemainingOptionsBytes ) )
 		{
 			/* If the length field is too small or too big, the options are
 			 * malformed, don't process them further.
@@ -1289,19 +1289,19 @@ static BaseType_t prvSingleStepTCPHeaderOptions( const unsigned char ** const pp
 			 */
 			if( ( *ppucPtr )[0] == TCP_OPT_SACK_A )
 			{
-				len -= 2;
+				ucLen -= 2;
 				( *ppucPtr ) += 2;
 
-				while( len >= 8 )
+				while( ucLen >= 8 )
 				{
-					prvSkipPastRemainingOptions( ppucPtr, ppxSocket, &len );
+					prvSkipPastRemainingOptions( ppucPtr, ppxSocket, &ucLen );
 				}
-				/* len should be 0 by now. */
+				/* ucLen should be 0 by now. */
 			}
 		}
 		#endif	/* ipconfigUSE_TCP_WIN == 1 */
 
-		( *ppucPtr ) += len;
+		( *ppucPtr ) += ucLen;
 	}
 	return xRet;
 }
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
@@ -1182,168 +1182,171 @@ BaseType_t should_continue_loop;
 
 static BaseType_t prvSingleStepTCPHeaderOptions( const unsigned char ** const pucPtr, const unsigned char ** const pucLast, FreeRTOS_Socket_t ** const pxSocket, TCPWindow_t ** const pxTCPWindow)
 {
-		UBaseType_t uxNewMSS;
-		UBaseType_t xRemainingOptionsBytes = ( *pucLast ) - ( *pucPtr );
+	UBaseType_t uxNewMSS;
+	UBaseType_t xRemainingOptionsBytes = ( *pucLast ) - ( *pucPtr );
 
-		if( ( *pucPtr )[ 0 ] == TCP_OPT_END )
-		{
-			/* End of options. */
-			return pdFALSE;
-		}
-		if( ( *pucPtr )[ 0 ] == TCP_OPT_NOOP)
-		{
-			/* NOP option, inserted to make the length a multiple of 4. */
-			( *pucPtr )++;
-			return pdTRUE;
-		}
-
-		/* Any other well-formed option must be at least two bytes: the option
-		type byte followed by a length byte. */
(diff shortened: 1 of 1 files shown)


karkhaz:
Ping @tautschnig @markrtuttle

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
```diff
@@ -1158,161 +1172,179 @@ UBaseType_t uxNewMSS;
...
 
-		if( pucPtr[ 0 ] == TCP_OPT_END )
-		{
-			/* End of options. */
-			break;
-		}
-		if( pucPtr[ 0 ] == TCP_OPT_NOOP)
+/*-----------------------------------------------------------*/
+
+static FORCE_INLINE LoopAction_t prvCheckOptions_outerLoop( const unsigned char ** const pucPtr, const unsigned char ** const pucLast, FreeRTOS_Socket_t ** const pxSocket, TCPWindow_t ** const pxTCPWindow)
```
dcgaws: It would be helpful to leave a comment about what the outer and inner loops are doing from a protocol perspective. Further, changing these respective function names to reflect that would be more self-documenting. 

I think that outer loop is walking the TCP header options field (i.e. which can consist of zero or more options of fixed and variable length). And I think that the inner loop is specific to parsing SACK.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
```diff
@@ -1158,161 +1172,179 @@ UBaseType_t uxNewMSS;
...
+	}
+	return LOOP_CONTINUE;
+}
+
+
+/*-----------------------------------------------------------*/
+
+static FORCE_INLINE void prvCheckOptions_innerLoop( const unsigned char ** const pucPtr, FreeRTOS_Socket_t ** const pxSocket, unsigned char * const pucLen)
+{
+uint32_t ulFirst = ulChar2u32( ( *pucPtr ) );
```
dcgaws: Assumes that pucPtr is long enough to hold a valid SACK. Maybe at least say that in a comment? Better, since you're passing the length anyway, add an explicit check. Yes, the latter is arguably redundant, but this is the kind of code that calls for paranoia.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
```diff
@@ -227,6 +227,20 @@ static BaseType_t prvTCPPrepareConnect( FreeRTOS_Socket_t *pxSocket );
  */
 static void prvCheckOptions( FreeRTOS_Socket_t *pxSocket, NetworkBufferDescriptor_t *pxNetworkBuffer );
 
+/*
+ * The body of the outer loop in prvCheckOptions, factored out to aid code
+ * comprehension and enable separate property-checking for the loop body and the
+ * rest of the function.
+ */
+static FORCE_INLINE LoopAction_t prvCheckOptions_outerLoop( const unsigned char ** const pucPtr, const unsigned char ** const pucLast, FreeRTOS_Socket_t ** const pxSocket, TCPWindow_t ** const pxTCPWindow);
```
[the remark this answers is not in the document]
karkhaz: Done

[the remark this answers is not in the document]
karkhaz: Done

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
karkhaz: Done, please make sure I'm not saying anything daft in the comments or function names

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
karkhaz: Agreed, but could I leave this for a separate PR? Two reasons:
- This PR contains only non-functional changes, I think it's cleaner to keep it that way
- If I add new assertions, CBMC will check them :) I'd like to get the proofs running on the code we already have before I start sprinkling assertions everywhere, it would be useful to start from a known-good state (where Mark left off)

sarenameas:
\bot run checks

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_IP.c
```diff
@@ -1158,162 +1171,183 @@ UBaseType_t uxNewMSS;
...
+
+static BaseType_t prvSingleStepTCPHeaderOptions( const unsigned char ** const pucPtr, const unsigned char ** const pucLast, FreeRTOS_Socket_t ** const pxSocket, TCPWindow_t ** const pxTCPWindow)
+{
+	UBaseType_t uxNewMSS;
+	UBaseType_t xRemainingOptionsBytes = ( *pucLast ) - ( *pucPtr );
+
+	if( ( *pucPtr )[ 0 ] == TCP_OPT_END )
+	{
+		/* End of options. */
+		return pdFALSE;
```
alexa-noxon: The FreeRTOS docs say that we are using a single-point-of-return and that there are only 2 deviations. (https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html) 

I think we should consider refactoring to maintain the single point of return, or discuss if this requires an update to documentation if we opt for replacing breaks with return pdFALSE.

karkhaz:
@alexa-noxon thank you for your review, I made the following changes:
* Pointer-to-pointer parameters are now prefixed with `pp`
* `should_continue_loop` is now called `xShouldContinueLoop`
* `prvSingleStepTCPHeaderOptions` now has a variable called `xRet`, returned at the end of the function and initially set to `pdTrue`. Currently, it gets re-set to `pdTrue` by sections of code that want to continue the loop, for readability; hopefully the compiler is smart enough to optimize this out.
* I also noticed two new code style issues: there was a variable called `len` (I renamed it to `ucLen`), and it was being declared at the beginning of an `else` block (we now declare it at the beginning of the function).

karkhaz:
All tests are passing on the simulator.

karkhaz:
@alexa-noxon @dcgaws cheers. I don't have write access, would you mind merging?

karkhaz:
Closing, as per an updated CI strategy discussion.

PR #267, opened 2019-01-30
update tinycbor to latest version

qiutongs:
Description
-----------

Update tinycbor (an unknown old version in 2016) to latest one: https://github.com/intel/tinycbor

I copy all the files in https://github.com/intel/tinycbor/tree/master/src

I test on windows desktop:
- FULL_SHADOW
- FULL_OTA_CBOR
- FULL_OTA_AGENT
- FULL_OTA_PAL

Checklist:
----------

- [ ] My code is Linted.


sarenameas:
We need to run the tests on all boards that support OTA. As I recall, we have made our own custom updates for our set of supported compilers in lib\third_party\tinycbor\compilersupport_p.h.

lib/third_party/tinycbor/compilersupport_p.h
```diff
@@ -77,14 +80,18 @@
...
+#endif
+
 #ifndef __has_builtin
 #  define __has_builtin(x)  0
 #endif
 
-
-/* Disable this optimization for TI ARM compiler v18 or higher because it has some issues
-with these intrinsics. */
-#if !defined(__TI_COMPILER_VERSION__) || __TI_COMPILER_VERSION__ < 18000000
```
sarenameas: Deleting this intrinsic might cause issues on the TI platform so we should definitely run tests on TI (and all other platforms).

qiutongs: Good catch. I blindly deleted all the old files. Let me try to run OTA E2E test on TI and see what happens

PR #273, opened 2019-02-01
add tests/demo for multiple socket connections

qiutongs:
Description
-----------

- add tests/demo for multiple socket connections: it relies on echo server.
- fix bug in defender collector when copying metrics data
- get IoT ip address in tests for verification
- refactor tests verification

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tests/common/defender/aws_iot_tests_defender_api.c
+++ b/tests/common/defender/aws_iot_tests_defender_api.c
@@ -30,6 +30,8 @@
 /* Defender internal includes. */
 #include "aws_iot_defender_internal.h"
 
+#include "aws_secure_sockets.h"
+
 #include "unity_fixture.h"
 
 #include "FreeRTOS_POSIX/unistd.h"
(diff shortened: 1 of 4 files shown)


tests/common/defender/aws_iot_tests_defender_api.c
```diff
@@ -546,48 +631,29 @@ static void _waitForMetricsAccepted( uint32_t timeoutSec )
     uint32_t maxIterations = timeoutSec / _WAIT_STATE_INTERVAL_SECONDS;
     uint32_t iter = 1;
 
-    /* Wait for an valid event type to be set. */
-    while( _callbackInfo.eventType != AWS_IOT_DEFENDER_METRICS_ACCEPTED )
+    /* Wait for an event type to be set. */
+    while( _callbackInfo.eventType == -1 )
```
lt72: Is there an ENUM value for -1?  better than a magic number...

qiutongs: No, it is a magic number. I am not sure what is the best thing to do here. we can deal with it later

PR #274, opened 2019-02-04
 Fix FreeRTOS_setsockopt()

yuhui-zheng:
Description
-----------
bug fix to "ARG FV Finding: FreeRTOS_setsockopt() returns wrong error codes".
ticket number: E039512548

Please check with current on-call for status update. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_Sockets.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_Sockets.c
@@ -1481,8 +1481,17 @@ FreeRTOS_Socket_t *pxSocket;
 					}
 
 					pxProps = ( ( WinProperties_t * ) pvOptionValue );
-					FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_SNDBUF, &( pxProps->lTxBufSize ), sizeof( pxProps->lTxBufSize ) );
-					FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_RCVBUF, &( pxProps->lRxBufSize ), sizeof( pxProps->lRxBufSize ) );
+
+					if ( FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_SNDBUF, &( pxProps->lTxBufSize ), sizeof( pxProps->lTxBufSize ) ) != 0 )
+					{
+						break;	/* will return -pdFREERTOS_ERRNO_EINVAL */
+					}
+
+					if ( FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_RCVBUF, &( pxProps->lRxBufSize ), sizeof( pxProps->lRxBufSize ) ) != 0 )
+					{
+						break;	/* will return -pdFREERTOS_ERRNO_EINVAL */
+					}
+
 					#if( ipconfigUSE_TCP_WIN == 1 )
 					{
 						pxSocket->u.xTCP.uxRxWinSize = ( uint32_t )pxProps->lRxWinSize;	/* Fixed value: size of the TCP reception window */


lib/FreeRTOS-Plus-TCP/source/FreeRTOS_Sockets.c
```diff
@@ -1481,8 +1482,19 @@ FreeRTOS_Socket_t *pxSocket;
 					}
 
 					pxProps = ( ( WinProperties_t * ) pvOptionValue );
-					FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_SNDBUF, &( pxProps->lTxBufSize ), sizeof( pxProps->lTxBufSize ) );
-					FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_RCVBUF, &( pxProps->lRxBufSize ), sizeof( pxProps->lRxBufSize ) );
+
+					xSetBufStatus = FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_SNDBUF, &( pxProps->lTxBufSize ), sizeof( pxProps->lTxBufSize ) );
```
danielsn: Would it be cleaner to just directly test the result code?

```
					if (FreeRTOS_setsockopt( xSocket, 0, FREERTOS_SO_SNDBUF, &( pxProps->lTxBufSize ), sizeof( pxProps->lTxBufSize ) ) != 0 )
					{
						break;	/* will return -pdFREERTOS_ERRNO_EINVAL */
					}

danielsn: Also, why check against 0 instead of PD_FAIL?

dcgaws: I'm a fan of APIs that bubble the lowest level error code up to the app. That makes things easier to debug. In this case, I'll admit that FreeRTOS_setsockopt isn't yet using sufficiently granular error codes for that approach to make much of a difference. But that can be improved gradually.

htibosch: I do not entirely understand the above patch:
It removes the call that sets `FREERTOS_SO_RCVBUF`, and it will check the return code for setting `FREERTOS_SO_SNDBUF`.

The setsockopt call can fail for two reasons:

- The socket is not a TCP socket
- The caller is too late because the buffer has been created already

Both errors are important to notice, but they can easily be avoided.

When creating a server socket, the TX/RX buffer sizes **must** be set by the parent socket. 
Every child socket will inherit the values which were set for their parent socket (`FREERTOS_SO_RCVBUF` and `FREERTOS_SO_SNDBUF`).

When a socket actively connects, the parameters must be set for this socket before connecting.

yuhui-zheng: To @danielsn's first comment, agreed test return code directly looks cleaner. The local variable is mainly to help aid debugging, and hope to get optimized away in prod build. Regardless, will update the CR.

And about the if condition, the logic was to make reading easier. I agree SOCKETS_ERROR_NONE would be better in a different context. Yet all other switch-case branches are directly returning 0 upon successful execution, which makes SOCKETS_ERROR_NONE causes unnecessary confusion.

sarenameas:
Current status:
Passed:
**espressif** - https://amazon-freertos-ci.corp.amazon.com/job/ESP/job/esp32/job/esp-linux-1/47/
**winsim** - https://amazon-freertos-ci.corp.amazon.com/job/pc/job/windows/job/sim-win-2/25/
**microchip** - https://amazon-freertos-ci.corp.amazon.com/job/mchip/job/pic32/job/win-2/15/
    - Failed baseline's intermittently failing test  AFQP_SECURE_SOCKETS_Threadsafe_DifferentSocketsDifferentTasks, so this project is still green for this PR.

Issue #214, opened 2019-01-15
vLoggingPrint is not scheduler-safe


dcgaws:
In 'aws_demo_logging.c', there is the following function:

    void vLoggingPrint( const char * pcMessage )
    {
        printf( "%s", pcMessage );
        fflush(stdout);
    }

The functions printf() and fflush() are system calls, and they may hold-up the task ( put it to sleep ). When that happens, a next task may ( will ) become active in an improper way.

Hence after a while, it may happen that several FreeRTOS tasks are running simultaneously (!). The synchronicity will get totally lost.

We solved this problem by starting an extra (non-FreeRTOS) thread which takes care of the logging. That thread ( prvWin32LoggingThread ) may do system calls. It may not call any of the FreeRTOS API's though.

Communication between FreeRTOS and a non-FreeRTOS task may seem a bit primitive: non of the mechanisms may be used between the tasks, neither Windows, nor FreeRTOS API's.

The above code will most probably disturb tests that need multiple FreeRTOS tasks.

You can either change:

void vLoggingPrint( const char * pcMessage )
{
    vLoggingPrintf( "%s", pcMessage );
}

or, if you really don't want the time-stamps and so, we need to make a new logging function.

sarenameas:
I have took this specific issue from #285 and it is in the code base: 61319c37179da9f1904fedbf4020bec5c802ab08

PR #264, opened 2019-01-30
Make vLoggingPrint scheduler-safe

htibosch:
Make vLoggingPrint compatible with FreeRTOS tasks.
-----------

For a detailed description of the problem, see [issue 124](https://github.com/aws/amazon-freertos/issues/214)

With this patch, both `vLoggingPrint` and `vLoggingPrintf` can be used from a FreeRTOS task.
The latter will mostly be used in the Windows self-test application.

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [x] My code is Linted.

PR #285, opened 2019-02-07
Fix/afr tcp

dcgaws:
Fixes for github issue #214; improving FreeRTOS+TCP handling of RST packets; overly-conservative buffer length checks in DNS response parsing; MSVC compiler warning in POSIX.

### DIFF
--- a/demos/pc/windows/common/application_code/aws_demo_logging.c
+++ b/demos/pc/windows/common/application_code/aws_demo_logging.c
@@ -113,6 +113,12 @@ static DWORD WINAPI prvWin32LoggingThread( void * pvParam );
 static void prvCreatePrintSocket( void * pvParameter1,
                                   uint32_t ulParameter2 );
 
+/*
+ * Write a messages to stdout, either with or without a time-stamp.
+ * A Windows thread will finally call printf() and fflush().
+ */
+static void prvLoggingPrintf( BaseType_t xFormatted, const char *pcFormat, va_list xArgs );
+
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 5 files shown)


htibosch:
About aws_demo_logging.c :

Dan, I see that you took over two of my proposed PR's, thanks for that!

However, I doubt whether this is what was intended:

    void vLoggingPrint( const char * pcMessage )
    {
        vLoggingPrintf( pcMessage );
    }

There are two problems:

1) The call to vLoggingPrintf() may lead to a crash in case 'pcMessage' contains a percentage sign.

This would solve it:

        vLoggingPrintf( "%s", pcMessage );

2) vLoggingPrint() is used to print messages in the Unity tests: those are bare messages without the time-stamp and the task name.

With your change, the output of the tests will look like this:

39 24534 [TestRunner] TEST(Full_FREERTOS_TCP, prvParseDnsResponse) PASS
40 24634 [TestRunner] TEST(Full_FREERTOS_TCP, ulDNSHandlePacket) PASS
41 24734 [TestRunner] TEST(Full_FREERTOS_TCP, prvCheckOptions) PASS
...

I'm not sure if the testers like these time-stamps to each line.

In [PR #264](https://github.com/aws/amazon-freertos/pull/264) I proposed a
vLoggingPrint() that is safe and that doesn't insert the prefixes:

TEST(Full_FREERTOS_TCP, prvParseDnsResponse) PASS
TEST(Full_FREERTOS_TCP, ulDNSHandlePacket) PASS
TEST(Full_FREERTOS_TCP, prvCheckOptions) PASS

Hein

Issue #275, opened 2019-02-04
TLS send erroneous hard error


mbruno-xmos:
During a TLS send if there isn't a free send buffer and the socket is non-blocking an erroneous hard error is triggered and the TLS context is freed. This is the case when publishing messages with the MQTT agent.

See:
https://github.com/aws/amazon-freertos/blob/29b654cfa76ad4eca2d11e186f19118c028f0cef/lib/tls/aws_tls.c#L799-L816

mbedtls_ssl_write() does not return 0 when no bytes are sent, but rather -pdFREERTOS_ERRNO_ENOSPC. I have modified the code as follows and this seems to have fixed the problem:
```
            else if( 0 == xResult || -pdFREERTOS_ERRNO_ENOSPC == xResult )
            {
                /* No data sent (and no error). The secure sockets
                 * API supports non-blocking send, so stop the loop but don't
                 * flag an error. */
                xResult = 0;
                break;
            }
```

mradulan:
Hi @mbruno-xmos , did you encounter this while running the PubSub demo or a custom application

mbruno-xmos:
A custom application. We are publishing larger (~4kbyte) messages multiple times per second.

mbruno-xmos:
I have also modified aws_mqtt_agent.c to set the socket to block on sends with a timeout of mqttconfigTCP_SEND_TIMEOUT_MS so that it does not sit in a tight loop in prvMQTTSendCallback(). I'm not sure why it isn't already done that way.

mradulan:
Hi @mbruno-xmos , we will be adding the fix for the condition when there is no space available. About adding the timeout in prvMQTTSendCallback, we will have to go over that condition internally again. Thank you for the fix!

mradulan:
The issue has been fixed. Please reopen the ticket if the problem persists.

PR #288, opened 2019-02-07
Why: TLS_Send does not handle the error condition when space is not available on disk.

mradulan:
TLS_Send does not handle the error condition when space is not available on disk.

Related Issues: https://github.com/aws/amazon-freertos/issues/275

Description
-----------
mbedtls_write sends back pdFREERTOS_ERRNO_ENOSPC when the send buffer is full.

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/lib/tls/aws_tls.c
+++ b/lib/tls/aws_tls.c
@@ -280,9 +280,9 @@ static int prvPrivateKeySigningCallback( void * pvContext,
                                          size_t xHashLen,
                                          unsigned char * pucSig,
                                          size_t * pxSigLen,
-                                         int ( *piRng )( void *,
-                                                         unsigned char *,
-                                                         size_t ), /*lint !e955 This parameter is unused. */
+                                         int ( * piRng )( void *,
+                                                          unsigned char *,
+                                                          size_t ), /*lint !e955 This parameter is unused. */
                                          void * pvRng )
 {
     BaseType_t xResult = 0;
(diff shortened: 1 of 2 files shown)


lib/tls/aws_tls.c
```diff
@@ -801,7 +801,7 @@ BaseType_t TLS_Send( void * pvContext,
                 /* Sent data, so update the tally and keep looping. */
                 xWritten += ( size_t ) xResult;
             }
-            else if( 0 == xResult )
+            else if( ( 0 == xResult ) || ( -pdFREERTOS_ERRNO_ENOSPC == xResult ) )
             {
                 /* No data sent (and no error). The secure sockets
```
dcgaws: Can you please remove the "(and no error)" text in the comment?

lib/tls/aws_tls.c
```diff
@@ -801,7 +801,7 @@ BaseType_t TLS_Send( void * pvContext,
                 /* Sent data, so update the tally and keep looping. */
                 xWritten += ( size_t ) xResult;
             }
-            else if( 0 == xResult )
+            else if( ( 0 == xResult ) || ( -pdFREERTOS_ERRNO_ENOSPC == xResult ) )
```
dcgaws: Since we're not treating the ENOSPC condition as an error here, should we set xResult back to zero?

mradulan: ya that is correct, so that we set the bytes sent as 0.

PR #293, opened 2019-02-09
FreeRTOS_DNS.c : Add DNS response transaction ID check

htibosch:
FreeRTOS_DNS.c : Add DNS response transaction ID check.

Description
-----------
This PR will replace [PR #146](https://github.com/aws/amazon-freertos/pull/146).

When receiving DNS replies, the identifier will be compare to the outstanding identifiers. Only when there is a match, the DNS cache will be updated.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
@@ -1,5 +1,5 @@
 /*
- * FreeRTOS+TCP V2.0.8
+ * FreeRTOS+TCP V2.0.10
  * Copyright (C) 2017 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
@@ -373,10 +373,11 @@ typedef struct xDNSAnswerRecord DNSAnswerRecord_t;
 	/*-----------------------------------------------------------*/
 
 	/* A DNS reply was received, see if there is any matching entry and
-	call the handler. */
-	static void vDNSDoCallback( TickType_t xIdentifier, const char *pcName, uint32_t ulIPAddress );
-	static void vDNSDoCallback( TickType_t xIdentifier, const char *pcName, uint32_t ulIPAddress )
+	call the handler.  Returns pdTRUE if xIdentifier was recognised. */
+	static BaseType_t xDNSDoCallback( TickType_t xIdentifier, const char *pcName, uint32_t ulIPAddress );
+	static BaseType_t xDNSDoCallback( TickType_t xIdentifier, const char *pcName, uint32_t ulIPAddress )
 	{
+		BaseType_t xResult = pdFALSE;
 		const ListItem_t *pxIterator;
 		const MiniListItem_t* xEnd = ( const MiniListItem_t* )listGET_END_MARKER( &xCallbackList );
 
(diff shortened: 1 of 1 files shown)


lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -446,7 +449,13 @@ TickType_t xIdentifier = 0;
 	/* Generate a unique identifier. */
 	if( 0 == ulIPAddress )
 	{
-		xIdentifier = ( TickType_t )ipconfigRAND32( );
+		/* DNS identifiers are 16-bit. */
+		xIdentifier = ( TickType_t )( ipconfigRAND32( ) & 0xffff );
+		/* An identifier with value 0 will be treated as: "no identifier". */
```
dcgaws: We unfortunately have an ambiguity wherein zero is both a valid random number and a possible error return. Fixing that is a pain, since ipconfigRAND32 tends to be #defined to a port specific implementation.

In any case, we don't want to issue a request with a non-random random transaction ID.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -809,25 +824,14 @@ static uint8_t *prvSkipNameField( uint8_t *pucByte, size_t xSourceLen )
...
-
-	if( pxNetworkBuffer->xDataLength > sizeof( UDPPacket_t ) )
-	{
-		prvParseDNSReply( pucUDPPayloadBuffer,
-			xPlayloadBufferLength,
-			( uint32_t )pxDNSMessageHeader->usIdentifier );
-	}
+	prvParseDNSReply( ( uint8_t * )pxDNSMessageHeader,
+					  pxNetworkBuffer->xDataLength,
+					  ( uint32_t )0u );
```
dcgaws: What would be the rationale for returning a DNS result to the app when the packet didn't match our (cryptographically random) transaction ID? Seems to me that it's not just a question of whether we cache the result.

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
htibosch: Elsewhere in the library, the result of `ipconfigRAND32()` is tested against 0 with the following comment ( not mine ) :

`	xDHCPData.ulTransactionId = ipconfigRAND32();`

`	/* Check for random number generator API failure. */`
`	if( 0 != xDHCPData.ulTransactionId )`
`	{`

Today I **did** have a complaint that DHCP didn't work. The user's implementation of `ipconfigRAND32` returned zero at the first call.

( I always emphasize to users that the first value of the seed must be random )

We can not require that `ipconfigRAND32()` never returns 0, or let things go wrong when ocasionally 0 is returned.

It is possible to call `ipconfigRAND32()` a few times until it returns a non-zero value?

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
htibosch: The above call is done for packets that come in asynchronously. `prvParseDNSReply()` will look-up the transaction ID in the list of outstanding DNS requests. Only if a match is found, the result will be added to the DNS cache.

htibosch:
I have created a new PR #860 that replaces this one. This PR had too many conflicts with the even changing repository.
This PR will be closed now.

PR #296, opened 2019-02-11
fix defender demo build for TI

qiutongs:
Description
-----------

fix defender demo build for TI

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/ti/cc3220_launchpad/common/config_files/aws_iot_config.h
+++ b/demos/ti/cc3220_launchpad/common/config_files/aws_iot_config.h
@@ -26,10 +26,12 @@
 #define _AWS_IOT_CONFIG_H_
 
 /* Library logging configuration. */
-#define AWS_IOT_GLOBAL_LOG_LEVEL      AWS_IOT_LOG_INFO
-#define AWS_IOT_PLATFORM_LOG_LEVEL    AWS_IOT_LOG_INFO
-#define AWS_IOT_NETWORK_LOG_LEVEL     AWS_IOT_LOG_INFO
-#define AWS_IOT_MQTT_LOG_LEVEL        AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_GLOBAL      AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_PLATFORM    AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_NETWORK     AWS_IOT_LOG_INFO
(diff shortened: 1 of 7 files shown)


lib/defender/aws_iot_defender_collector.c
```diff
@@ -230,6 +230,7 @@ static void _serialize()
                                                         : assertSuccess;
 
     uint8_t metricsGroupCount = 0;
```
lt72: I did not see this before... should never use anything other than a size_t (uint32_t) for an index. 8 bit quantityt does not really save anything, since CPU will have to load value in a register. And It may create stack disalignment...

demos/ti/cc3220_launchpad/common/config_files/aws_iot_config.h
```diff
@@ -26,10 +26,12 @@
...
 /* Library logging configuration. */
-#define AWS_IOT_GLOBAL_LOG_LEVEL      AWS_IOT_LOG_INFO
-#define AWS_IOT_PLATFORM_LOG_LEVEL    AWS_IOT_LOG_INFO
-#define AWS_IOT_NETWORK_LOG_LEVEL     AWS_IOT_LOG_INFO
-#define AWS_IOT_MQTT_LOG_LEVEL        AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_GLOBAL      AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_PLATFORM    AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_NETWORK     AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_MQTT        AWS_IOT_LOG_INFO
+#define AWS_IOT_LOG_LEVEL_DEFENDER    AWS_IOT_LOG_DEBUG
```
lt72: Is DEBUG the correct level for a demo? Maybe INFO is better?

lib/defender/aws_iot_defender_collector.c
```diff
@@ -358,6 +359,7 @@ static void tcpConnectionsCallback( void * param1,
     IotMetricsTcpConnection_t * pConnection = NULL;
 
     size_t total = IotListDouble_Count( pTcpConnectionsMetricsList );
+    uint8_t i = 0;
```
lt72: Same... use size_t or  uint32_t to count...

htibosch: Or use `BaseType_t` if `FreeRTOS.h` is included. `BaseType_t` promises that the width is equal to the size of a register.

Issue #304, opened 2019-02-13
Improve feedback for Shadow Client Failure


tombresson:
### Problem:

During implementation of the Shadow Client, the initial push of shadow data would indicate "Timed out". This clearly was not the case in the backend, as the data would actually make it to AWS (as it was seen appear on the Device Shadow). 

The real issue was, JSMN was not allocated enough tokens and was failing to tokenize the response that is received on the `shadow/update/accepted` topic. The tokenization is done in the the MQTT callback, but the JSMN feedback is not passed back to the Shadow Client. 

The semaphore wait times out (see relevant lines linked below) long after the MQTT Client has called back to `prvShadowUpdateCallback` (through a series of callback function calls) and subsequently failed to give the semaphore (lines also linked below) indicating that the update was actually received. The description of the header related to the `SHADOW_Update` function don't give indication of this being a possible failure case. Funny enough, the JSMN errors have been defined in the header (`aws_shadow.h`) further up in the file.

https://github.com/aws/amazon-freertos/blob/c542bafb75b774ec3f4b3e2de17f909169654123/lib/shadow/aws_shadow.c#L1242-L1249

https://github.com/aws/amazon-freertos/blob/c542bafb75b774ec3f4b3e2de17f909169654123/lib/shadow/aws_shadow.c#L1026-L1061

### Describe the solution you would like:
The feedback for the failure case could be much more indicative of the specific failure instead of simply "timed out".

WRansohoff:
Hi Tom,

Thank you for the feedback and detailed summary - if I'm understanding you right, it sounds like our error reporting logic in the Shadow client is missing the type of error that your application ran into.

Looking at the code snippet you linked to, it seems like the Shadow client reports a 'timeout' when it cannot acquire the relevant semaphore, and I can definitely see that sounding confusing without more details about why the semaphore was never released; it sounds a lot like the client is reporting a network timeout.

Thanks for reporting this, we want our error messages to be as helpful as possible. We'll look into how this message is phrased, and whether we can catch the errors you describe earlier to provide a clearer picture of what went wrong.

tombresson:
Hi @WRansohoff,

I think the key takeaway is, in the Shadow Client lib:

* Semaphore is taken by the application call into the Shadow Client
* The message is received on the `/update/accepted` or `/update/rejected` topic
* Calback fails to tokenize the payload received on the topic **because there are too few JSMN tokens allocated in the data array.**
[...]

PR #312, opened 2019-02-15
Update shadow error message in response to issue #304

WRansohoff:
Our Shadow client has an error message which reports a 'timeout' in reference to trying to acquire a semaphore handle. But that sounds a lot like a network timeout in the context of MQTT applications; we could be a bit clearer.

Description
-----------
Update the error message to clarify that the application appears to have encountered an error in the callback method which the application was waiting on, inferred from the fact that its callback never released its semaphore.

Checklist:
----------

- [X] My code is Linted.


WRansohoff:
We were also missing debug information around an error which someone experienced; the case where we did not allocate enough tokens for the JSMN library to parse a JSON document; I added another message for that.

PR #318, opened 2019-02-17
Add support for Wireshark TLS packet decryption to aws_tls.c

sarenameas:
This change supports writing the CLIENT_RANDOM symmetric session keys to SSLKEYLOGFILE.
SSLKEYLOGFILE is supported by Chrome, Firefox, and Openssl. This file allows Wireshark to decrypt TLS traffic. 
You can find instructions on how to enable this feature on Windows here: https://jimshaver.net/2015/02/11/decrypting-tls-browser-traffic-with-wireshark-the-easy-way/

This feature is enabled only when MBEDTLS_DEBUG_C is enabled.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/lib/tls/aws_tls.c
+++ b/lib/tls/aws_tls.c
@@ -52,6 +52,16 @@
 #include <time.h>
 #include <stdio.h>
 
+#ifdef MBEDTLS_DEBUG_C
+    /* Support for writing to SSLKEYLOGFILE for Wireshark TLS decryption. */
+    typedef struct {
+        FILE * xKeyLogFile;
+        int lInClientRandom : 1;
+        int lInMasterSecret : 1;
+        int lHexdumpLinesToProces;
+    } TLSKeyLogDebugContext_t;
+#endif
+
 /**
  * @brief Internal context structure.
  *
@@ -91,6 +101,9 @@ typedef struct TLSContext
     mbedtls_x509_crt xMbedX509Cli;
     mbedtls_pk_context xMbedPkCtx;
     mbedtls_pk_info_t xMbedPkInfo;
+    #ifdef MBEDTLS_DEBUG_C
+        TLSKeyLogDebugContext_t xKeyLogDebugCtx;
+    #endif
 
     /* PKCS#11. */
     CK_FUNCTION_LIST_PTR xP11FunctionList;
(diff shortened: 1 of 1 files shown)


lib/tls/aws_tls.c
```diff
@@ -675,7 +771,37 @@ BaseType_t TLS_Connect( void * pvContext )
...
+        /* Wireshark debugging support on the Windows Simulator.
+         * When https://github.com/ARMmbed/mbedtls/issues/1504 is implemented consider
+         * deleting this code. To give proper credit, this code is adapted from:
+         * https://github.com/Lekensteyn/mbedtls/commit/68aea15. */
+        #ifdef WIN32
+            {
+                /* On windows an environment variable value has a maximum size
+                 * of 32,627 characters, including the null-terminating character. */
+                char cKeyLogFileName[ 32767 ];
+                unsigned int ulReturnValue = GetEnvironmentVariable(
```
dcgaws: I don't think it's safe to make Win32 API calls from a FreeRTOS task (i.e. in WinSim). The problem is that the Win32 thread scheduler and the FreeRTOS task scheduler don't know about each other. For example, it is reasonably common for an API call such as this to (a) take the loader lock and (b) block the thread, at least the first time it's called. Since the FreeRTOS scheduler doesn't know that this app task just got blocked, it no longer has coherent state. I also suspect that this could deadlock. 

I think our only option for supporting this feature is via the logging task.

lib/tls/aws_tls.c
```diff
@@ -675,7 +771,37 @@ BaseType_t TLS_Connect( void * pvContext )
...
+
+        /* Wireshark debugging support on the Windows Simulator.
+         * When https://github.com/ARMmbed/mbedtls/issues/1504 is implemented consider
+         * deleting this code. To give proper credit, this code is adapted from:
+         * https://github.com/Lekensteyn/mbedtls/commit/68aea15. */
+        #ifdef WIN32
+            {
+                /* On windows an environment variable value has a maximum size
+                 * of 32,627 characters, including the null-terminating character. */
+                char cKeyLogFileName[ 32767 ];
```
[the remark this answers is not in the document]
sarenameas: Thanks!

lib/tls/aws_tls.c
sarenameas: Or perhaps I can grab the environment variable in main.c? Or I can add it as a TLS config parameter?

sarenameas:
This will be reworked when: https://github.com/aws/amazon-freertos/pull/394 goes through.

PR #319, opened 2019-02-17
Update Windows Simulator to use AWS Logging Task Dynamic Buffers.

sarenameas:
The Windows Simulator is updated to use the same logging task as the rest of the platforms for consistency. 
aws_logging_task_dynamic_buffers.c also facilitates printing a large variable sized buffer easier.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/pc/windows/common/application_code/main.c
+++ b/demos/pc/windows/common/application_code/main.c
@@ -50,7 +50,7 @@
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"
 #include "FreeRTOS_DHCP.h"
-#include "aws_demo_logging.h"
+#include "aws_logging_task.h"
 #include "aws_system_init.h"
 #include "aws_demo_runner.h"
 #include "aws_application_version.h"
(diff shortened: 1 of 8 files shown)


htibosch:
Hi, very good to start using standard logging task for the Windows project as well.

One objection though:

`    /* Map the logging task's printf to the board specific output function. */`
`    #define configPRINT_STRING( x )    printf( x );`

The system API `print()` may not be called in a FreeRTOS task, only in a Windows thread.
And a Windows thread may not call any API from FreeRTOS, like `xQueueReceive()`.

That is the reason why in `aws_demo_logging.c` a logging thread is started and not a task.
The communication between a thread and a task is not so easy because they don't have common things like semaphores or queues.

What if you do call `printf()` and/or `fflush()` from a task? The tasking mechanism may get corrupted: 2 or more task may become active at the same time.

Hein

htibosch:
For the Windows implementation, I would create a thread and use a stream buffer. There will be a single giver ( `prvLoggingTask()` ), and a singer receiver ( the Windows thread ), so no locking mechanism is needed.

The Windows thread can safely call `print()` and `fflush()` for every string received.

htibosch:
@Sarena

htibosch:
Hi Sarena, I know you're a very busy person. But did the above message reach you and was it clear what I meant to say?

sarenameas:
> Hi Sarena, I know you're a very busy person. But did the above message reach you and was it clear what I meant to say?
Hello! Thanks for the valuable information. The above messages were clear. I apologize for not ack'ing earlier. With this PR and the other I opened around the same time I realized the hazards of using Windows system calls in FreeRTOS threads. Thanks!

htibosch:
> I realized the hazards of using Windows system calls in FreeRTOS threads

Ok, good.

One detail about the Windows logging thread: as you can not wake it up from a FreeRTOS task, the thread will have to poll every N msec if there is something to print.
That is not elegant, but as this is a project running on Windows, it shouldn't be a problem.

sarenameas:
Closed for rework.

PR #320, opened 2019-02-17
Fix: priority

kwadrat:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


sarenameas:
Thanks for the PR. This will be updated in the FreeRTOS Kernel distribution.

Issue #353, opened 2019-02-27
Packet and/or TCP/IP stack corruption


marjon-halter:
**Describe the bug**
TCP Echo demo runs on repeat, but after about 1000++ successful packets, an assert happens saying the sent packet is not the same as the received back packet. Getting packet corruption and/or memory corruption.
The failure is not always at the exact same place of execution, but the assert failure is always the same one.
```cpp
/* Compare the transmitted string to the received string. */
configASSERT( strncmp( pcReceivedString, pcTransmittedString, xTransmitted ) == 0 );
```

**System information**
- Which hardware board or part numbers? ESP-WROVER-KIT V4.1 
- IDE used - Eclipse + IDF programming scripts
- Operating System [Linux]
- Version of Amazon FreeRTOS : v1.4.6-69-g3b189e55
- Project [TCP Echo]
- If your project is a Custom Application, please add the relevant code snippet in the section `Code to reproduce the bug`. 

**Expected behavior**
Device runs the TCP Echo demo to a local TCP Echo Server, sleeps, wakes up, repeat.

**Screenshots or console output**
If appropriate, please cut-and-paste the console output to explain the issue. If possible, include the call stack. 

```cpp
5943 95004 [Echo0] Received correct string from echo server.
5944 95004 [Echo0] Sending TxRx message number 1658 of length 1659 to echo server
5945 95004 [Echo0] Received correct string from echo server.
5946 95005 [Echo0] Sending TxRx message number 1659 of length 1660 to echo server
5947 95087 [Echo0] Received correct string from echo server.
5948 95087 [Echo0] Sending TxRx message number 1660 of length 1661 to echo server
5949 95187 [Echo0] Received correct string from echo server.
5950 95187 [Echo0] Sending TxRx message number 1661 of length 1662 to echo server
5951 95499 [Echo0] Timed out receiving from echo server
5952 95499 [Echo0] Received correct string from echo server.
5953 95499 [Echo0] Sending TxRx message number 1662 of length 1663 to echo server
/home/user/workspace/amazon-freertos/demos/common/tcp/aws_tcp_echo_client_single_task.c:323 (prvEchoClientTask)- assert failed!
abort() was called at PC 0x400e1003 on core 0
0x400e1003: prvEchoClientTask at /home/user/workspace/amazon-freertos/demos/common/tcp/aws_tcp_echo_client_single_task.c:323 (discriminator 1)


[...]

PR #354, opened 2019-02-27
aws_tcp_echo_client_single_task.c : a time-out is not fatal

htibosch:
aws_tcp_echo_client_single_task.c : a time-out is not fatal

Description
-----------
When the reception of data times out, and at least 1 byte has been received, an ASSERT will be thrown because the contents does not match.
When the number of bytes received is less than expected, it would be nicer to just consider this as a failure. The link may have failed, the remote party is too slow, etc.

Only when the expected number of bytes was received, but the received string does not match, this should be a fatal error ( calling `configASSERT()` ).

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/common/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/common/tcp/aws_tcp_echo_client_single_task.c
@@ -307,7 +307,7 @@ static void prvEchoClientTask( void * pvParameters )
                 /* If an error occurred it will be latched in xReceivedBytes,
                  * otherwise xReceived bytes will be just that - the number of
                  * bytes received from the echo server. */
-                if( xReceivedBytes > 0 )
+                if( xReceivedBytes == xTransmitted )
                 {
                     /* Compare the transmitted string to the received string. */
                     configASSERT( strncmp( pcReceivedString, pcTransmittedString, xTransmitted ) == 0 );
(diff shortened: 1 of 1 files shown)


htibosch:
This PR is a follow up to this [bug report](https://github.com/aws/amazon-freertos/issues/353)
I could replicate the findings and the patch will avoid that a fatal error is thrown: it's just a time-out

dcgaws:
Hein, here's the diff for the formatting that causing the Code style check hook to fail. In summary, spaces are required inside the parentheses of your casts, and the left-most paren of the first cast needs to be in the same column as the opening quote of "ERROR in the line above it. Can you please apply those changes to this branch? Thank you.

The staged files are not formatted correctly.
--- "a/demos/common/tcp/aws_tcp_echo_client_single_task.c"
+++ "b/demos/common/tcp/aws_tcp_echo_client_single_task.c"
@@ -340,7 +340,7 @@ static void prvEchoClientTask( void * pvParameters )
                      * length. */
                     ulTxRxFailures[ xInstance ]++;
                     configPRINTF( ( "ERROR: xTransmitted %d xReceivedBytes %d \r\n",
-						(int)xTransmitted, (int)xReceivedBytes ) );
+                                    ( int ) xTransmitted, ( int ) xReceivedBytes ) );
                     /* Timed out without receiving anything? */
                     break;
                 }

demos/common/tcp/aws_tcp_echo_client_single_task.c
```diff
@@ -338,7 +338,7 @@ static void prvEchoClientTask( void * pvParameters )
                      * length. */
                     ulTxRxFailures[ xInstance ]++;
                     configPRINTF( ( "ERROR: xTransmitted %d xReceivedBytes %d \r\n",
-						(int)xTransmitted, (int)xReceivedBytes ) );
+                                    (int)xTransmitted, (int)xReceivedBytes ) );
```
dcgaws: This update is still failing the style checks because we we need the spaces inside the parentheses on the casts.

htibosch:
I will close this PR because it will be followed up by PR #861, which doesn't have so many merge conflicts. Hein.

PR #355, opened 2019-02-27
Report the correct WIN size

htibosch:
Report the correct WIN size

Description
-----------
A FreeRTOS+TCP user Wolfgang Schur, reported that under some rare circumstances, the reported value of WIN ( the available space in the TCP r4ecption window) is not correct.
In fact, it would be a bit behind: the last read from a socket would not be taken into account, so the reported WIN size sometimes wat too small.

The WIN size would be stored in the TCP socket field called `ulRxCurWinSize`.

In this patch, the field `ulRxCurWinSize` will be removed, and WIN will be calculated at the moment it must be send to the peer.

Under certain circumstances, this patch will make the communication a bit more efficient.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
+++ b/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_IP_Private.h
@@ -577,7 +577,6 @@ BaseType_t xIPIsNetworkTaskReady( void );
 			FOnConnected_t pxHandleConnected;	/* Actually type: typedef void (* FOnConnected_t) (Socket_t xSocket, BaseType_t ulConnected ); */
 		#endif /* ipconfigUSE_CALLBACKS */
 		uint32_t ulWindowSize;		/* Current Window size advertised by peer */
-		uint32_t ulRxCurWinSize;	/* Constantly changing: this is the current size available for data reception */
 		size_t uxRxWinSize;	/* Fixed value: size of the TCP reception window */
 		size_t uxTxWinSize;	/* Fixed value: size of the TCP transmit window */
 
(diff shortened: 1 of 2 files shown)


htibosch:
In the previous version, `ulOutstanding` would become non-zero in case there are bytes missing: the difference between the highest received (`ulHighestSequenceNumber`) and the highest acknowledged (`ulCurrentSequenceNumber`).
I believe that was not correct. The reported "Win" (windows size) always begins at "Ack".
I tested the change on both an STM32F and on a Xilinx platform.

dcgaws:
More detail about the proposed change, as well as related lab testing:

The problem was that WIN is derived from a variable 'xTCP.ulRxCurWinSize', That variable is only calculated within 'prvSendData()'. The variable is used ( being read ) in 'prvTCPReturnPacket()' in order to report the current WIN size. There are 5 functions that can call 'prvTCPReturnPacket()' :

    prvSendData()  ( OK )
    xTCPSocketCheck()
    prvTCPSendPacket()
    prvTCPSendRepeated()
    prvTCPPrepareConnect()

When prvTCPReturnPacket() is not called by prvSendData(), the value of 'xTCP.ulRxCurWinSize' may run behind. Lab testing shows that happening form prvTCPSendRepeated().

It is proving difficult to write test code that shows the merits of the patch unconditionally. There is a lot of randomness in the flow of TCP communication: the occupation rate of the network, the quality of the transport layers, speed of the sender, speed of the receiver. For example, sending 10 MB of data through a LAN to a FreeRTOS+TCP device, it might happen 10 times that the reported WIN size is not running behind. This is causing unnecessary delays.

The other problem with the current code is : the WIN size has to start with 'ulCurrentSequenceNumber' (ACK), and not with 'ulHighestSequenceNumber'. This was not correct:

    pxSocket->u.xTCP.ulRxCurWinSize = pxTCPWindow->xSize.ulRxWindowLength -
        ( pxTCPWindow->rx.ulHighestSequenceNumber - pxTCPWindow->rx.ulCurrentSequenceNumber );

The difference only becomes apparent when there is data missing in the reception window, and so it will affect connections that run through a bad connection.

PR #358, opened 2019-02-28
Convert errors in TLS wrapper to negative error codes

alexa-noxon:
TLS library states that it returns negative error codes, but PKCS #11
library returns positive error codes.  Some positive PKCS #11 error codes
were propogated out of TLS, breaking API contract and occasionally causing
error code to be interpreted as number of bytes sent or received.

This commit also fixes a threading bug in the Windows PKCS #11 PAL layer
when reading files, and cleans up formatting on an error print statement in
the sockets tests.

Description
-----------
Ran TLS & Sockets tests on WinSIm for testing.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/include/aws_tls.h
+++ b/lib/include/aws_tls.h
@@ -32,6 +32,20 @@
 #endif
 
 /**
+ * @defgroup TlsErrors TLS Error Codes
+ * @brief Error codes returned by the TLS API.
+ *
+ * Note that TLS API may also propagate port-specific
+ * error codes, or codes from mbedTLS.
+ */
+/**@{ */
+#define TLS_ERROR_HANDSHAKE_FAILED    ( -2001 )   /*!< Error in handshake. */
+#define TLS_ERROR_RNG                 ( -2002 )   /*!< Error in RNG. */
+#define TLS_ERROR_SIGN                ( -2003 )   /*!< Error in sign operation. */
+
+/**@} */
+
+/**
  * @brief Defines callback type for receiving bytes from the network.
  *
  * @param[in] pvCallerContext Opaque context handle provided by caller.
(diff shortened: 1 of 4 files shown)


lib/include/aws_tls.h
```diff
@@ -31,6 +31,20 @@
...
 
+/**
+ * @defgroup TlsErrors TLS Error Codes
+ * @brief Error codes returned by the TLS API.
+ *
+ * Note that TLS API may also propagate port-specific
+ * error codes, or codes from mbedTLS.
+ */
+/**@{ */
+#define TLS_ERROR_HANDSHAKE_FAILED    ( -2001 )   /*!< Error in handshake. */
```
dcgaws: We should probably be explicit about the size and signedness of these.

lib/include/aws_tls.h
```diff
@@ -31,6 +31,20 @@
...
+#define TLS_ERROR_RNG                 ( -2002 )   /*!< Error in RNG. */
```
dcgaws: Pending the above, is the implication that the -20XX error code namespace is reserved for this library?

lib/include/aws_tls.h
alexa-noxon: As discussed, leaving without explicit type to match error codes in other Amazon FreeRTOS files.

lib/include/aws_tls.h
alexa-noxon: As discussed, there is no guarantee that we are not conflicting with a device library.   Sockets errors are smaller (0-200, -10xx), MQTT, defender use an enum (so they will be smaller), OTA uses a combination of high byte lib error + lower byte driver errors,  Crypto & GG use true/false.

PR #364, opened 2019-03-05
CBMC memory safety proofs for selected entry points in FreeRTOS TCP/IP.

markrtuttle:
CBMC memory safety proofs for selected entry points in FreeRTOS TCP/IP.

Description
-----------
This commit adds the CBMC memory safety proofs to the FreeRTOS public repository.  All CBMC proofs are contained in a top-level cbmc directory added to the source tree.  Continuous integration will recheck these proofs with each commit and pull request.  Future proofs will go into this directory.

New: The latest commits implement a single method for building and running the proofs on both Windows and Linux (now customers and developers can run and write proofs in their own development environments), and move the proofs from a top-level directory into the tools directory.

Checklist:
----------
I have run the proofs against the lastest FreeRTOS commit and all proofs pass.

- [ ] My code is Linted.


htibosch:
Hi Mark,

Every FreeRTOS+TCP socket consists of a part that is common to TCP and UDP, but also a type-specific part, which is expressed in a union:

`	union`
`	{`
`		IPUDPSocket_t xUDP;`
`		IPTCPSocket_t xTCP;`
`	}`

If you change that `union` into a `struct`. Is that only for the memory proofs, or is this a proposal to change the library?

The same about taking away the 'static' keyword in e.g. FreeRTOS_DHCP.c, is that for testing and proving only?

Hein

dcgaws:
Hein, the CBMC patches are only for running the proofs, at least for the foreseeable future.

markrtuttle:
I have pushed a new commit that cleans up the patch directory.

The patches are only for testing by CBMC.

There are now just four patches.
- One to remove static attributes from the functions under test.
- One to change a union to a struct.  This deals with a CBMC issue being fixed now and should be removed soon.
- One to change a minilistitem to a listitem.  This deals with a CBMC issue being fixed now and should be removed soon.
- One refactoring prvCheckOptions into three functions to facilitate the CBMC proof.  This will be submitted as its own PR and should be removed soon.

sarenameas:
Hi, this is great. Could we have some documentation with step by step what to install and how to run?

karkhaz:
@sarenameas we're working on it, cheers. We had a discussion offline with @alexa-noxon about what documentation would be useful, and I recently merged [#4299](https://github.com/diffblue/cbmc/pull/4299) and [#4382](https://github.com/diffblue/cbmc/pull/4382) into CBMC to make it easier for Windows users to run these proofs.

karkhaz:
We've just pushed some changes, hopefully these address everyone's suggestions. To make reviewing easier, here's what to focus on:
- We've added a [README](https://github.com/aws/amazon-freertos/pull/364/files#diff-d893d02333c3e5947431db261ef05106) describing how to install CBMC and run the proofs on your local machine.
- We're no longer driving the proofs from a hand-written makefile, since this doesn't work on both Windows and Linux. Instead, we're generating the Makefiles. We can generate Makefiles both for Linux (to be used by the CI), and for Windows (for anybody wanting to run the proofs locally using NMake).
  - [This python script](https://github.com/aws/amazon-freertos/pull/364/files#diff-c0c1c97bb2d22c759927808ed4c0dc92) generates a common Makefile from [this template](https://github.com/aws/amazon-freertos/pull/364/files#diff-e9ebb8c276ceecf4a13fe80bf585e18d), containing generic rules for running all the proofs.
  - [This python script](https://github.com/aws/amazon-freertos/pull/364/files#diff-e43da8430e5ab85f6bc6e347e9927699) generates Makefiles for each of the proofs, containing proof-specific definitions, including what assumptions we're making about the sizes of data structures etc. Those proof-specific Makefiles get generated from commented JSON files like [this one](https://github.com/aws/amazon-freertos/pull/364/files#diff-323de7e13f773571846346e0cc62dbb8). There is one JSON file for each proof. The prolog of the python script describes the JSON format.
- The actual proofs (the `.c` and `.h` files) are unchanged since last time. We've moved the entire `cbmc` directory under `tools`.

alexa-noxon:
Thanks for moving into tools and updating the docs!

PR #369, opened 2019-03-06
CBMC refactoring of prvCheckOptions

markrtuttle:
Refactor prvCheckOptions for CBMC memory safety proof.

Description
-----------
This pull request depends on pull request https://github.com/aws/amazon-freertos/pull/364.  Do not merge this pull request until that pull request has been merged.

This pull request was discussed and accepted by dcgaws and alexa-noxon (but never merged) as https://github.com/aws/amazon-freertos/pull/259 (now closed).

This pull request refactors the prvCheckOptions function into three functions to facilitate its proof of memory safety by CBMC.  The second and third functions represent the bodies of the outer and inner loops of prvCheckOptions.

This pull request violates the coding guideline that a function should only have one return statement: The function prvSingleStepTCPHeaderOptions returns a boolean from multiple locations (instead of setting a boolean and returning the boolean at the bottom of the function).  The reason is that this function is the body of the outer loop in the original prvCheckOptions, and the boolean pdTRUE or pdFALSE corresponds to the statement continue or break in the original outer loop indicating whether the outer loop should continue another iteration or not.  

It would be possible to set a boolean instead of immediately returning, and to return that boolean at the bottom of the function.  But that would require setting and checking another boolean to guard against executing any of the following code in the function, and doing this in six places.  That can be done, but the code is not a clean as just returning.

Checklist:
----------
The CBMC proofs of memory safety succeed.
- [ ] My code is Linted.


markrtuttle:
This pull request passes the tests in tests\pc\windows\visual_studio\aws_tests.sln.

markrtuttle:
This is a fairly old pull request that refactors the TCP/IP CheckOptions function to make it possible to prove CheckOptions memory safety with CBMC.  

We need to decide if we are going to do this or not.  

Kareem is working on adding to CBMC the ability to check loop invariants, which would make this refactoring unnecessary, but that is in the future. I think this refactoring makes the code easier to read, although it can be improved to get rid of some of the call-by-reference in the refactoring.  

We are currently checking CheckOptions by applying a patch to the source tree that does the refactoring done in this pull request.   That patch is conflicting a bit with the work Hein is doing.  If we don't accept merge this pull request, maybe we should just drop the patch and drop CheckOptions from the set of CBMC proofs we are running.

PR #371, opened 2019-03-07
Atomic support for FreeRTOS.

yuhui-zheng:
Atomic support for FreeRTOS

Description
-----------
Two implementations are provided in this header file:
1. Disabling interrupts globally to achieve atomic operation. (available to all ports, default option.)
2. Using ISA native atomic instruction. (available to ports using GCC compiler only.)

To use, user needs to include ```FreeRTOS.h``` and ```portmacro.h``` before including ```atomic.h```. To choose which atomic implementation to use, user could set/clear ```configUSE_ATOMIC_INSTRUCTION``` in ```FreeRTOSConfig.h```.

API tests are done in personal fork, instead of this repo. See here
https://github.com/yuhui-zheng/amazon-freertos/blob/feature/perf_counter_st/demos/st/stm32l475_discovery/common/application_code/atomic_tryout.c. 
ST ac6 project is configured to work with atomic examples. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/include/atomic.h
+++ b/lib/include/atomic.h
@@ -71,16 +71,20 @@ extern "C" {
     /* Port specific definitions --
      * how to enter/exit critical section from atomic.
      * Refer template -- ./lib/FreeRTOS/portable/Compiler/Arch/portmacro.h
+     *
+     * Every call to ATOMIC_EXIT_CRITICAL() must be closely paired with
+     * ATOMIC_ENTER_CRITICAL().
      */
     #if defined( portSET_INTERRUPT_MASK_FROM_ISR )
+
         /* Nested interrupt scheme is supported in this port. */
-        #define ATOMIC_ENTER_CRITICAL()     portSET_INTERRUPT_MASK_FROM_ISR()
-        #define ATOMIC_EXIT_CRITICAL( x )   portCLEAR_INTERRUPT_MASK_FROM_ISR( x )
-        typedef UBaseType_t CriticalSectionType_t;
+        #define ATOMIC_ENTER_CRITICAL()     UBaseType_t uxCriticalSectionType = portSET_INTERRUPT_MASK_FROM_ISR()
+        #define ATOMIC_EXIT_CRITICAL()      portCLEAR_INTERRUPT_MASK_FROM_ISR( uxCriticalSectionType )
     #else
+
+        /* Nested interrupt scheme is NOT supported in this port. */
         #define ATOMIC_ENTER_CRITICAL()     portENTER_CRITICAL()
-        #define ATOMIC_EXIT_CRITICAL( x )   portEXIT_CRITICAL()
-        typedef void CriticalSectionType_t;
+        #define ATOMIC_EXIT_CRITICAL()      portEXIT_CRITICAL()
     #endif /* portSET_INTERRUPT_MASK_FROM_ISR() */
 
     /* Port specific definitions --
@@ -114,9 +118,8 @@ static portFORCE_INLINE bool Atomic_CompareAndSwap_u32( uint32_t volatile * pDes
 #else
 
     bool bExecutionStatus = false;
-    CriticalSectionType_t temp;
 
-    temp = ATOMIC_ENTER_CRITICAL( );
+    ATOMIC_ENTER_CRITICAL();
 
     if ( *pDestination == ulComparand )
     {
@@ -124,7 +127,7 @@ static portFORCE_INLINE bool Atomic_CompareAndSwap_u32( uint32_t volatile * pDes
         bExecutionStatus = true;
     }
 
-    ATOMIC_EXIT_CRITICAL( temp );
+    ATOMIC_EXIT_CRITICAL();
 
     return bExecutionStatus;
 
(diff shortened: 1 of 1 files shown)


lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+        #define ATOMIC_ENTER_CRITICAL()     portSET_INTERRUPT_MASK_FROM_ISR()
+        #define ATOMIC_EXIT_CRITICAL( x )   portCLEAR_INTERRUPT_MASK_FROM_ISR( x )
+        typedef UBaseType_t CriticalSessionType_t;
+    #else
+        #define ATOMIC_ENTER_CRITICAL()     portENTER_CRITICAL()
+        #define ATOMIC_EXIT_CRITICAL( x )   portEXIT_CRITICAL()
+        typedef void CriticalSessionType_t;
+    #endif /* portSET_INTERRUPT_MASK_FROM_ISR() */
+
+    /* Port specific definitions --
```
yuhui-zheng: Currently known port which doesn't have portFORCE_INLINE defined -- WinSim

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+
+/* Standard includes. */
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if defined ( configUSE_ATOMIC_INSTRUCTION ) && ( configUSE_ATOMIC_INSTRUCTION == 1 )
```
lt72: What 'configUSE_ATOMIC_INSTRUCTION' really does is to use the compiler builtin implemenation. 
Should we change the name to state that fact more clearly?
e.g. configUSE_BUILTIN_ATOMICS...

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+
+    /* This branch is for GCC compiler and GCC compiler only. */
+    #ifndef portFORCE_INLINE
+        #define portFORCE_INLINE  inline __attribute__((always_inline))
```
lt72: __attribute__((always_inline)) is specific to GCC, maybe clang... but I see nowhere a check that states we are using GCC or clang...

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+ *
+ * @return bool True for swapped, false for not swapped.
+ *
+ * @note This function only swaps *pDestination with ulExchange, if previous *pDestination value equals ulComparand.
+ */
+static portFORCE_INLINE bool Atomic_CompareAndSwap_u32( uint32_t volatile * pDestination, uint32_t ulExchange, uint32_t ulComparand )
+{
+
+#if defined ( configUSE_ATOMIC_INSTRUCTION ) && ( configUSE_ATOMIC_INSTRUCTION == 1 )
+    return __atomic_compare_exchange( pDestination, &ulComparand, &ulExchange, false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST );
```
lt72: Should w add a few words (and/or links) here to explain what the flags __ATOMIC_xxx are?

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+
+#else
+
+    bool bExecutionStatus = false;
+    CriticalSessionType_t temp;
```
lt72: mispelling: CriticalSectionType_t  => Section, not Session

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+    temp = ATOMIC_ENTER_CRITICAL( );
+
+    if ( *ppDestination == pComparand )
+    {
+        *ppDestination = pExchange;
+        bExecutionStatus = true;
+    }
+
+    ATOMIC_EXIT_CRITICAL( temp );
+    return bExecutionStatus;
```
lt72: Insert one new line

lib/include/atomic.h
```diff
@@ -0,0 +1,484 @@
...
+ * Atomic add
+ *
+ * @brief Adds count to the value of the specified 32-bit variable as an atomic operation.
+ *
+ * @param[in,out] pAddend  Pointer to memory location from where value is to be loaded and written back to.
+ * @param[in] lCount       Value to be added to *pAddend.
+ *
+ * @return previous *pAddend value.
+ */
+static portFORCE_INLINE int32_t Atomic_Add_i32( int32_t volatile * pAddend, int32_t lCount )
```
lt72: I thought we decided to do all uint32_t (unsigned) first... I believe signed may be just as good, just confirming...

yuhui-zheng: Per offline discussion, updating to uint32_t for all APIs. And it's up to user to handle overflow/underflow correctly. 

C99 standard, thanks to Cobus.
```
6.3.1.3 Signed and unsigned integers

1 When a value with integer type is converted to another integer type other than _Bool, if the value can be represented by the new type, it is unchanged.

2 Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type until the value is in the range of the new type.49)

3 Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.
```

lib/include/atomic.h
yuhui-zheng: updated to ```configUSE_GCC_BUILTIN_ATOMICS```.

lib/include/atomic.h
yuhui-zheng: My understanding is that when people start a new project, they should have some idea of what compiler and architecture combination they want to use. So that, baseline port related stuff is selected from ```./lib/FreeRTOS/portable/```. In this sense, checking what compiler people are using and branching based on that may seem a little bit redundant. 

In current implementation, by default, ```configUSE_GCC_BUILTIN_ATOMICS``` is not set, and disabling interrupt is used instead of GCC intrinsics. A power user can optionally set ```configUSE_GCC_BUILTIN_ATOMICS``` to 1 when GCC compiler is used on that port AND the port architecture supports atomic instructions.

lib/include/atomic.h
yuhui-zheng: link to gcc is given in comment section.

PR #384, opened 2019-03-14
Make TCP low-water/high-water limits configurable with a socket option

htibosch:
Make TCP low-water/high-water limits configurable with a socket option

Description
-----------
When receiving data on a TCP socket, there is a mechanism that controls the flow: when the reception stream buffer gets filled for more than 80%, a WIN update will be sent to the peer.
When the buffer drops below 20%, again a WIN update will be sent in order to resume transmission.

These 20/80 limits have been hard-coded up until now.

This patch allows to control the 2 limits per socket. It adds a new socket option:

`#define FREERTOS_SO_SET_LOW_HIGH_WATER	( 18 )`

and it should be called with a (pointer to a) parameter of this type:

`typedef struct xLOW_HIGH_WATER {`
`	/* Structure to pass for the 'FREERTOS_SO_SET_LOW_HIGH_WATER' option */`
`	size_t uxLittleSpace;	/* Send a STOP when buffer space drops below X bytes */`
`	size_t uxEnoughSpace;	/* Send a GO when buffer space grows above X bytes */`
`} LowHighWater_t;`

Note that the two variables express as a number of bytes, not a percentage.

As this is an expert option, the supplied values are not being checked on their validity.

When the socket option is not used, the defaults of 20/80 will still be applied.

Most users won't need this option, but some have found it quiet useful.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_Sockets.h
+++ b/lib/FreeRTOS-Plus-TCP/include/FreeRTOS_Sockets.h
@@ -130,6 +130,7 @@ FreeRTOS_setsockopt(). */
 	#define FREERTOS_SO_WAKEUP_CALLBACK	( 17 )
 #endif
 
+#define FREERTOS_SO_SET_LOW_HIGH_WATER	( 18 )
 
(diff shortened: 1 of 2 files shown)


htibosch:
As suggested by dcgaws, I added some basic validity checks. Also I replaced some magic values with macros. Hein

PR #391, opened 2019-03-15
Created abstraction layer for HW button.

mhattiga:
Created abstraction layer for HW button.

-----------

Abstracts HW specific information as a generic push button in the aws_iot_combined_demo.c file.  Provides a clean template for mapping the abstract HW push button functions into different versions of HW.

Modified the ~/demos/common/combined/aws_iot_combined.c that turns on/off BLE file by removing all the HW specific information. No    w BLE on/off code is only called if HW_BUTTON is defined. Now makes calls to generic functions:
  buttonInit()
  buttonWasPushed()

Added header file in ~/demos/common/include/aws_hw_button.h with:
  #define HW_BUTTON
  declarations for buttonInit() and buttonWasPushed()

Added .c file ~/demos/espressif/esp32_devkitc_esp_wrover_kit/common/ application_code/esp_wrover_sw3_button.c with definitions for buttonInit() 
and buttonWasPushed() and HW specific defines.

Checklist:
----------

- [ Y] I have tested my changes. No regression in existing tests.
- [ N] My code is Linted.

### DIFF
--- a/demos/common/combined/aws_iot_combined_demo.c
+++ b/demos/common/combined/aws_iot_combined_demo.c
@@ -67,7 +67,11 @@
 
 /* Amazon FreeRTOS OTA agent includes. */
 #include "aws_ota_agent.h"
-#include "driver/gpio.h"
+
+/* 
+ * Push button HW platform.
+ */
+#include "aws_hw_button.h"
 
 /**
  * @brief Transport types supported for MQTT Echo demo.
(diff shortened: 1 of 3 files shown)


demos/common/combined/aws_iot_combined_demo.c
```diff
@@ -67,7 +67,11 @@
 
 /* Amazon FreeRTOS OTA agent includes. */
 #include "aws_ota_agent.h"
-#include "driver/gpio.h"
+
+/* 
+ * Push button HW platform.
+ */
+#include "aws_hw_button.h"
```
mtalreja-a: As discussed, we can just leave this commented out and let users uncomment it. This way, the code will be portable for other platforms besides ESP32

mhattiga:
Should be turned off by default.

PR #398, opened 2019-03-18
Feature/ble beta afr abstractions

cobusve:
AFR Abstractions for IOT_Clock and IOT_Threads

Description
-----------
This is the first pass of the AFR abstractions. Let's review the approach first and look over the mechanisms to switch between this implementation and the POSIX implementation.

Also the stacksize and priority defaults need a better approach I would say.

Checklist:
----------

- [ ] My code is Linted.


lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+    vTaskSetTimeOutState( &xCurrentTime );
+
+    /* Adjust the tick count for the number of times a TickType_t has overflowed. */
+    ullTickCount = ( uint64_t ) ( xCurrentTime.xOverflowCount ) << ( sizeof( TickType_t ) * 8 );
+
+    /* Add the current tick count. */
+    ullTickCount += xCurrentTime.xTimeOnEntering;
+
+    /* Return the ticks converted to Milliseconds */
+    return ( int64_t ) ullTickCount * _MILLISECONDS_PER_TICK;
```
gordonwang0: Cast to `int64_t` not needed.

lib/include/platform/iot_platform_types_afr.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef _IOT_PLATFORM_TYPES_POSIX_H_
+#define _IOT_PLATFORM_TYPES_POSIX_H_
+
+#include "timers.h"
+
+
+struct iot_mutex_internal
+{
+    StaticSemaphore_t xMutex;           /**< FreeRTOS mutex. */
+    int type;                           /**< Type; used for indicating if this is reentrant or normal. */
```
gordonwang0: Maybe we can just use `bool recursive`, since that's what `IotMutex_Create` signature specifies.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,357 @@
...
+/*-----------------------------------------------------------*/
+
+bool IotSemaphore_Create( IotSemaphore_t * const pNewSemaphore,
+                             uint32_t initialValue,
+                             uint32_t maxValue )
+{
+    struct iot_sem_internal * internalSemaphore = ( struct iot_sem_internal * ) pNewSemaphore;
+    IotLogDebug( "Creating new semaphore %p.", pNewSemaphore );
+
+    if( maxValue > ( uint32_t ) SEM_VALUE_MAX )
```
gordonwang0: `SEM_VALUE_MAX` is a POSIX constant.  Probably don't have to check if `maxValue` is too big on FreeRTOS (can be any `UBaseType_t`.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+uint32_t IotSemaphore_GetCount( IotSemaphore_t * const pSemaphore )
+{
+    struct iot_sem_internal * internalSemaphore = ( struct iot_sem_internal * ) pSemaphore;
+    int count = 0;
+
+    count = ( int ) uxSemaphoreGetCount( ( SemaphoreHandle_t ) &internalSemaphore->xSemaphore );
```
gordonwang0: Returns `UBaseType_t`.

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+        }
+    }
+}
+ 
+/*-----------------------------------------------------------*/
+
+bool IotClock_TimerArm( IotTimer_t * const pTimer,
+                           uint64_t relativeTimeoutMs,
+                           uint64_t periodMs )
+{
```
lt72: Did we agree to change this to 
```
bool IotClock_TimerArm( IotTimer_t * const pTimer,
                           uint32_t relativeTimeoutMs,
                           uint32_t periodMs )
```
?

lib/include/platform/iot_platform_types_afr.h
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/**
+ * @brief The native mutex type on AFR systems.
+ */
+typedef struct iot_mutex_internal _IotSystemMutex_t;
```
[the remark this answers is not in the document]
cobusve: Yes I agree that would be more consistent. This started out in different files ...

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+
+bool IotClock_GetTimestring( char * const pBuffer,
+                                size_t bufferSize,
+                                size_t * const pTimestringLength )
+{
+    uint64_t milliSeconds = IotClock_GetTimeMs( );
+    size_t timestringLength = 0;
+
+    /* Convert the localTime struct to a string. */
+    timestringLength = snprintf( pBuffer, bufferSize, "%llu",  milliSeconds );
```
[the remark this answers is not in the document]
cobusve: I also thought that was weird but kept it the same as the posix version. If we change this one we should change them both.

lib/utils/platform/iot_clock_afr.c
cobusve: Agreed

lib/include/platform/iot_platform_types_afr.h
cobusve: Agreed. I started making all 4 types posix has and then removed 2 of them as they made less sense to have.

lib/utils/platform/iot_threads_afr.c
cobusve: Just removed the check, I think any reference to POSIX should be avoided of course.

lib/utils/platform/iot_threads_afr.c
cobusve: Indeed - fixed it

lib/utils/platform/iot_clock_afr.c
cobusve: Yes indeed. I tried to do the PR without changing the posix impl and without changing any API, to cover those in the review, but it is time now so I updated this in both implementations.

lib/include/platform/iot_platform_types_afr.h
```diff
@@ -0,0 +1,92 @@
...
+
+typedef struct _iot_sem_internal
+{
+    StaticSemaphore_t xSemaphore;       /**< FreeRTOS semaphore. */
+} _iot_sem_internal_t;
+
+/**
+ * @brief The native semaphore type on AFR systems.
+ */
+typedef _iot_sem_internal_t _IotSystemSemaphore_t;
```
cobusve: These types seem redundant, they are used in iot_platform_types.h to tie the internal semaphore struct to the specific underlying implementation via typdef proxy. I think it would be cleaner if the contract was implicit and implementors created the internal type as e.g. _IotSystemSemaphore_t directly and remove the proxy from platform_types? Wherever the type was used (IotMutex_t e.g.) it was always required that the implementation specific platform types was included in that file as well via the config file.

Opinions about the design here please

tests/common/include/iot_config_common.h
```diff
@@ -57,7 +57,8 @@
 #endif
 
 /* Use POSIX types on FreeRTOS. */
-#include "platform/iot_platform_types_posix.h"
```
cobusve: I never really liked the fact that we have to swap out header files to swap implementations. A strong abstraction would only require us to swap out .o files, but of course if you do not have dynamic allocation and your e.g. taskPool wants to contain a mutex which is statically allocated this is necessary so I kept it like this.

tests/espressif/esp32_devkitc_esp_wrover_kit/common/application_code/espressif_code/amazon-freertos-tests/component.mk
```diff
@@ -65,7 +65,9 @@ COMPONENT_OBJEXCLUDE := $(AMAZON_FREERTOS_LIB_DIR)/FreeRTOS-Plus-TCP/source/port
         $(AMAZON_FREERTOS_LIB_DIR)/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_pthread_cond.o \
         $(AMAZON_FREERTOS_LIB_DIR)/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_sched.o \
         $(AMAZON_FREERTOS_TESTS_DIR)/common/ota/aws_test_ota_cbor.o \
-	$(AMAZON_FREERTOS_LIB_DIR)/lob/ota/aws_ota_agent.o
+	$(AMAZON_FREERTOS_LIB_DIR)/ota/aws_ota_agent.o \
+        $(AMAZON_FREERTOS_LIB_DIR)/utils/platform/iot_clock_posix.o \
```
cobusve: If we do completely remove the dependency on posix for this we can remove the posix implementations as well as this exclusion. I believe we need the posix implementations in the C-SDK for people not running on FreeRTOS, but this is aFR so here having the option does not really make sense?

tests/common/utils/platform/iot_test_platform_threads.c
```diff
@@ -0,0 +1,387 @@
...
+    }
+    
+    printf ("Expected Pri = 7, actual = %d\r\n", attrData );
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief helper function for testing thread priority
+ */
+void threadStackSizeTestFunction(void * param)
```
cobusve: This test is very "iffy" as there is no easy way to get the stack size from FreeRTOS. I improvised a little but I think this is very hacky. Is there value in this test or do we just loose it? I added this test when the ESP code failed because the stack size I asked for was not what I received...

See the code in the threads_afr.c file handling defaults. When we change that to be a config parameter it is possible to create failing scenarios through configuration, which will be caught by this test.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+    int status = 0;
+    _threadInfo_t * pThreadInfo = IotThreads_Malloc(sizeof(_threadInfo_t));
+
+    if ( pThreadInfo == NULL)
+    {
+        IotLogDebug( "Unable to allocate memory for threadRoutine %p.", threadRoutine );
+        status = 1;
+    }
+
+    /* Create the FreeRTOS task that will run the thread. */
```
cobusve: FR is careful to not use the word thread, but use Task instead. Should we update the comments here to use Task instead?

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+        /* update supplied if we require the default stack size */
+        if ( stackSize == IOT_THREAD_DEFAULT_STACK_SIZE)
+        {
+            stackSize = 3072;
+        }
+
+        /* update supplied if we require the default thread priority */
+        if ( priority == IOT_THREAD_DEFAULT_PRIORITY )
+        {
+            priority = 5;
```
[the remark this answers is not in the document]
cobusve: Also - here it is important to note that the IOT_THREAD_DEFAULT_PRIORITY is set to 0 which makes it impossible to schedule a task with lowest (idle) priority. If you tried that you would unexpectedly get a priority of 5 instead.

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+    configASSERT( pxTimer != NULL );
+
+    /* Restart the timer if it is periodic. */
+    if( pxTimer->xTimerPeriod > 0 )
+    {
+        xTimerChangePeriod( xTimerHandle, pxTimer->xTimerPeriod, 0 );
+    }
+
+    /* Call timer Callback from this task */
+    pxTimer->threadRoutine( ( void * ) pxTimer->pArgument);
```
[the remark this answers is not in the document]
cobusve: Had a long discission about this one, this order is also what FreeRTOS does, so we will keep the behavior consistent.

lib/utils/platform/iot_clock_posix.c
```diff
@@ -163,7 +163,7 @@ bool IotClock_GetTimestring( char * const pBuffer,
     bool status = true;
     const time_t unixTime = time( NULL );
     struct tm localTime = { 0 };
-    size_t timestringLength = 0;
+    int timestringLength = 0;
```
qiutongs: Is it good to use int here, instead of fixed size?

cobusve: Good catch! I should not have changed this one. In a place like this you do not really get to choose the data type, if this variable is assigned from strftime then the type should be the same as the function. I was intending to make the change which Gordon pointed out to this function in iot_clock_afr.c where it was using strnprintf, which returns int, that is why it had to be int, but strftime returns size_t, which means this must remain size_t. I will fix this by changing it back.

lib/utils/platform/iot_clock_afr.c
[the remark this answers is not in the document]
cobusve: After much debate we decided to keep this one like this as this is also the behavior implemented in FreeRTOS, so this will remain consistent. There are many pro's and con's, we had a long debate, too long to repeat here. I think staying consistent with FreeRTOS is enough reason to keep it this way.

lib/include/platform/iot_platform_types_afr.h
lt72: Your proposal makes sense to me, as it does the current implementation. As long as one or the other can be driven via cmake (if required), we are OK.  In this case, we will have one and only one implementation, so I am actually in favor of removing this extra layer.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+    /* Create the FreeRTOS task that will run the thread. */
+    if ( status == 0 )
+    {
+        pThreadInfo->threadRoutine = threadRoutine;
+        pThreadInfo->pArgument = pArgument;
+        
+        /* update supplied if we require the default stack size */
+        if ( stackSize == IOT_THREAD_DEFAULT_STACK_SIZE)
+        {
+            stackSize = 3072;
```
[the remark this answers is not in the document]
lt72: Yes, it should!  I thought it was in the config already...

lib/utils/platform/iot_threads_afr.c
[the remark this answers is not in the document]
lt72: +1, this should be configurable through config file!  No magic numbers.

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+bool IotClock_GetTimestring( char * const pBuffer,
+                                size_t bufferSize,
+                                size_t * const pTimestringLength )
+{
+    uint64_t milliSeconds = IotClock_GetTimeMs( );
```
lt72: Check parameters?

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+    return ullTickCount * _MILLISECONDS_PER_TICK;
+}
+
+/*-----------------------------------------------------------*/
+
+bool IotClock_TimerCreate( IotTimer_t * const pNewTimer,
+                           IotThreadRoutine_t expirationRoutine,
+                           void * pArgument )
+{
+    _timerInfo_t * pxTimer = ( _timerInfo_t*) pNewTimer;
```
lt72: Check parameters? or Assert? (this comment applies to all public APIs, so i will not repeat it)

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+        /* Stop the FreeRTOS timer. Because the timer is statically allocated, no call
+        * to xTimerDelete is necessary. The timer is stopped so that it's not referenced
+        * anywhere. xTimerStop will not fail when it has unlimited block time. */
+        ( void ) xTimerStop( pTimerInfo->timer, portMAX_DELAY );
+
+
+        /* Wait until the timer stop command is processed. */
+        while( xTimerIsTimerActive( pTimerInfo->timer ) == pdTRUE )
+        {
+            vTaskDelay( 100 );
```
lt72: 100? long time...  1?  are there any examples of prior art in this department?
Also it is probably better implemented with a static const variable initialized as in this example, so it clearly says how many milliseconds we are waiting for:

void vTaskFunction( void * pvParameters )
 {
 /* Block for 500ms. */
 const TickType_t xDelay = 500 / portTICK_PERIOD_MS;

     for( ;; )
     {
         /* Simply toggle the LED every 500ms, blocking between each toggle. */
         vToggleLED();
         vTaskDelay( xDelay );
     }
}

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+
+        if( xTaskCreate( _threadRoutineWrapper,
+                        "iot_thread",
+                        stackSize,
+                        pThreadInfo,
+                        priority,
+                        NULL ) != pdPASS )
+        {
+            /* Task creation failed. */
+            IotLogWarn( "Failed to create thread.\r\n" );
```
lt72: Maybe just one call?

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+{
+    _iot_sem_internal_t * internalSemaphore = ( _iot_sem_internal_t * ) pSemaphore;
+
+    IotLogDebug( "Waiting on semaphore %p.", internalSemaphore );
+
+    /* Take the semaphore using the FreeRTOS API. */
+    if( xSemaphoreTake( ( SemaphoreHandle_t ) &internalSemaphore->xSemaphore,
+                          portMAX_DELAY ) != pdTRUE )
+    {
+        IotLogWarn( "Failed to wait on semaphore %p.",
```
lt72: Should we *add* an assert after the log here? SO, on a debug build, the developer will definitely realize there is an issue.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+
+/*-----------------------------------------------------------*/
+
+bool IotSemaphore_TryWait( IotSemaphore_t * const pSemaphore )
+{
+    _iot_sem_internal_t * internalSemaphore = ( _iot_sem_internal_t * ) pSemaphore;
+
+    IotLogDebug( "Attempting to wait on semaphore %p.", internalSemaphore );
+
+    return IotSemaphore_TimedWait( pSemaphore, 0 );
```
lt72: tail cal... generally speaking for debuggability purposes, kernel and related libraries favor a pattern such as the following:


     error = IotSemaphore_TimedWait( pSemaphore, 0 );

if(error is an error)
{
   ...log...
   ...assert... (in this case we should assert (see comment above) because this should NOT fail)
}

return error;

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+bool IotSemaphore_TimedWait( IotSemaphore_t * const pSemaphore,
+                                uint64_t timeoutMs )
```
lt72: timeout is relative, so it should be a 32 bit quantity!  All timeout that are relative should be 32 bits.
This is an interface change, so other code will have to change too...

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+{
+    _iot_sem_internal_t * internalSemaphore = ( _iot_sem_internal_t * ) pSemaphore;
+
+    /* Take the semaphore using the FreeRTOS API. */
+    if( xSemaphoreTake( ( SemaphoreHandle_t ) &internalSemaphore->xSemaphore,
+                          timeoutMs * configTICK_RATE_HZ / 1000 ) != pdTRUE )
```
lt72: Is there an overflow or underflow lurking around here? 

timeoutMs * configTICK_RATE_HZ / 1000

lib/include/platform/iot_platform_types_afr.h
gordonwang0: The extra layer is required because we need to keep all headers in lib/include/platform platform-independent.

As noted above, the current system is:
```c
typedef pthread_mutex_t _IotSystemMutex_t; // platform specific file in platform/include
typedef _IotSystemMutex_t IotMutex_t; // platform independent file in lib/include
```
So, we can't do
```c
typedef pthread_mutex_t IotMutex_t
```
because it breaks the platform-independent nature of iot_threads.h

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file iot_clock_afr.c
+ * @brief Implementation of the functions in iot_clock.h for Amazon FreeRTOS systems.
+ */
+
+/* Build using a config header, if provided. */
```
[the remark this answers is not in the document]
gordonwang0: It's possible to define everything with a compiler command line option like `-D` (though not recommended).  And I think some people may want to build with the defaults (libraries should provide default values for undefined configuration constants).  Though it might not make sense for the platform abstractions.

lib/utils/platform/iot_threads_afr.c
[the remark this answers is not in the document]
gordonwang0: We generally represent all stack sizes as a multiple of `configMINIMAL_STACK_SIZE` to better portability.

lib/utils/platform/iot_threads_afr.c
lt72: Please check all public abstractions API for other instances of uint64_t timeouts...

tests/espressif/esp32_devkitc_esp_wrover_kit/common/application_code/espressif_code/amazon-freertos-tests/component.mk
gordonwang0: Agreed.  Remove the POSIX implementations.

lib/utils/platform/iot_threads_afr.c
gordonwang0: In the C SDK, I am careful to always say "thread" over "task".  Not sure which is better.

lib/utils/platform/iot_clock_afr.c
```diff
@@ -0,0 +1,215 @@
...
+#endif
+
+#define _LIBRARY_LOG_NAME    ( "CLOCK" )
+#include "iot_logging_setup.h"
+
+/*-----------------------------------------------------------*/
+/*
+ * Time conversion constants.
+ */
+#define _NANOSECONDS_PER_SECOND         ( 1000000000 ) /**< @brief Nanoseconds per second. */
```
gordonwang0: Do you use these constants?  If not, consider removing them.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+    #define IotThreads_Free    free
+#endif
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @defgroup Mutex types.
+ */
+/**@{ */
+#ifndef IOT_THREAD_MUTEX_NORMAL
```
gordonwang0: These appear to be POSIX constants; consider removing them.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+
+/*-----------------------------------------------------------*/
+
+bool Iot_CreateDetachedThread(  IotThreadRoutine_t threadRoutine,
+                                void * pArgument,
+                                int32_t priority,
+                                size_t stackSize )
+{
+    IotLogDebug( "Creating new thread." );
+    int status = 0;
```
gordonwang0: I think it would be simpler (and more consistent with other code) to initialize:

`bool status = true;`
and on failure:
`status = false;`
and return:
`return status;`

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+        if( xTaskCreate( _threadRoutineWrapper,
+                        "iot_thread",
+                        stackSize,
+                        pThreadInfo,
+                        priority,
+                        NULL ) != pdPASS )
+        {
+            /* Task creation failed. */
+            IotLogWarn( "Failed to create thread.\r\n" );
+            IotLogWarn( "stackSize = %d\r\npriority = %d" , stackSize, priority);
```
gordonwang0: This seems like an extra log message?

Also the `IotLog` functions automatically put a newline at the end of their message.  So no need to do `\r\n` like in the previous log function call.

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+uint32_t IotSemaphore_GetCount( IotSemaphore_t * const pSemaphore )
+{
+    _iot_sem_internal_t * internalSemaphore = ( _iot_sem_internal_t * ) pSemaphore;
+    UBaseType_t count = 0;
+
+    count = ( UBaseType_t ) uxSemaphoreGetCount( ( SemaphoreHandle_t ) &internalSemaphore->xSemaphore );
```
gordonwang0: No need to case to `UBaseType_t` here (return value is of matching type).

lib/utils/platform/iot_threads_afr.c
```diff
@@ -0,0 +1,350 @@
...
+
+/*-----------------------------------------------------------*/
+
+void IotSemaphore_Post( IotSemaphore_t * const pSemaphore )
+{
+    _iot_sem_internal_t * internalSemaphore = ( _iot_sem_internal_t * ) pSemaphore;
+
+    IotLogDebug( "Posting to semaphore %p.", internalSemaphore );
+    /* Give the semaphore using the FreeRTOS API. */
+    int result =  xSemaphoreGive( ( SemaphoreHandle_t ) &internalSemaphore->xSemaphore );
```
gordonwang0: `xSemaphoreGive` returns `BaseType_t`, not int.

lib/include/platform/iot_platform_types_afr.h
cobusve: You are saying tht lib/include should know about _IotSystemMutex_t, this can only be true if the file in lib/include is including the file from platform/include, which it does via the config file.

But this means that if platform/include were to define `typedef pthread_mutex_t IotMutex_t` and if we were to not have _IotSystemMutex_t at all you would have the same properties with one fewer type?

I see no value added by _IotSystemMutex_t here.

lib/utils/platform/iot_clock_afr.c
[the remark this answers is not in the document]
cobusve: If we were careful and ONLY had defines in the config file this would be true, but we have let some type definitions sneak into the config files. It is not possible to define types using -D, so that is not quite true any longer... 

Platform types are now included from the header and I spotted some other types in there as well. We should fix this either way.

lib/utils/platform/iot_threads_afr.c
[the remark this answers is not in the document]
cobusve: For ESP it is defined as CONFIG_ESP32_PTHREAD_TASK_STACK_SIZE_DEFAULT in the SDK config. For others it is in FreeRTOSConfig.h as PTHREAD_STACK_MIN (PTHREAD specific name). Perhaps ESP is just the odd one out?

lib/utils/platform/iot_threads_afr.c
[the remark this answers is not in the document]
cobusve: Current default thread priority (when using PThreads) is different on ESP of course (where it is 5), and on FreeRTOS+Posix it is 
( ( uint16_t ) tskIDLE_PRIORITY & pthreadSCHED_PRIORITY_MASK) | ( PTHREAD_CREATE_JOINABLE << pthreadDETACH_STATE_SHIFT ) which is best I can tell (0 & 0x7FFF) | (1 << 15), which seems extremely high.

lib/utils/platform/iot_threads_afr.c
cobusve: Richard writes this in "Mastering the FreeRTOS™ Real Time Kernel"

A Note About Terminology
In FreeRTOS, each thread of execution is called a ‘task’. There is no consensus on
terminology within the embedded community, but I prefer ‘task’ to ‘thread,’ as thread can have
a more specific meaning in some fields of application.

lib/utils/platform/iot_clock_afr.c
cobusve: Yes! I kept the checks and behavior identical to the posix implementation. We should add checks to both, probably assert on both pointers is the best we can do here? Agree?

Also this function prints time as Ms since startup, while the posix one uses strftime, to here is a very different behavior, we discussed this before, but I wanted to point it out in review.

lib/utils/platform/iot_clock_afr.c
cobusve: Agreed.

lib/utils/platform/iot_clock_afr.c
cobusve: This loop is intended to block forever, or until the timer is no longer active. I think making it 100 was probably a mistake, in the posix implementation it has vTaskDelay( 1 ), so the task just yields for one tick.

lib/utils/platform/iot_threads_afr.c
cobusve: The second one was test code I intended to remove and missed! Do you think we should keep it?

lib/utils/platform/iot_threads_afr.c
cobusve: Yes, otherwise you will think you took the semaphore and end up chasing a hard bug. I will add it.

lib/utils/platform/iot_threads_afr.c
cobusve: Agreed on the concept yes, but this function is just really implementing function overloading, the real implementation does log when a timeout happened, and if there is no timeout it always returns false, so above does not work here.

lib/utils/platform/iot_threads_afr.c
cobusve: OK, will do

lib/utils/platform/iot_threads_afr.c
cobusve: Overflow if we change the time to be 32 bit, since it was 64-bit ... 

What do we want to do here? If we divide first it will round, tick rates are typically things like 100Hz which is 10ms, or possibly 1000 which is 1ms.

lib/utils/platform/iot_clock_afr.c
cobusve: I should remove them. They were a holdover from time.h which this replaced. The last 2 are being used.

lib/utils/platform/iot_threads_afr.c
cobusve: Yes!

lib/utils/platform/iot_threads_afr.c
cobusve: Agree

lib/utils/platform/iot_threads_afr.c
cobusve: Yes, it was a debugging holdover, will remove the extra message.

lib/utils/platform/iot_threads_afr.c
cobusve: Agree

lib/utils/platform/iot_threads_afr.c
cobusve: Yes!

lib/utils/platform/iot_clock_afr.c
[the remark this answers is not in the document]
lt72: Here is the exemplification of the two cases:
files i.h is the platform independent file (e.g. the one that define IotMutex_t as ```typedef _IotSystemMutex_t       IotMutex_t;```)
file d.h is the one that defines the implementation, e.g. the one that typedefs 
```typedef pthread_mutex_t   _IotSystemMutex_t;```)
then there is a config.h.

with the current implementation, i.h needs to include config.h, which includes d.h.
pros: i.h and d.h an be documented independently. 
cons: one ore intermediate type.

with the alternate proposal, config.h would need to typedef those types directly
pro: one less type
cons: cannot do (or more difficult) doxygen docs directly in code

PR #408, opened 2019-03-22
Feature/ble beta afrabstractions

cobusve:
Adding implementation of clocks and threads to use FR directly

Description
-----------
Adds iot_threads_afr.c and iot_clock_afr.c to provide an implmentation which calls to FreeRTOS directly instead of going thought the POSIX layer.

Checklist:
----------
- [ ] My code is Linted.


demos/common/include/iot_config_common.h
```diff
@@ -155,6 +156,16 @@
     #define IOT_NETWORK_RECEIVE_TASK_STACK_SIZE    ( 5 * configMINIMAL_STACK_SIZE )
 #endif
 
+/* Default iot thread stack_size. */
+#ifndef IOT_THREAD_DEFAULT_STACK_SIZE
+    #define IOT_THREAD_DEFAULT_STACK_SIZE    configMINIMAL_STACK_SIZE
```
gordonwang0: We should be careful about using the minimal stack size as the default (although I don't have a good default value).

demos/common/include/iot_config_common.h
```diff
@@ -155,6 +156,16 @@
...
+#endif
+
+/* Default iot thread priority. */
+#ifndef IOT_THREAD_DEFAULT_PRIORITY
+    #define IOT_THREAD_DEFAULT_PRIORITY      5
```
gordonwang0: Why 5?  Also it's not guaranteed that some boards have 5 discrete priorities.

We generally default to idle priority.

demos/common/include/iot_config_common.h
cobusve: I only matched the previous behavior

demos/common/include/iot_config_common.h
cobusve: Again, I defaulted this to what the POSIX code was doing. This is a config param so you can make it what you need to.

demos/common/include/iot_config_common.h
cobusve: For now I am leaving this as minimal, it is the best on resources and if any platform needs more stack space it should be overridden for that platform.  We can always increase this configuration parameter later.

demos/common/include/iot_config_common.h
cobusve: I updated this one to tskIDLE_PRIORITY as per your request, when a platform needs this to be higher it can be overriden easy enough.

PR #414, opened 2019-03-25
POSIX compliance fix.

yuhui-zheng:
POSIX compliance fix.

Description
-----------
This PR comes from staging repo feature/ble-beta branch commit SHA 4b4d61db72079236e307e9fcf48873f581575118.
This PR passes winsim test project, POSIX test suite. 

FreeRTOS_POSIX_clock.c -- clock_getcpuclockid(), Open Group "Upon successful completion, clock_getcpuclockid() shall return zero; otherwise, an error number shall be returned to indicate the error."

FreeRTOS_POSIX_clock.c -- clock_gettime(), Open Group does not require this branch, and returning -1 without errno set is confusing. Thus taking out tp null check.

FreeRTOS_POSIX_mqueue.c -- mq_open(), Open Group "[EINVAL]
O_CREAT was specified in oflag, the value of attr is not NULL, and either mq_maxmsg or mq_msgsize was less than or equal to zero." Additional args -- mode and attr are only required when O_CREAT.

FreeRTOS_POSIX_mqueue.c -- mq_unlink(). To be consistent in our own implementation, if prvValidateQueueName() fails, lets always return EINVAL. Refer to mq_open().

FreeRTOS_POSIX_pthread.c -- pthread_attr_destroy(), Open Group does not require null check, and also to be consistent in our own implementation, remove null check.

FreeRTOS_POSIX_pthread.c -- pthread_attr_init(), Open Group does not require null check.

FreeRTOS_POSIX_pthread.c -- pthread_equal(), if one of the thread was null, this function would tell you threads are equal, which is a false statement.

FreeRTOS_POSIX_pthread_mutex.c -- pthread_mutexattr_init(), Open Group does not requrie EINVAL, and also to be consistent within our own implementation.

test update -- Expected return value update. 

API doc update -- We removed memory allocation inside a couple of APIs. Thus, error value "memory not available" will never be returned from these APIs. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
+++ b/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
@@ -63,11 +63,8 @@ int clock_getcpuclockid( pid_t pid,
     ( void ) pid;
     ( void ) clock_id;
 
-    /* This function is currently unsupported. It will always return -1 and
-     * set errno to EPERM. */
-    errno = EPERM;
-
-    return -1;
+    /* This function is currently unsupported. It will always return EPERM. */
+    return EPERM;
 }
 
 /*-----------------------------------------------------------*/
@@ -103,14 +100,6 @@ int clock_gettime( clockid_t clock_id,
 
     /* Silence warnings about unused parameters. */
     ( void ) clock_id;
-
-    /* Check tp. */
-    if( tp == NULL )
-    {
-        /* POSIX does not specify this function setting errno for invalid
-         * parameters, so just set the return value. */
-        iStatus = -1;
-    }
 
     if( iStatus == 0 )
     {
--- a/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
+++ b/lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
@@ -91,7 +91,6 @@ int clock_gettime( clockid_t clock_id,
                    struct timespec * tp )
 {
     TimeOut_t xCurrentTime = { 0 };
-    int iStatus = 0;
 
     /* Intermediate variable used to convert TimeOut_t to struct timespec.
      * Also used to detect overflow issues. It must be unsigned because the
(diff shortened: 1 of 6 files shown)


lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
```diff
@@ -104,14 +101,6 @@ int clock_gettime( clockid_t clock_id,
     /* Silence warnings about unused parameters. */
     ( void ) clock_id;
 
-    /* Check tp. */
-    if( tp == NULL )
```
abhidixi11: If this check is removed, iStatus will never change,  so

>  if(iStatus ==0)

 may not be needed.

lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
```diff
@@ -63,11 +63,8 @@ int clock_getcpuclockid( pid_t pid,
     ( void ) pid;
     ( void ) clock_id;
 
-    /* This function is currently unsupported. It will always return -1 and
-     * set errno to EPERM. */
-    errno = EPERM;
-
-    return -1;
+    /* This function is currently unsupported. It will always return EPERM. */
+    return EPERM;
```
abhidixi11: I didn't know clock_getcpuclockid actually returns errno, but should we still set it for perror to work ?

lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
yuhui-zheng: updated in the second commit.

lib/FreeRTOS-Plus-POSIX/source/FreeRTOS_POSIX_clock.c
yuhui-zheng: Discussed offline. Leaving it as-is for now, as we are following POSIX original definition of [clock_getcpuclockid()](http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getcpuclockid.html). Error number is returned instead of passed via errno variable.  

This brings up a good question of what our POSIX API tests actually check. A good API test case should cover reasonable input combinations, output patterns, and side effects. Clearly we don't do all above all the time...

PR #430, opened 2019-03-27
fix metrics library and defender demo

qiutongs:
Description
-----------

- fix metrics library
- fix defender demo

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/include/platform/iot_metrics.h
+++ b/lib/include/platform/iot_metrics.h
@@ -97,9 +97,9 @@
 
 #endif /* if IOT_STATIC_MEMORY_ONLY */
 
-/* This is supposed to be same as the data type of socket. */
+/* This data type varies on different platform. */
 #ifndef IotMetricsConnectionId_t
-    #define IotMetricsConnectionId_t    void *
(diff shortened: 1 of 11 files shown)


lib/include/platform/iot_metrics.h
```diff
@@ -87,10 +87,19 @@
         #define IotMetrics_FreeTcpConnection    free
     #endif
 
+    #ifndef IotMetrics_MallocIpAddress
+        #define IotMetrics_MallocIpAddress    malloc
```
lt72: all malloc/free functions should be defined to pvPortMalloc and pvPortFree, the FreeRTOS wrappers for malloc and free
whatever else is wrong, even in the libraries...
this is a difference between C-SDK code and a:FR code...

lt72: For example, check this file
c:\src2\amazon-freertos\tests\common\include\iot_config_common.h
for the test config... we should have the same setup as this.

qiutongs: Thanks for catching this. I missed it. Let me update iot_config_common.h

PR #434, opened 2019-03-28
Atomic & POSIX with atomic.

yuhui-zheng:
Atomic & POSIX with atomic.

Description
-----------
- ```atomic.h``` header. 
   post previous review:
(1) added macro ```ATOMIC_COMPARE_AND_SWAP_SUCCESS```/```ATOMIC_COMPARE_AND_SWAP_FAILURE```.
(2) CAS API returns type uint32_t, which indicates swapped or not, instead of previous value. 
(3) if ```portmacro.h``` does not provide compiler specific force inline, don't inline. 
- POSIX barrier, pthread mutex condition, and semaphore are updated. (```pthread_cond_broadcast()``` and ```pthread_cond_signal()``` would be fun to review, CAS is involved. To see which mutexes are removed, ```FreeRTOS_POSIX_internal.h```)
- POSIX semaphore maximum allowed value got updated, since POSIX requires sem value to be int, and atomic is using uint32_t.
- Task pool maximum allowed semaphore value also got updated due to above.
- windows port test project and project filter.

### !!!Note for reviewers!!!
- Before rebasing with most recent feature/ble-beta branch, this commit was built on WinSim port and passed POSIX test suite and memory leak test. Though our tests don't do much.
- <del>WinSim port test project currently doesn't built after rebase. I will create another PR to fix.</del> 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-POSIX/include/FreeRTOS_POSIX_internal.h
+++ b/lib/FreeRTOS-Plus-POSIX/include/FreeRTOS_POSIX_internal.h
@@ -45,9 +45,10 @@
 #endif
 
 #if posixconfigENABLE_PTHREAD_MUTEX_T == 1
-    /**
-     * @brief Mutex.
-     */
+
+/**
+ * @brief Mutex.
+ */
     typedef struct pthread_mutex_internal
     {
         BaseType_t xIsInitialized;          /**< Set to pdTRUE if this mutex is initialized, pdFALSE otherwise. */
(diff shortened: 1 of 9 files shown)


yuhui-zheng:
WinSim port aws_tests and aws_demo build fine. aws_tests project fails at this test. 
https://github.com/aws/amazon-freertos/blob/master/tests/common/posix/aws_test_posix_clock.c#L331

Given team is fixing projects on all ports, merging this PR.

PR #438, opened 2019-03-28
Check for DNS lookup failure in platform network

gordonwang0:
Description
-----------
Check for errors from gethostbyname() in platform network abstraction.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/utils/platform/iot_network_afr.c
+++ b/lib/utils/platform/iot_network_afr.c
@@ -316,6 +316,13 @@ IotNetworkError_t IotNetworkAfr_Create( void * pConnectionInfo,
     serverAddress.usPort = SOCKETS_htons( pServerInfo->port );
     serverAddress.ulAddress = SOCKETS_GetHostByName( pServerInfo->pHostName );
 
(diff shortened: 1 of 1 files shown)


htibosch:
Looking at your changes, I have 2 remarks:
After you have logged that `ulAddress` equals zero ( url not found ), would it still make sense to call `SOCKETS_Connect()`?
Secondly, when you set `SOCKETS_SO_RCVTIMEO` to zero, I think that means (non-blocking) polling mode. So both the old and the proposed new comment seem wrong.

qiutongs:
> Looking at your changes, I have 2 remarks:
> After you have logged that `ulAddress` equals zero ( url not found ), would it still make sense to call `SOCKETS_Connect()`?
> Secondly, when you set `SOCKETS_SO_RCVTIMEO` to zero, I think that means (non-blocking) polling mode. So both the old and the proposed new comment seem wrong.

About first remark:
"_IOT_SET_AND_GOTO_CLEANUP" macro will internally jump to another section of code, thus skipping connecting

PR #450, opened 2019-03-30
Enable CMake for ESP

gordonwang0:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


lib/serializer/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+afr_module(serializer)
```
tgsong: Is this module supposed to be used by end user?

gordonwang0: Yes, it may be.

lib/mqtt/CMakeLists.txt
```diff
@@ -1,28 +1,40 @@
 afr_module(mqtt)
 
+# Compile BLE MQTT serializers on supported platforms.
+if(${BLE_SUPPORTED})
+    set(extra_mqtt_sources "${AFR_MODULES_DIR}/mqtt/iot_ble_mqtt_serialize.c")
+    set(extra_mqtt_private_includes ${BLE_LIBRARY_INCLUDE_DIR})
+    set(extra_mqtt_dependencies AFR::serializer)
```
tgsong: Should we add `AFR:ble` to the dependencies list here? Instead of adding `BLE_LIBRARY_INCLUDE_DIR` to `extra_mqtt_private_includes`

cmake/afr.cmake
```diff
@@ -15,6 +15,19 @@ set(AFR_DEMOS_DIR "${AFR_ROOT_DIR}/demos/common" CACHE INTERNAL "Amazon FreeRTOS
 set(AFR_TESTS_DIR "${AFR_ROOT_DIR}/tests/common" CACHE INTERNAL "Amazon FreeRTOS tests root.")
 set(AFR_3RDPARTY_DIR "${AFR_MODULES_DIR}/third_party" CACHE INTERNAL "3rdparty libraries root.")
 
+# These paths will be used for boards that support BLE.
+set(BLE_LIBRARY_INCLUDE_DIR "${AFR_MODULES_DIR}/include/bluetooth_low_energy" CACHE INTERNAL "BLE include directory.")
+set(BLE_DEMO_COMMON_SOURCES "${AFR_DEMOS_DIR}/ble/iot_ble_numericComparison.c" CACHE INTERNAL "BLE demo common source files.")
+
+# Network manager will temporarily be built into the demo executables.
+set(NETWORK_MANAGER_SOURCES
```
tgsong: So this will be refactored to a demo in the future?

gordonwang0: It will be a separate library in the future.

PR #477, opened 2019-04-04
Check for EWOULDBLOCK on timeout

gordonwang0:
Description
-----------
Secure sockets is inconsistent and returns either 0 or `EWOULDBLOCK` on timeout, depending on the port.  Change the network layer to check for both values.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/common/include/iot_config_common.h
+++ b/demos/common/include/iot_config_common.h
@@ -34,7 +34,7 @@
 #define IOT_SDK_VERSION    "4.0.0"
 
 /* Standard library function overrides. */
-#define IotLogging_Puts( str )                 configPRINTF( ( "%s\n", str ) )
+#define IotLogging_Puts( str )                 configPRINTF( ( "%s\r\n", str ) )
 #define IotContainers_Assert( expression )     configASSERT( expression )
 #define IotMqtt_Assert( expression )           configASSERT( expression )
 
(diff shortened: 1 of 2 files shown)


htibosch:
Hi, as you probably know, the return values from `recvfrom()` and `recv()` in FreeRTOS+TCP are not entirely following the BSD rules.

In case there is no data, `recvfrom()` will always return `-EWOULDBLOCK`, whereas `recv()` returns zero.

I think that now it is too late to change this behaviour: users would complain. But yes, it is good to always check for **both** zero and `-EWOULDBLOCK`, and treat them the same.

All negative return codes except `-EWOULDBLOCK` and `-EINTR` are fatal.

To summarise:

`FreeRTOS_recvfrom()` may return:

`    <positive number>          // Data was received`
`    -pdFREERTOS_ERRNO_EINTR    // Operation was interrupted`
`    -pdFREERTOS_ERRNO_EWOULDBLOCK // No data awas vailable`

`FreeRTOS_recv()` may return:

`    <positive number>          // Data was received`
`    -pdFREERTOS_ERRNO_EINVAL   // This is not a valid TCP socket, or socket not bound`
`    -pdFREERTOS_ERRNO_ENOMEM   // RX-stream buffer could not be allocated`
`    -pdFREERTOS_ERRNO_ENOTCONN // socket is not yet connected`
`    -pdFREERTOS_ERRNO_EINTR    // Operation was interrupted`
`    0                          // No data was available`

Issue #452, opened 2019-03-31
esp32 WIFI_GetMAC returns wrong mac address


chegewara:
esp32 is having 4 mac addresses, from which 2 are wifi related. one mac is for AP and one mac is for STA. In STA mode (connected to router AP) `FreeRTOS_GetAddressConfiguration` is returning AP mac address.

abhidixi11:
Hello @chegewara,
Can you please  explain what do you mean by "FreeRTOS_GetAddressConfiguration() is returning AP mac address" ?  The function returns IP address/ Gateway address obtained from DHCP server.  Where are you seeing mac address. ? Can you please  post some logs? Thanks.

chegewara:
Sorry, my bad, of course i meant `WIFI_GetMAC`.

abhidixi11:
Thanks @chegewara ,
As you have pointed out, in aws_wifi.c  WIFI_GetMAC() calls esp_wifi_get_mac with WIFI_MODE_STA (hardcoded here) ,which is equivalent of ESP_IF_WIFI_AP, so AP address is returned instead of STA. 

I'll make sure change is made soon, it should be interface and not mode.  It should be either ESP_IF_WIFI_STA or WIFI_IF_STA as defined in esp_wif_types.h. 
Thanks.

abhidixi11:
I have created PR. It will be merged soon. Closing the issue. https://github.com/aws/amazon-freertos/pull/481

chegewara:
Thanks.

PR #481, opened 2019-04-04
fix: ESP WIFI_GetMac returning wrong mac.

abhidixi11:
ESP WIFI_GetMac was calling esp_wifi_get_mac with wrong argument.

# Related Issues:
https://github.com/aws/amazon-freertos/issues/452

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/wifi/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_wifi.c
+++ b/lib/wifi/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_wifi.c
@@ -591,7 +591,23 @@ WIFIReturnCode_t WIFI_GetMAC( uint8_t * pucMac )
     /* Try to acquire the semaphore. */
     if( xSemaphoreTake( xWiFiSem, xSemaphoreWaitTicks ) == pdTRUE )
     {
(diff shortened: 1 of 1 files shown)


lib/wifi/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_wifi.c
```diff
@@ -591,7 +591,23 @@ WIFIReturnCode_t WIFI_GetMAC( uint8_t * pucMac )
     /* Try to acquire the semaphore. */
     if( xSemaphoreTake( xWiFiSem, xSemaphoreWaitTicks ) == pdTRUE )
     {
-        esp_err_t ret = esp_wifi_get_mac(WIFI_MODE_STA, pucMac);
+        wifi_mode_t mode;
+        esp_err_t ret = esp_wifi_get_mode(&mode);
+        if( ret == ESP_OK )
+        {
+            if( mode == WIFI_MODE_STA )
```
mradulan: will a switch case be better in this?

abhidixi11: For many conditions, switch is better, but for just three condition, it won't make a difference.

PR #502, opened 2019-04-09
Restore CMake builds on TI, ST and ESP

gordonwang0:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


tests/CMakeLists.txt
```diff
@@ -48,10 +62,27 @@ afr_module_dependencies(
 
 # MQTT test
 afr_test_module(test_mqtt)
+
+# Enable tests for BLE-over-MQTT serializers on supported boards.
+if(${BLE_SUPPORTED})
```
tgsong: you can just write `if(BLE_SUPPORTED)`

tests/CMakeLists.txt
```diff
@@ -141,6 +187,18 @@ afr_module_dependencies(
         AFR::wifi
 )
 
+# BLE test
+# Only build on boards that support BLE.
+if(${BLE_SUPPORTED})
```
tgsong: You don't need to guard this, just make it depends on `AFR::ble` and it will be automatically disabled when `AFR::ble` is not available.

demos/common/mqtt/CMakeLists.txt
```diff
@@ -53,3 +53,14 @@ afr_module_dependencies(
     INTERFACE
         AFR::mqtt
 )
+
+# MQTT-over-BLE demo.
+# Only build on boards that support BLE.
+if(${BLE_SUPPORTED})
```
tgsong: You don't need to guard this, just make it depends on `AFR::ble` and it will be automatically disabled when `AFR::ble` is not available.

cmake/vendors/espressif/esp32.cmake
```diff
@@ -116,7 +139,13 @@ target_sources(
 afr_mcu_port(ota)
 target_sources(
     AFR::ota::mcu_port
-    INTERFACE "${AFR_MODULES_DIR}/ota/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_ota_pal.c"
+    INTERFACE
+        "${AFR_MODULES_DIR}/ota/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_ota_pal.c"
+        "${AFR_MODULES_DIR}/ota/portable/espressif/esp32_devkitc_esp_wrover_kit/aws_esp_ota_ops.c"
+)
+target_link_libraries(
```
tgsong: Do we need this dependency information here? OTA already depends on MQTT, which should chain to pkcs11

cmake/vendors/espressif/esp32.cmake
```diff
@@ -77,12 +77,35 @@ target_sources(
...
+        "${AFR_MODULES_DIR}/bluetooth_low_energy/portable/espressif/"
+)
+
 # PKCS11
 afr_mcu_port(pkcs11)
 target_link_libraries(
     AFR::pkcs11::mcu_port
     INTERFACE AFR::pkcs11_mbedtls
 )
+target_include_directories(
```
tgsong: Need to verify why this is needed.

PR #503, opened 2019-04-09
Updated Nordic Tests

cobusve:
Updated the Nordic Test project to match demo changes

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/common/test_runner/iot_test_afr.c
+++ b/tests/common/test_runner/iot_test_afr.c
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2018 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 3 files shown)


tests/common/test_runner/iot_test_afr.c
```diff
@@ -0,0 +1,86 @@
...
+
+
+#include "aws_clientcredential.h"
+#include "platform/iot_network_afr.h"
+#include "iot_network_manager_private.h"
+#include "platform/iot_threads.h"
+#include "iot_common.h"
+#include "iot_mqtt.h"
+
+static IotNetworkManagerSubscription_t subscription = IOT_NETWORK_MANAGER_SUBSCRIPTION_INITIALIZER;
```
ravibhagavandas: We can remove this and the header file as network manager is not used in the tests.

cobusve: Ok I removed it from the tests, will check in with the PR for the next vendor

PR #516, opened 2019-04-11
fix: Cypress build error on IDE

mradulan:
The file hw_poll.c complains of unknown type for uint32_t.
Included FreeRTOS.h to fix the error.


lib/pkcs11/portable/cypress/CYW943907AEVAL1F/hw_poll.c
```diff
@@ -42,6 +42,9 @@
  * truly random in nature.
  */
 
+/* Amazon FreeRTOS Includes. */
+#include "FreeRTOS.h"
```
sarenameas: This is better than stdint.h right?

Issue #471, opened 2019-04-03
FreeRTOS DNS Behavior


sbertrand-witekio:
How does the DNS cache uses entries passed aged out records ?

Ageing out of record is done by setting the [DNS name to a NULL string](https://github.com/aws/amazon-freertos/blob/master/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c#L1367). 

However looking for each entry in the DNS table stop at the [first NULL string DNS Name](https://github.com/aws/amazon-freertos/blob/master/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c#L1349).

Therefore when a record is aged out, all record of higher index in the table will not be checked.

abhidixi11:
Hello @sbertrand-witekio,

Thanks for pointing it out.  I checked the code looks like a bug.  Did you actually see  DNS cache lookup failing or did you just noticed it in the code? 
I'll verify it again and post an update.
Thanks.

sbertrand-witekio:
I just looked that the code, my DNS cache table is only of 1 so that configuration does not have the issue.
I can do a few tests with a bigger cache table.

Another issue I found was the [IP address reported](https://github.com/aws/amazon-freertos/blob/master/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c#L436) when a record is found, is not formatted correctly, I added a FreeRTOS_ntohl to have the correct byte order.

abhidixi11:
Hello @sbertrand-witekio ,
We will work on fixing the issue. 
Thanks

abhidixi11:
I have committed the changes and closing the issue.  Thanks.

PR #525, opened 2019-04-15
fix: DNS cache processing

abhidixi11:
prvProcessDNSCache() stops the search when pcName[0] is zero
pcName[0] is set to zero when cache entry is invalidated.

https://github.com/aws/amazon-freertos/issues/471

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
+++ b/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
@@ -1348,7 +1348,7 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 		{
 			if( xDNSCache[ x ].pcName[ 0 ] == 0 )
 			{
-				break;
+				continue;
 			}
 
 			if( 0 == strcmp( xDNSCache[ x ].pcName, pcName ) )


lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -1346,11 +1346,6 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 		/* For each entry in the DNS cache table. */
 		for( x = 0; x < ipconfigDNS_CACHE_ENTRIES; x++ )
 		{
-			if( xDNSCache[ x ].pcName[ 0 ] == 0 )
-			{
-				break;
```
dcgaws: Let's not proceed into the loop if it's an invalid entry, though. Could we replace the break with a continue?

abhidixi11: Done. The only if statement in the loop would have achieved the same effect, but continue is cleaner.

alexa-noxon:
The re-review request was an accident. Please disregard.

PR #529, opened 2019-04-16
Make CBMC proofs run on Linux

markrtuttle:
Make CBMC proofs run on Linux

Description
-----------
The last PR to the CBMC proofs included scripts to generate makefiles on Windows, MacOS, and Linux to run the CBMC proofs.  This PR corrects some pathnames for Linux and ensures that all scripts run under both python2 and python3.

Checklist:
----------
The proofs have been checked on MacOS and Linux.
The proofs have been shown to run under continuous integration.


tools/cbmc/proofs/make-proof-makefiles.py
```diff
@@ -153,13 +153,13 @@ def dump_makefile(dyr, system):
...
 
 def compute(value, so_far, system, key, harness, appending=False):
-    if not isinstance(value, (str, float, int)):
-        logging.error(wrap("""\
-                        in file %s, the key '%s' has value '%s',
-                        which is of the wrong type (%s)"""),
-                      os.path.join(harness, "Makefile.json"), key,
-                      str(value), str(type(value)))
-        exit(1)
+#    if not isinstance(value, (str, float, int)):
```
karkhaz: Please could you:
1. Change this line to `if not isinstance(value, (basestring, float, int)):`
2. Uncomment the block
3. Near the top of the file, add the following snippet:
```python
try:
  basestring
except NameError:
  basestring = str
```
This works because `basestring` is the supertype of `str` and `unicode` on python2, so the try block succeeds. It does not exist on python3, so the `except` block makes it an alias for `str` (which is also the same as `unicode` on py3), so it can then be used in the `instanceof` call just like on py2.

PR #555, opened 2019-04-24
combine mcuxpresso and iar in cmake file

qiutongs:
Description
-----------

Both IAR and GCC compile successfully.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/nxp/boards/lpc54018iotmodule/CMakeLists.txt
+++ b/vendors/nxp/boards/lpc54018iotmodule/CMakeLists.txt
@@ -1,14 +1,22 @@
+# Set variables.
 set(lpc54018_dir "${AFR_VENDORS_DIR}/nxp/LPC54018")
 set(lpc54018iotmodule_dir "${AFR_VENDORS_DIR}/nxp/boards/lpc54018iotmodule")
 
 set(lpc54018_ports_dir "${lpc54018iotmodule_dir}/ports")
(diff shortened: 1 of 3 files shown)


vendors/nxp/boards/lpc54018iotmodule/CMakeLists.txt
```diff
@@ -1,14 +1,22 @@
...
+    set(lpc54018_aws_dir "${lpc54018iotmodule_dir}/aws_demos")
+endif()
+
+# Include IDE specific cmake file.
+if(${AFR_TOOLCHAIN} STREQUAL "arm-gcc")
+    include("mcuxpresso.cmake")
+elseif(${AFR_TOOLCHAIN} STREQUAL "arm-iar") 
+    include("iar.cmake")
+else()
+    message( FATAL_ERROR "The toolchain is not supported." )
```
tgsong: Throwing a fatal error here would be an issue for OCW. They want to run it in configuration mode only which use the default host compiler (gcc for x86 on linux). I have a variable for OCW `AFR_NON_BUILD_MODE`, you can instead check,
```cmake
elseif(NOT AFR_NON_BUILD_MODE)
    message( FATAL_ERROR "The toolchain is not supported." )
endif()
```

qiutongs: Got it. Let me update.

PR #558, opened 2019-04-25
Changed some rules for uncrustify

htibosch:
Changed some rules for uncrustify

Description
-----------

About .uncrustify.cfg:

I tested the existing .uncrustify.cfg in combination with "tasks.c" to see what the outcome would look like: not so good.

Here below a few recommendations:

The FreeRTOS coding style never used a repeated star in multi-line comments like this one:

    /* This comment uses a
     * repeated star.
     */

There are 2 settings that allow for multi-line comments of this type:

    /* If configUSE_PORT_OPTIMISED_TASK_SELECTION is 1 then task selection is
    performed in a way that is tailored to the particular microcontroller
    architecture being used. */

I suggest to use:

    cmt_star_cont                   = false
    cmt_indent_multi                = false

No space between a pointer star and an identifier, like here:

    TaskHandle_t xTaskGetHandle( const char *pcNameToQuery )

In the kernel there was never a space between the '*' and the identifier.

    sp_after_ptr_star               = ignore # ignore/add/remove/force

Do not use "remove" because there is an exception when const is used:

    static void prvInitialiseNewTask(( const char * const pcName

Very important: **indent with tabs, not spaces** :

When using the current .uncrustify.cfg, all tab characters are replaced by spaces!
That is very disturbing when committing or comparing old and new source code.

I don't think that was meant to happen.

// How to use tabs when indenting code
// 0=spaces only
// 1=indent with tabs to brace level, align with spaces (default)
// 2=indent and align with tabs, using spaces when not on a tabstop

I would suggest to use 2:

    indent_with_tabs                = 2        # unsigned number

There are sources on github/aws that miss all leading tab characters and have spaces in stead ( for instance aws_test_freertos_tcp.c ).

Also, in the original sources, alignment after defines was done with the help of TAB's and not spaces:

`#define tskRUNNING_CHAR<tab>( 'X' )`
`#define tskBLOCKED_CHAR<tab>( 'B' )`
`#define tskREADY_CHAR<tab><tab>( 'R' )`
`#define tskDELETED_CHAR<tab>( 'D' )`
`#define tskSUSPENDED_CHAR<tab>( 'S' )`

which needs:

    align_with_tabs                 = true     # false/true

You can play with the minimum gap width:

    align_pp_define_gap             = 4        # unsigned number

Local variables in a function are placed at brace level, **without** indentation:

    BaseType_t xTaskCreateRestrictedStatic( const TaskParameters_t... )
    {
    TCB_t *pxNewTCB;
    BaseType_t xReturn = errCOULD_NOT_ALLOCATE_REQUIRED_MEMORY;
            <empty line>
            configASSERT( pxTaskDefinition->puxStackBuffer != NULL );
            configASSERT( pxTaskDefinition->pxTaskBuffer != NULL );

This rule will fix it:

    indent_var_def_blk              = -4       # number

The #if blocks with code used to be declared like this

            #if( tskSTATIC_AND_DYNAMIC_ALLOCATION_POSSIBLE != 0 )
            {
                /* Tasks can be created statically or dynamically, so note this
                task was created statically in case the task is later deleted. */
                pxNewTCB->ucStaticallyAllocated = tskSTATICALLY_ALLOCATED_STACK_AND_TCB;
            }
            #endif /* configSUPPORT_DYNAMIC_ALLOCATION */

The following define is needed:

    pp_indent_brace                 = false     # false/true

    #if( INCLUDE_vTaskDelete == 1 )

Uncrustify wants a space between #if and the bracket. I don't know how to avoid that behaviour.

Please play with the adapted `.uncrustify.cfg` and for instance `tasks.c`. You will see that the FreeRTOS kernel style will be better preserved.

Checklist:
----------

- [x] My code is Linted.


dcgaws:
Hein, hello, we deviated from the FreeRTOS kernel coding style for our libraries and demos (i.e. essentially everything that's not under lib\FreeRTOS) when we started the project (i.e. post acquisition). That's one of the reasons that the kernel files have never been run through .uncrustify.cfg. 

We do our best to be consistent within that context: preserve the kernel coding style for changes we make to the core kernel files, and stick with the libraries coding style for everything else. 

Having said that, to your point, it's true that we don't have an Uncrustify configuration in this repo for the kernel coding style. I have no objection to there being one, but I suppose it would best be placed under lib\FreeRTOS.

sarenameas:
Once again thanks for this uncrustify config update for the FreeRTOS kernel, but updating it to match the kernel code styling completely/perfectly is going to take some time. I have create an issue internally, but for now II don't believe we can't really accept these changes. Given that you are using it to update the FreeRTOS+TCP code and it looks fine, I would suggest placing it under folder of the library at **amazon-freertos\libraries\freertos_plus\standard\freertos_plus_tcp**

htibosch:
Sarena, your proposal to put a new uncrustify.cfg in the +TCP directory sounds good to me.
I don't think it needs an extra /tools directory?
I will close this PR now, because I wrote a new #913 which replaces this one.

Issue #367, opened 2019-03-06
How to create a new ESP32 project with my own app_main?


jrr:
**Describe the bug**

The [getting started guide](https://docs.aws.amazon.com/freertos/latest/userguide/getting_started_espressif.html) covers building and running the included `esp32_devkitc_esp_wrover_kit` ("aws_demos") demo. This works great!

For a standard (non-AFR) ESP-IDF project, you can build it anywhere so long as the `$IDF_PATH` environment variable is configured. In the [documentation](https://docs.espressif.com/projects/esp-idf/en/stable/get-started/index.html#start-a-project), an example is copied out of IDF and built in place. This works:

```
$ export IDF_PATH=~/esp/esp-idf
$ cp -R $IDF_PATH/examples/get-started/hello_world .
$ cd hello_world/
$ make all -j
```

I can do something similar with the aws_demos demo:

```
$ export IDF_PATH=~/repos/amazon-freertos/lib/third_party/mcu_vendor/espressif/esp-idf
$ export AMAZON_FREERTOS_PATH=~/repos/amazon-freertos
$ cp -R ~/repos/amazon-freertos/demos/espressif/esp32_devkitc_esp_wrover_kit/make new-project
$ cd new-project
$ make all -j
```

The problem arises when I try to add a component with my own `app_main()`. Here I copy it from the original esp-idf example:

```
$ cp -R ~/esp/esp-idf/examples/get-started/hello_world/main .
$ make all
CC build/main/hello_world_main.o
AR build/main/libmain.a
LD build/aws_demos.elf
~/new-project/build/amazon-freertos-common/libamazon-freertos-common.a(main.o): In function `app_main':
~/repos/amazon-freertos/demos/espressif/esp32_devkitc_esp_wrover_kit/common/application_code/main.c:136: multiple definition of `app_main'
~/new-project/build/main/libmain.a(hello_world_main.o):/Users/jrr/new-project/main/hello_world_main.c:17: first defined here
collect2: error: ld returned 1 exit status
make: *** [~/new-project/build/aws_demos.elf] Error 1
```

This is because the `amazon-freertos-common` component includes its own main.c:
https://github.com/aws/amazon-freertos/blob/ab5e7d5d2c20b7dd2103dc53b4b1f398ac853132/demos/espressif/esp32_devkitc_esp_wrover_kit/common/application_code/espressif_code/amazon-freertos-common/component.mk#L40

Are we meant to use this component? How can I exclude its main.c?

[...]

Issue #497, opened 2019-04-06
make flash fails on OSX due to hwcrypto/aes.h: No such file or directory


mluis:
**Describe the bug**
While running `make flash monitor` from the tutorial the procedure fails.

**System information**
- Which hardware board or part numbers? 
ESP32 WROOM-32
- Operating System
OSX
- Version of Amazon FreeRTOS 
v1.4.7-39-ge1c0f7f7
- Project 
MQTT

**Expected behavior**
Procedure of `make flash monitor` show proceed.

**Additional context**
```
amazon-freertos/demos/espressif/esp32_devkitc_esp_wrover_kit/common/application_code/espressif_code/mbedtls/port/include/aes_alt.h:31:26: fatal error: hwcrypto/aes.h: No such file or directory
compilation terminated.
make[1]: *** [lib/tls/aws_tls.o] Error 1
make: *** [component-amazon-freertos-common-build] Error 2
```

Thank you!

abhidixi11:
Hello @mluis ,

The file should be part of your repo. Can you please verify that you have aes.h your repository here  :
lib/third_party/mcu_vendor/espressif/esp-idf/components/esp32/include/hwcrypto/aes.h ?
Thanks!

mluis:
Hello @abhidixi11 ,

The file is part of the repo.
```
$ file lib/third_party/mcu_vendor/espressif/esp-idf/components/esp32/include/hwcrypto/aes.h
lib/third_party/mcu_vendor/espressif/esp-idf/components/esp32/include/hwcrypto/aes.h: c program text, ASCII text
```

abhidixi11:
Hello @mluis,
Have you tried building with the latest master branch ?
Also can you please try 'make all' and see if the build goes through.
Thanks!

mluis:
Hi @abhidixi11 ,

Yes, just tried it.

```
[mluis@macb0x:~/research/aws/amazon-freertos/demos/espressif/esp32_devkitc_esp_wrover_kit/make]$ git pull
Already up to date.
[mluis@macb0x:~/research/aws/amazon-freertos/demos/espressif/esp32_devkitc_esp_wrover_kit/make]$ make all
Toolchain path: /Users/miguelluis/research/esp/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc
Toolchain version: crosstool-ng-1.22.0-80-g6c4433a
Compiler version: 5.2.0
[...]

Issue #375, opened 2019-03-11
Compiling demo ESP32 issues


Smeedy:
Hi there,

I'm trying to setup a Greengrass/ThingsGraph from the demo directory using some left-over ESP32 devkits. I tried to compile the demo on the IDF release/v3.3 and I ran into some compile issues.

`/opt/esp/esp-idf/components/esp_event/default_event_loop.c:67:28: error: 'ESP_TASKD_EVENT_STACK' undeclared (first use in this function)`

I can quick and dirty fix this in the IDF, but than I ran into some other lwip conflicting types and so on. I must be doing something wrong. Could someone share some pointers?

platform: OSX
python: 2.7.14 virtualenv
xtensa: crosstool-ng-1.22.0-80-g6c4433a
ESP-IDF: release/v3.3

Thx,
Martijn

gordonwang0:
Hi @Smeedy,

Are you using a separate install of ESP IDF?  Currently, we only support building with the ESP-IDF hosted on this repository.  Please try clearing the IDF_PATH environment variable and then rebuilding.

jghankins:
I had the same problem, having made the same mistake.  Missed the statement in the docs about this. Easy to do.  Works like a charm after backing out that IDF_PATH, and rebuilding.

qiutongs:
Yeah, there is a note about "ESP-IDF" in ESP starting guide.
https://docs.aws.amazon.com/freertos/latest/userguide/getting_started_espressif.html

PR #561, opened 2019-04-25
Make ESP demo project copyable

danhorner:
Make ESP Demo project copyable.  #367

Description
-----------

Uncouple main.c from the demo libraries by moving it into its own
component.  To begin developing a new project
based on the demo libraries, copy the demo directory esp32_devkitc_esp_wrover_kit/make
outside of the Amazon FreeRTOS tree.

Checklist:
----------

- [N/A] I have tested my changes. No regression in existing tests.
- [ ] My code is Linted.
- Demo project compiles and runs. Test tree is unaffected.


demos/espressif/esp32_devkitc_esp_wrover_kit/make/Makefile
```diff
@@ -2,18 +2,25 @@
...
 endif
 
+# Amazon FreeRTOS relies on the bundled ESP-IDF. If you are sure you need to override the ESP-IDF,
+# also set the variable CUSTOM_IDF_PATH
 ifndef IDF_PATH
-export IDF_PATH := $(AMAZON_FREERTOS_PATH)/lib/third_party/mcu_vendor/espressif/esp-idf
+	export IDF_PATH := $(AMAZON_FREERTOS_PATH)/lib/third_party/mcu_vendor/espressif/esp-idf
+else
+	ifndef CUSTOM_IDF_PATH
+	$(error "IDF path is set. Amazon FreeRTOS expects a specific ESP-IDF. Pleae unset IDF_PATH or if you are sure, set CUSTOM_IDF_PATH and rerun.")
```
WRansohoff: Are we sure we want to allow this? It would be nice if we didn't need to use a specific version, but anecdotally I've gotten all kinds of strange build errors from accidentally using a different version of ESP-IDF.

danhorner: Right now there is no check for a user-provided IDF PATH. See related issues: #375 , #497, #338, #87. This check tightens things up a little but still provides an escape route

WRansohoff: Is anyone asking for an escape route, though? It looks like those issues were all resolved by un-setting the IDF_PATH variable. Should we go out of our way to include this behavior if it is basically just an option which lets you voluntarily break your build?

People can still modify these Makefiles if they want to tinker with the build system themselves, but I'm not sure if we should make it sound like something that we support.

alexa-noxon:
Closing as latest release made signficant changes to make system.

Issue #578, opened 2019-04-30
ulApplicationGetNextSequenceNumber shouldn't stop the scheduler during crypto calls


dcgaws:
Same rationale as #570.

PR #579, opened 2019-04-30
ulApplicationGetNextSequenceNumber must be thread safe without stopping the scheduler

dcgaws:
See #578.

### DIFF
--- a/lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
+++ b/lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
@@ -27,6 +27,7 @@
 #include "FreeRTOS.h"
 #include "FreeRTOSIPConfig.h"
 #include "list.h"
+#include "semphr.h"
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"
 #include "aws_secure_sockets.h"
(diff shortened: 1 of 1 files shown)


lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
```diff
@@ -531,52 +532,56 @@ BaseType_t SOCKETS_Init( void )
...
     {
         /* One-time initialization. */
 
         /* Ensure that the PKCS#11 module is initialized. We don't keep the
-        above lock here, since we don't want to make assumptions about hardware
+        scheduler stopped here, since we don't want to make assumptions about hardware
         requirements for accessing a crypto module. */
 
-        if( 0 == xResult )
+        if( CKR_OK == xResult )
```
alexa-noxon: NP: Technically we don't need this first check if CKR_OK == xResult, but I don't think it hurts anything.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
```diff
@@ -531,52 +532,56 @@ BaseType_t SOCKETS_Init( void )
  * and FreeRTOS+TCP threads. Assume that two or more threads may race to be the
  * first to initialize the static and handle that case accordingly.
  */
-static CK_RV prvSocketsGetCryptoSession( CK_SESSION_HANDLE * pxSession,
+static CK_RV prvSocketsGetCryptoSession( SemaphoreHandle_t *pxSessionLock,
```
alexa-noxon: NP: this file looks like it needs to be uncrustified.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
```diff
@@ -645,20 +630,24 @@ static CK_RV prvSocketsGetCryptoSession( CK_SESSION_HANDLE * pxSession,
...
-    xResult = prvSocketsGetCryptoSession( &xPkcs11Session,
+    xResult = prvSocketsGetCryptoSession( &xSessionLock,
+                                          &xPkcs11Session,
                                           &pxPkcs11FunctionList );
 
     if( 0 == xResult )
     {
         /* Request a sequence of cryptographically random byte values using
          * PKCS#11. */
+        xSemaphoreTake( xSessionLock, portMAX_DELAY );
```
alexa-noxon: These semaphore lock/give can be deleted -- locking the session for a GenerateRandom call should not be necessary, since the PKCS #11 implementation should protect calls to GenerateRandom already.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
```diff
@@ -690,22 +680,27 @@ uint32_t ulApplicationGetNextSequenceNumber( uint32_t ulSourceAddress,
...
                                           &pxPkcs11FunctionList );
 
     if( 0 == xResult )
     {
+        xSemaphoreTake( xSessionLock, portMAX_DELAY );
+
         if( 0 == ullKey )
         {
             /* One-time initialization, per boot, of the random seed. */
             xResult = pxPkcs11FunctionList->C_GenerateRandom( xPkcs11Session,
```
alexa-noxon: Do we want to continue to exclude the possibility of a "key" value being 0? 

I realize that this is a restriction of ulRand that we are stuck with atm for backward compatibility, and that the odds of this occurring are very low, but could we initialize ullKey at the same time as we initialize all the other stuff?

This would also require us to find a way to access it across the different functions which might be a pain.

dcgaws: Fixed.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
dcgaws: Fixed.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
dcgaws: Fixed.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
dcgaws: Fixed.

lib/secure_sockets/portable/freertos_plus_tcp/aws_secure_sockets.c
dcgaws: Fixed.

PR #586, opened 2019-05-02
Deactivate CBMC proofs pending source code update

karkhaz:
This commit fixes proof failures on the release-1.5 branch. The
branch does not contain any of the source code that these proofs are
verifying, so the proofs are failing.

The CBMC directory should not exist on a branch where the source code
that it verifies is not available. This commit renames the `cbmc`
directory to `cbmc-inactive` so that the CI does not attempt to run
the proofs.


karkhaz:
Closing this in favour of a longer-term solution

PR #593, opened 2019-05-06
add cmake support for renesas and update demo project

qiutongs:
Description
-----------

- add cmake file for renesas
- update demo project
- fix code conflicts and config

Checklist:
----------

- [ ] My code is Linted.


vendors/renesas/boards/rx65n-rsk/CMakeLists.txt
```diff
@@ -0,0 +1,273 @@
...
+
+# -------------------------------------------------------------------------------------------------
+# Compiler settings
+#
+# Note1: Cmake doesn't support cc-rx compiler. The settings are commented out.
+# Note2: The settings below are NOT test yet. Some flags might be wrong and some might be missing.
+# -------------------------------------------------------------------------------------------------
+afr_mcu_port(compiler)
+
+#set(compiler_flags
```
cobusve: You should remove the parts that are commented out, just like with code.

tgsong: Actually we don't need to comment those out, just leave them here, the notes above is suffice imo. Since we're not building it with CMake anyway.

PR #595, opened 2019-05-06
Add compute_path.py script to wrap used defines with ifndef macros

mmuesly:
Automated computation of header patches for CBMC proofs

Description
-----------
Checking configurations of define values requires to overwrite some of the configurations specified in the used config header files. Therefore, this script scans through all macro values used in CBMC proofs and computes a patch wrapping any define statement affected into an ifndef statement.

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/tools/cbmc/README.md
+++ b/tools/cbmc/README.md
@@ -55,6 +55,11 @@ Installing CBMC
 Setting up the proofs
 ---------------------
 
+Change into the `patches` directory. Run
+```
+python compute_patch.py
+```
+
 Change into the `proofs` directory. On Windows, run
 ```
 python make-common-makefile.py
--- a/tools/cbmc/patches/patch.py
+++ b/tools/cbmc/patches/patch.py
@@ -3,24 +3,18 @@
 import subprocess
 import os
 import sys
+from glob import glob
 
 if os.path.isfile('patched'):
     sys.exit()
 
-patchfiles = [filename for filename in os.listdir('.')
-              if filename.endswith('.patch') ]
-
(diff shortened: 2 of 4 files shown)


tools/cbmc/patches/patch.py
```diff
@@ -3,24 +3,18 @@
...
-
-tmpdir = 'c:\\temp\\' if sys.platform == 'win32' else '/tmp/'
-for patchfile in patchfiles:
-    tmpfile = tmpdir + patchfile
-    print("Patch file: {}".format(patchfile))
-    with open(patchfile) as inp, open(tmpfile, "w", newline="\n") as outp:
-        outp.write(inp.read())
-    cwd = os.getcwd()
-    result = subprocess.run(["git", "apply", tmpfile], cwd='../../..')
+for tmpfile in glob("*.patch"):
```
markrtuttle: Check that this code works under both bash and visual studio command prompt.

mmuesly: I have run this code successfully in a 64-bit visual studio command prompt.

tools/cbmc/README.md
```diff
@@ -55,6 +55,11 @@ Installing CBMC
 Setting up the proofs
 ---------------------
 
+Change into the `patches` directory. Run
```
mmuesly: @karkhaz

These lines are added to address your comment. As you say this is not sufficient, could you describe in more detail, what kind of user documentation enhancement you envision?

karkhaz: Thanks, I completely missed this for some reason!

mmuesly:
closed in favor of  #653

PR #602, opened 2019-05-07
Add CBMC harnesses for non default functions

mmuesly:
Description
-----------

While eARPGetCacheEntryByMac and FreeRTOS_PrintARPCache are compiled out by default,
the provided proofs ensure that they are memory safe, if compiled in on occasion.

### DIFF
--- a/tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/Makefile.json
+++ b/tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/Makefile.json
@@ -1,0 +1,17 @@
+{
+  "ENTRY": "ARPGetCacheEntryByMac",
+  "CBMCFLAGS":
+  [
+      "--unwind 7",
+      "--nondet-static"
+  ],
+  "OBJS":
+  [
+    "$(ENTRY)_harness.goto",
+    "$(FREERTOS)/lib/FreeRTOS-Plus-TCP/source/FreeRTOS_ARP.goto"
+  ],
+  "DEF":
+  [
+    "ipconfigUSE_ARP_REVERSED_LOOKUP=1", "ipconfigARP_CACHE_ENTRIES=6"
+  ]
+}
--- a/tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/README.md
+++ b/tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/README.md
@@ -1,0 +1,4 @@
+ARPGetCacheEntryByMac is memory safe,
+if it is enabled.
+
+ARPGetCacheEntryByMac does not use multiple configurations internally.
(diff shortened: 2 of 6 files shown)


tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/README.md
```diff
@@ -0,0 +1,2 @@
+This harness proves the memory safety of the ARPGetCachEntryByMac function,
```
karkhaz: ARPGetCach**e**EntryByMac

tools/cbmc/proofs/ARP/ARP_FreeRTOS_PrintARPCache/README.md
```diff
@@ -0,0 +1,4 @@
+The FreeRTOS_PrintARPCache_harness.c harness
+proves that the FreeRTOS_PrintARPCache is memory safe,
+if included and given that vLoggingPrintf
```
markrtuttle: In these readme files, the language I would use is "X is memory safe, assuming vLoggingPrintf is correct and memory safe."  

And also explicitly say that there is no need to check multiple configurations as you have been doing with other entry points.

tools/cbmc/proofs/ARP/ARPGetCacheEntryByMac/Makefile.json
```diff
@@ -0,0 +1,16 @@
+{
+  "ENTRY": "ARPGetCacheEntryByMac",
+  "CBMCFLAGS":
+  [
+      "--unwind 7"
```
karkhaz: Since using `--nondet-static` seems to be the norm for FreeRTOS, please could you write a comment explaining why you don't use it here?

mmuesly: There is no specific reason so I added the `--nondet-static` flag.

tools/cbmc/proofs/ARP/ARP_FreeRTOS_PrintARPCache/README.md
```diff
@@ -0,0 +1,4 @@
+FreeRTOS_PrintARPCache_harness.c is memory safe,
+assuming vLoggingPrintf is correct and memory safe.
```
dcgaws: I would not assume that.

mmuesly: We might put vLoggingPrintf on our task list, if you are concerned about it and get a harness strengthening the assumption. But we cannot include the vLoggingPrintf to the current harness. This harness abstracts away the vLoggingPrintf function. Including the function requires to deal with many loops that will lead to performance problems.

PR #622, opened 2019-05-09
Feature/short localname support

huguesBouvier:
Short and complete local name support

Description
-----------
Reworked BLE drivers in ESP and NORDIC to support short local name in advertisement
Added more advertisement possibilities and tweaks
Added a callback for fully user parameter able advertisement

Checklist:
----------

- [ ] My code is Linted.


modules/ports/ble/include/bt_hal_manager_adapter_ble.h
```diff
@@ -79,17 +79,21 @@ typedef enum
     BTAdvNonconnInd,
 } BTAdvProperties_t;
 
+#define BTGattAdvNameNone     0
+#define BTGattAdvNameShort    1
+#define BTGattAdvNameComplete 2
+typedef uint8_t BTGattAdvName;
```
ravibhagavandas: This could be an enum type.

PR #632, opened 2019-05-15
Don't assume that Thing name is a static string define (see #219).

dcgaws:
Per #219 the analogous fix to our Greengrass Discovery library was previously made. This fix allows the Defender client code to use Thing names that are determined at runtime rather than statically.

### DIFF
--- a/lib/defender/aws_defender.c
+++ b/lib/defender/aws_defender.c
@@ -59,6 +59,11 @@ static TaskHandle_t xDefenderTaskHandle = NULL;
 /* Timeout period for MQTT connections. */
 static TickType_t xMQTTTimeoutPeriodTicks = pdMS_TO_TICKS( 10U * 1000U );
 
+/* Device-specific MQTT topic for defender metrics. */
+#define defenderMQTT_TOPIC_PUBLISH      "$aws/things/%s/defender/metrics/cbor"
+#define defenderMQTT_TOPIC_ACCEPTED     "$aws/things/%s/defender/metrics/cbor/accepted"
+#define defenderMQTT_TOPIC_REJECTED     "$aws/things/%s/defender/metrics/cbor/rejected"
+
 /**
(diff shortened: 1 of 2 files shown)


lib/defender/aws_defender.c
```diff
@@ -276,11 +281,42 @@ static DefenderState_t prvStateSubscribe( void )
...
 }
 
+static char * prvGetDefenderMqttTopicString( char * pcTopicTemplate )
+{
+    char * pcTopic = NULL;
+    uint32_t ulTopicLength = 0;
+    uint32_t ulCharsWritten = 0;
+
+    /* Determine the maximum resulting string length. */
+    ulTopicLength = 1 + strlen( pcTopicTemplate ) + strlen( clientcredentialIOT_THING_NAME );
```
alexa-noxon: This should subtract off the characters for the '%s' included in the formatting string... otherwise, when it goes to zero out the last character further down, there could be a couple extra chars at the end of the topic length that are unintended.

alexa-noxon: Actually snprintf will put the null terminator in the correct place... so we can both drop the extra 2 characters and remove the part zeroing out the last character

lib/defender/aws_defender.c
```diff
@@ -276,11 +281,41 @@ static DefenderState_t prvStateSubscribe( void )
...
+
+    if( NULL != pcTopic )
+    {
+        /* Build the topic string including the device name. */
+        ulCharsWritten = snprintf( pcTopic,
+                                   ulTopicLength,
+                                   pcTopicTemplate,
+                                   clientcredentialIOT_THING_NAME );
+
+        if( ulCharsWritten >= ulTopicLength )
```
alexa-noxon: If snprintf fails it returns a negative number.

Should we just check if (ulCharsWritten != ulTopicLength) --> https://github.com/aws/amazon-freertos/pull/635

PR #653, opened 2019-05-22
Add prepare.py for automated preparation of the code for CBMC proofs [includes: #595]

mmuesly:
Description
-----------

The changes in this PR depend on #595. Therefore #595 is included in this PR and will be closed in favor of this PR for eliminating merge dependencies.

The prepare.py script will run all required scripts to generate Makefiles, required patches and cbmc-batch.yaml files. Once this is done, the code base is ready for running any of the proofs.

Checklist:
----------

- [x] My code is Linted.


tools/cbmc/proofs/make_proof_makefiles.py
```diff
@@ -139,7 +144,12 @@ def dump_makefile(dyr, system):
             makefile[name] = " ".join(new_value)
         else:
             makefile[name] = compute(value, so_far, system, name, dyr)
-
+    
+    if (("EXPECTED" not in makefile.keys()) or
+            makefile["EXPECTED"].lower() == "true"):
```
karkhaz: Over here you're checking for `"true"`, but on [line 148 of `make_configuration_directories.py`](https://github.com/aws/amazon-freertos/pull/653/files#diff-db83c98f72f65488a4d3ad843aee5664R148) you're dumping `True`. Is this a mistake?

tools/cbmc/proofs/make_remove_makefiles.py
```diff
@@ -0,0 +1,44 @@
...
+
+from make_cbmc_batch_files import remove_cbmc_yaml_files
+
+def main():
+    try:
+        os.remove("Makefile.common")
+    except FileNotFoundError:
+        pass
+
+    for root, _, files in os.walk("."):
```
karkhaz: Suggest
```
# We do not want to remove hand-written Makefiles, so
# only remove Makefiles that are in the same directory as
# a Makefile.json. Such Makefiles are generated from the
# JSON file.
```

tools/cbmc/proofs/make_proof_makefiles.py
```diff
@@ -139,7 +144,12 @@ def dump_makefile(dyr, system):
             makefile[name] = " ".join(new_value)
         else:
             makefile[name] = compute(value, so_far, system, name, dyr)
-
+    
+    if (("EXPECTED" not in makefile.keys()) or
+            str(makefile["EXPECTED"]).lower() == "true"):
```
mmuesly: I made it an explicit string cast.

tools/cbmc/proofs/make_proof_makefiles.py
mmuesly: In line 148, I use a boolean primitive. So it must be a capital `T` there.

Before reaching this line, `compute()` is called which does a string cast in [L178 of make_proof_makefiles.py](https://github.com/aws/amazon-freertos/blob/a3b46a51f6c276e00c32b5ca3f9223420fd1737b/tools/cbmc/proofs/make_proof_makefiles.py#L178)

That is the magic, why it work here. I will add explicit string cast to point it out.

PR #658, opened 2019-05-23
Adds option for goto-instrument flags specification

Baha:
Description
-----------
This change adds the option for goto-instrument flags specification. These were not required until now, but the task pool code base is full of volatile variables whose behavior cannot be correctly modeled unless the flag `--nondet-volatile` is passed to goto-instrument.

Checklist:
----------
- [x] Changes have been tested.


Baha:
@karkhaz

PR #663, opened 2019-05-24
CBMC memory-safety proof for DNSgetHostByName_a

dlohar:
Description
-----------
This PR includes the CBMC proof for memory safety of the DNSgetHostByName_a function, which prepares and sends a message to a DNS server along with callback.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -420,6 +420,7 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
+if(pcHostName) {
 
 	/* If the supplied hostname is IP address, convert it to uint32_t
 	and return. */
@@ -482,6 +483,7 @@ TickType_t xIdentifier = 0;
 	}
 
 	return ulIPAddress;
+  }
 }
 /*-----------------------------------------------------------*/
 
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -420,7 +420,7 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
-if(pcHostName) {
+if(pcHostName != NULL) {
 
 	/* If the supplied hostname is IP address, convert it to uint32_t
 	and return. */
@@ -481,9 +481,10 @@ if(pcHostName) {
 	{
 		ulIPAddress = prvGetHostByName( pcHostName, xIdentifier, xReadTimeOut_ms );
 	}
-
-	return ulIPAddress;
   }
+
+  return ulIPAddress;
+
 }
 /*-----------------------------------------------------------*/
 
--- a/tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
+++ b/tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
@@ -1,0 +1,78 @@
+/* FreeRTOS includes. */
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "list.h"
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_DNS.h"
+#include "FreeRTOS_IP_Private.h"
+
+/* This proof assumes the length of pcHostName is bounded by MAX_HOSTNAME_LEN and the size of UDPPayloadBuffer is bounded by 
+MAX_REQ_SIZE. This also abstracts the concurrency. */ 
+
+void *safeMalloc(size_t xWantedSize) { /* This returns a NULL pointer if the requested size is 0. */
+	if(xWantedSize == 0) {
+		return NULL;
+	}
+	uint8_t byte;
+	return byte ? malloc(xWantedSize) : NULL;
+}
+
+/* Abstraction of pvPortMalloc which calls safemalloc internally. */
+void *pvPortMalloc(size_t xWantedSize) {
+	return safeMalloc(xWantedSize);
+}
+
+/* Abstraction of FreeRTOS_GetUDPPayloadBuffer.
+We always return MAX_REQ_SIZE bytes to keep the proof performant.
+This is safe because:
+- If the caller requested more bytes, then there is a risk that they
+    will write past the end of the returned buffer. This proof
+    therefore shows that the code is memory safe even if
+    xRequestedSizeBytes > MAX_REQ_SIZE.
+- If the caller requested fewer bytes, then they will not be
+    iterating past the end of the buffer anyway.*/
+void * FreeRTOS_GetUDPPayloadBuffer(size_t xRequestedSizeBytes, TickType_t xBlockTimeTicks ) {
+	void *pvReturn = safeMalloc(MAX_REQ_SIZE);
+	return pvReturn;
+}
+
+/* Abstraction of FreeRTOS_socket. This abstraction allocates a memory of size Socket_t. */
+Socket_t FreeRTOS_socket( BaseType_t xDomain, BaseType_t xType, BaseType_t xProtocol ){
+	Socket_t xCreatedSocket = safeMalloc(sizeof(Socket_t)); // replacing malloc with safeMalloc
+	return xCreatedSocket;
+}
+
+/* This function FreeRTOS_gethostbyname_a only uses the return value of prvParseDNSReply. Hence it returns an unconstrained uint32 value */
+uint32_t prvParseDNSReply( uint8_t *pucUDPPayloadBuffer, size_t xBufferLength, TickType_t xIdentifier) { }
+
+/* Abstraction of xTaskResumeAll from task pool. This also abstracts the concurrency. */
+BaseType_t xTaskResumeAll(void) { }
+
+/* The function func mimics the callback function.*/
+void func(const char * pcHostName, void * pvSearchID, uint32_t ulIPAddress) { }
+
+typedef struct xDNS_Callback {
+	TickType_t xRemaningTime;		/* Timeout in ms */
+	FOnDNSEvent pCallbackFunction;	/* Function to be called when the address has been found or when a timeout has beeen reached */
+	TimeOut_t xTimeoutState;
+	void *pvSearchID;
+	struct xLIST_ITEM xListItem;
+	char pcName[ 1 ];
+} DNSCallback_t;
+
+void harness() {
+	FOnDNSEvent pCallback = func;
+	TickType_t xTimeout;
+	void *pvSearchID;
+	size_t len;
+	__CPROVER_assume(len >= 0 && len <= MAX_HOSTNAME_LEN);
+	char *pcHostName = safeMalloc(len); // replacing malloc with safeMalloc
+	if (len && pcHostName) {
+		pcHostName[len-1] = NULL;
+	}
+	if (pcHostName) { // Guarding against NULL pointer
+		FreeRTOS_gethostbyname_a(pcHostName, pCallback, pvSearchID, xTimeout);
+	}
+}
--- a/tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
+++ b/tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
@@ -8,10 +8,11 @@
 #include "FreeRTOS_DNS.h"
 #include "FreeRTOS_IP_Private.h"
 
-/* This proof assumes the length of pcHostName is bounded by MAX_HOSTNAME_LEN and the size of UDPPayloadBuffer is bounded by 
-MAX_REQ_SIZE. This also abstracts the concurrency. */ 
+/* This proof assumes the length of pcHostName is bounded by MAX_HOSTNAME_LEN and the size of UDPPayloadBuffer is bounded by
+MAX_REQ_SIZE. This also abstracts the concurrency. */
 
-void *safeMalloc(size_t xWantedSize) { /* This returns a NULL pointer if the requested size is 0. */
+void *safeMalloc(size_t xWantedSize) { /* This returns a NULL pointer if the requested size is 0.
+	The implementation of malloc does not return a NULL pointer instead returns a pointer for which there is no memory allocation. */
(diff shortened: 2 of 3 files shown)


tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+
+uint32_t prvParseDNSReply( uint8_t *pucUDPPayloadBuffer, size_t xBufferLength, TickType_t xIdentifier) { }
+
+Socket_t FreeRTOS_socket( BaseType_t xDomain, BaseType_t xType, BaseType_t xProtocol ){
+	Socket_t xCreatedSocket = malloc(sizeof(Socket_t));
+	return xCreatedSocket;
+}
+
+void * FreeRTOS_GetUDPPayloadBuffer(size_t xRequestedSizeBytes, TickType_t xBlockTimeTicks ) {
+	void *pvReturn = malloc(50);
```
karkhaz: What is 50?

I wonder whether the returned buffer is supposed to be of length `xRequestedSizeBytes`, or some related number.

Otherwise, if 50 is a random number, please write a comment saying so. However, I would rather you find the RFC for UDP, and figure out if there is a maximum size for UDP payloads, and use that or something related (and write a comment with a link to the RFC).

If the number needs to be kept low for performance reasons, again: please write a comment saying so. Even then, I still think it would be worth modifying this so that it returns *up to* n bytes, where n could be zero.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+	return pvReturn;
+}
+
+void *pvPortMalloc(size_t xWantedSize) {
+	void *pvSize = malloc(xWantedSize);
+	return pvSize;
```
karkhaz: I'm not requesting any changes for this now, but a broader discussion we need to have is: can malloc fail. If it can, we should be ensuring memory safety even in that case.

We might build up a nice library of utility functions, similar to what aws-c-common has, that includes a might-fail mallocator.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+} DNSCallback_t;
+
+BaseType_t xTaskResumeAll( void ) {
+	BaseType_t xRet;
+	return xRet;
+}
+void func(const char * pcHostName, void * pvSearchID, uint32_t ulIPAddress) {}
+
+void harness() {
+	if (ipconfigDNS_USE_CALLBACKS==1) {
```
karkhaz: space around `==` please.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+		FOnDNSEvent pCallback = func;
+		TickType_t xTimeout;
+		void *pvSearchID;
+		char *pcHostName = malloc(10);
```
karkhaz: Same comment as [here](https://github.com/aws/amazon-freertos/pull/665#discussion_r288116679).

tools/cbmc/proofs/DNS/DNSgetHostByName_a/Makefile.json
```diff
@@ -0,0 +1,23 @@
+{
+  "ENTRY": "DNSgetHostByName_a",
+  ################################################################
+	# This configuration flag sets callback to 1
+  "callback": 1,
+  "CBMCFLAGS": 
+  [
+  		"--unwind 1",
```
karkhaz: Please don't use tabs here.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/Makefile.json
```diff
@@ -0,0 +1,23 @@
...
+  		"--unwindset prvCreateDNSMessage.0:11,prvCreateDNSMessage.1:11,prvGetHostByName.0:11,prvProcessDNSCache.0:5,strlen.0:11,__builtin___strcpy_chk.0:11,strcmp.0:11,xTaskResumeAll.0:11,xTaskResumeAll.1:11",
```
karkhaz: I see a lot of `11` here. Is this because it's one more than the length of `pcHostName`?

If so, please define a variable to use both in the Makefile and the C file to avoid duplicating numbers.

```json
"MAX_HOSTNAME_LEN": 10,
"HOSTNAME_UNWIND": "__eval {MAX_HOSTNAME_LEN} + 1",
...
"--unwindset prvCreateDNSMessage.0:{HOSTNAME_UNWIND}",
...
"DEF":
[
  "MAX_HOSTNAME_LEN={MAX_HOSTNAME_LEN}"
]
```
```c
char *pcHostName = malloc(MAX_HOSTNAME_LEN);
```

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+Socket_t FreeRTOS_socket(BaseType_t xDomain, BaseType_t xType, BaseType_t xProtocol){
+	Socket_t xCreatedSocket = malloc(sizeof(Socket_t));
+	return xCreatedSocket;
+}
+
+void * FreeRTOS_GetUDPPayloadBuffer(size_t xRequestedSizeBytes, TickType_t xBlockTimeTicks ) {
+	__CPROVER_assume(xRequestedSizeBytes >= 0 && xRequestedSizeBytes <= MAX_REQ_SIZE);
+	void *pvReturn = malloc(MAX_REQ_SIZE);
+	return pvReturn;
+}
```
karkhaz: This function doesn't work for several reasons:
* You're assuming that `xRequestedSizeBytes` is between 0 and `MAX_REQ_SIZE`, but you don't actually do anything with `xRequestedSizeBytes` after that.
* More worrying: `MAX_REQ_SIZE` is defined as `100`. So if the code actually calls this function with the value `101` for `xRequestedSizeBytes`, then this becomes `assume(false)`. So that makes the whole proof unsound.

@tautschnig: we were hoping to have a discussion about CBMC warnings at some point, and `assume(false)`-type mistakes do worry me somewhat. I'm not confident that I can catch all of these during review.

I'm not sure, but I think what you want to do here is
```c
void * FreeRTOS_GetUDPPayloadBuffer(size_t xRequestedSizeBytes, TickType_t xBlockTimeTicks ) {
  size_t len;
  __CPROVER_assume(len >= 0 && len <= xRequestedSizeBytes);
  void *pvReturn = malloc(len);
  return pvReturn;
}
```
(I'm not sure whether `len` should be `>=` or `>` 0, if it's the latter then it's of course possible that this function will return a null pointer. Not sure if this is what you want.)

Also, I have a question: is it really possible for the original function to return a buffer that is *smaller* than the requested size? Please confirm this by reading the code, and write a comment above this function explaining why this might happen, e.g. "this function will return a buffer of size up to xRequestedSizeBytes. It may return fewer bytes if the UDP payload was smaller than xRequestedSizeBytes". (I made this up, but please confirm what the original code does).

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,62 @@
...
+
+void func(const char * pcHostName, void * pvSearchID, uint32_t ulIPAddress) {}
+
+void harness() {
+	if (ipconfigDNS_USE_CALLBACKS == 1) {
+		FOnDNSEvent pCallback = func;
+		TickType_t xTimeout;
+		void *pvSearchID;
+		size_t len;
+		__CPROVER_assume(len > 0 && len <= MAX_HOSTNAME_LEN);
```
karkhaz: I think it's important to check for zero-length hostnames too. I advocate
```c
__CPROVER_assume(len >= 0 && len <= MAX_HOSTNAME_LEN);
...
if(len)
  pcHostName[len-1] = NULL;
```

tools/cbmc/proofs/DNS/DNSgetHostByName_a/cbmc-batch.yaml
```diff
@@ -0,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: '=--object-bits;7;--32;--bounds-check;--pointer-check;--unwind;1;--unwindset;prvProcessDNSCache.0:5,strlen.0:11,__builtin___strcpy_chk.0:11,strcmp.0:11,xTaskResumeAll.0:11,xTaskResumeAll.1:11;--nondet-static;--unwinding-assertions='
```
karkhaz: @markrtuttle here's another reason not to commit these files. The `Makefile.json` has a nice variable called `HOSTNAME_UNWIND` which is computed as `1 + MAX_HOSTNAME_LEN`, and the variable is used throughout the file. But in `cbmc-batch.yaml`, we just have the magic number `11` everywhere. This will certainly go out of date if we update `Makefile.json`, I highly doubt we'll remember to change `11` to the correct value.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
tautschnig: > @tautschnig: we were hoping to have a discussion about CBMC warnings at some point, and assume(false)-type mistakes do worry me somewhat. I'm not confident that I can catch all of these during review.

We should actually include a report of assumptions in CBMC's output. Discussed with others yesterday, I'll try to make a stab at it over the next days.

karkhaz: That would be excellent, thank you! But hopefully the assumptions would (also) be machine-readable, or at least in a format that we can grep for. Having them for human consumption only works if somebody regularly reads the output. In a CI environment, the proof should ideally break if something changes that makes an assumption become false.

dlohar: If I make `len <= xRequestedSizeBytes` cbmc is just stuck while running propositional reduction. It generated 3303 VCCs and 2397 were remaining after simplification. It is not able to prove/disprove anything even in 2 hrs.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -420,6 +420,7 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
+configASSERT(pcHostName);
```
dan4thewin: This needs to be a boolean expression.  Likewise for the `configASSERT(pcName)` that occurs later in this file.  It's pedantic, I know.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,80 @@
...
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_DNS.h"
+#include "FreeRTOS_IP_Private.h"
+
+/* This proof assumes the length of pcHostName is bounded by MAX_HOSTNAME_LEN and the size of UDPPayloadBuffer is bounded by 
+MAX_REQ_SIZE. This also abstracts the concurrency. */ 
+
+void *safeMalloc(size_t xWantedSize) {
```
dan4thewin: I'd like to understand how this wrapper is different from CBMC's default malloc behavior.  Can you add a comment and maybe a reference?

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -0,0 +1,80 @@
...
+	TickType_t xRemaningTime;		/* Timeout in ms */
+	FOnDNSEvent pCallbackFunction;	/* Function to be called when the address has been found or when a timeout has beeen reached */
+	TimeOut_t xTimeoutState;
+	void *pvSearchID;
+	struct xLIST_ITEM xListItem;
+	char pcName[ 1 ];
+} DNSCallback_t;
+
+void harness() {
+	if (ipconfigDNS_USE_CALLBACKS == 1) {
```
dan4thewin: I see ipconfigDNS_USE_CALLBACKS referenced in the Makefile.json.  If we know we always set it true, why test?  If there is any possibility of it being false, shouldn't the harness protest loudly?

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
dlohar: I have added a comment here.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
dlohar: I have added an `if` statement here instead of `configASSERT`.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
dlohar: I have removed the `if` from harness. At this point it is redundant.

dlohar:
@dan4thewin I have addressed your comments.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -482,6 +483,7 @@ TickType_t xIdentifier = 0;
 	}
 
 	return ulIPAddress;
+  }
```
dan4thewin: Maybe I'm reading this wrong, but the return should come after this brace, yes?  Before, when I said it should be a boolean, I meant `( pcHostName != NULL )`, so this code exhibits the same shortcoming.

tools/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
dan4thewin: The comment you added described this function's behavior, but not how it compares to the unwrapped CBMC malloc.  That comparison is what I'd like to see.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
dlohar: Yes, the brace comes before the return. Sorry for this. `if (pcHostName)` is equivalent to the what you said, but I changed it to the explicit `NULL` check.

dan4thewin: MISRA rule 14.4: "The controlling expression of an if statement and the controlling expression of an iteration-statement shall have essentially Boolean type."

PR #665, opened 2019-05-24
CBMC memory-safety proof for DNSlookup

dlohar:
Description
-----------
This PR includes the CBMC proof for memory safety of the DNSlookup function, which looks for the indicated host name in the DNS cache

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -1338,6 +1338,8 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 	BaseType_t xFound = pdFALSE;
 	uint32_t ulCurrentTimeSeconds = ( xTaskGetTickCount() / portTICK_PERIOD_MS ) / 1000;
 	static BaseType_t xFreeEntry = 0;
+	configASSERT(pcName);
+
 
 		/* For each entry in the DNS cache table. */
 		for( x = 0; x < ipconfigDNS_CACHE_ENTRIES; x++ )
--- a/tools/cbmc/proofs/DNS/DNSlookup/DNSlookup_harness.c
+++ b/tools/cbmc/proofs/DNS/DNSlookup/DNSlookup_harness.c
@@ -1,0 +1,32 @@
+/* FreeRTOS includes. */
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "list.h"
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_DNS.h"
+#include "FreeRTOS_IP_Private.h"
+
+/* This assumes that the length of the hostname is bounded by MAX_HOSTNAME_LEN. */
+void *safeMalloc(size_t xWantedSize) {
+	if(xWantedSize == 0) {
+		return NULL;
+	}
+	uint8_t byte;
+	return byte ? malloc(xWantedSize) : NULL;
+}
+
+void harness() {
+	if(ipconfigUSE_DNS_CACHE != 0) {
+		size_t len;
+		__CPROVER_assume(len >= 0 && len <= MAX_HOSTNAME_LEN);
+		char *pcHostName = safeMalloc(len); /* malloc is replaced by safeMalloc */
+		if (len && pcHostName) {
+			pcHostName[len-1] = NULL;
+		}
+		if (pcHostName) { /* guarding against NULL pointer */
+			FreeRTOS_dnslookup(pcHostName);	
+		}
+	}	
+}
--- a/tools/cbmc/proofs/DNS/DNSlookup/Makefile.json
+++ b/tools/cbmc/proofs/DNS/DNSlookup/Makefile.json
@@ -1,0 +1,26 @@
+{
+  "ENTRY": "DNSlookup",
+  ################################################################
+  # This configuration uses DNS cache and the MAX_HOSTNAME_LEN is set to 255 according to the specification
+  "MAX_HOSTNAME_LEN": 255,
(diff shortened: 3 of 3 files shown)


tools/cbmc/proofs/DNS/DNSlookup/DNSlookup_harness.c
```diff
@@ -0,0 +1,18 @@
...
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_DNS.h"
+#include "FreeRTOS_IP_Private.h"
+
+// This assumes that the length of the hostname is 10 and it is null terminated.
+void harness() {
+	if( ipconfigUSE_DNS_CACHE != 0 ) {
+		char *pcHostName = malloc(10);
```
karkhaz: We're going to discuss assumptions today. Does it make sense to test for the maximum permissable hostname? The [RFC for hostnames](https://tools.ietf.org/html/rfc1123#section-2) says that 255-character hostnames are allowed.

If this would totally kill performance, then perhaps you might test all hostnames up to length 10 (or whatever), plus hostnames of length 255, plus hostnames of length 256? You might use @mmuesly's configuration work to support this.

tools/cbmc/proofs/DNS/DNSlookup/Makefile.json
```diff
@@ -0,0 +1,23 @@
+{
+  "ENTRY": "DNSlookup",
+  ################################################################
+  # This configuration flag uses DNS cache
+  "USE_CACHE":1,
+  "CBMCFLAGS": 
+  [
+  		"--unwind 1",
+  		"--unwindset prvProcessDNSCache.0:5,strcmp.0:11",
```
karkhaz: Is the strcmp unwind limit 11 because it's one more than the length of `pcHostName`?

If so, please define a variable to use both in the Makefile and the C file to avoid duplicating numbers.

```json
"MAX_HOSTNAME_LEN": 10,
"HOSTNAME_UNWIND": "__eval {MAX_HOSTNAME_LEN} + 1",
...
"--unwindset prvCreateDNSMessage.0:{HOSTNAME_UNWIND}",
...
"DEF":
[
  "MAX_HOSTNAME_LEN={MAX_HOSTNAME_LEN}"
]
```
```c
char *pcHostName = malloc(MAX_HOSTNAME_LEN);
```

tools/cbmc/proofs/DNS/DNSlookup/cbmc-batch.yaml
```diff
@@ -0,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: '=--object-bits;7;--32;--bounds-check;--pointer-check;--unwind;1;--unwindset;prvProcessDNSCache.0:5;--unwinding-assertions='
```
karkhaz: Shouldn't `unwindset` also include `strcmp.0:{HOSTNAME_UNWIND}`, as it does in `Makefile.json`?

@markrtuttle could we please make a policy decision about whether to commit `cbmc-batch.yaml` files? I can't remember if we decided either way. I don't think this is the first time that we haven't updated a `cbmc-batch.yaml` file after making changes to the source.

I advocate that we do not commit these files, to prevent this kind of mistake. I've been trying to look out for them, and I caught this one, but these semicolon-delimited lists are difficult to read and compare with `Makefile.json`. It's entirely possible that I've let something slip in a different review, this worries me.

If you would like to continue committing these files, then I advocate adding a "linter" exclusively for our proofs. I could write that. One thing it would check is that `cbmc-batch.yaml` and `Makefile.json` match. If the linter fails then it blocks the PR.

tools/cbmc/proofs/DNS/DNSlookup/DNSlookup_harness.c
```diff
@@ -0,0 +1,20 @@
...
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_DNS.h"
+#include "FreeRTOS_IP_Private.h"
+
+/* This assumes that the length of the hostname is bounded by MAX_HOSTNAME_LEN. */
+void harness() {
+	if(ipconfigUSE_DNS_CACHE != 0) {
+		size_t len;
+		__CPROVER_assume(len > 0 && len <= MAX_HOSTNAME_LEN);
```
karkhaz: I think it's important to check for zero-length hostnames too. I advocate
```c
__CPROVER_assume(len >= 0 && len <= MAX_HOSTNAME_LEN);
...
if(len)
  pcHostName[len-1] = NULL;
```

lib/FreeRTOS-Plus-TCP/source/FreeRTOS_DNS.c
```diff
@@ -1338,6 +1338,8 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 	BaseType_t xFound = pdFALSE;
 	uint32_t ulCurrentTimeSeconds = ( xTaskGetTickCount() / portTICK_PERIOD_MS ) / 1000;
 	static BaseType_t xFreeEntry = 0;
+		configASSERT(pcName);
```
karkhaz: please unindent

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1338,6 +1338,8 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 	BaseType_t xFound = pdFALSE;
 	uint32_t ulCurrentTimeSeconds = ( xTaskGetTickCount() / portTICK_PERIOD_MS ) / 1000;
 	static BaseType_t xFreeEntry = 0;
+	configASSERT(pcName);
```
karkhaz: @RichardBarry are you happy with this change?

PR #667, opened 2019-05-24
New CBMC header for TCB structs

Baha:
Description
-----------
This PR adds a new CBMC header that declares some TCB structs which are currently defined in `tasks.c`. It also changes the last include path on `MakefileCommon.json` to be absolute, since otherwise the headers will not be found when proofs are grouped into a subfolder.

Checklist:
----------

- [x] Changes have been tested.
- [x] Updates README.md with a line for `tcb.h`


Baha:
@karkhaz @markrtuttle

karkhaz:
Closing in favour of an approach that moves `task.c` declarations into a private header file, which is `#include`d by `task.c` and the CBMC proofs. PR for that soon come.

PR #686, opened 2019-05-30
Final folder structure

tgsong:


demos/shadow/aws_iot_demo_shadow.c
```diff
@@ -477,7 +477,7 @@ static int _initializeDemo( void )
 /*-----------------------------------------------------------*/
 
 /**
- * @brief Clean up the MQTT library and the Shadow library.
+ * @brief Clean up the the MQTT library and the Shadow library.
```
smanas546: minor: "the" is repeated twice

directories.txt
```diff
@@ -1,15 +0,0 @@
-This file describes the directories found at the root of the Amazon FreeRTOS
```
smanas546: Do we have to update it with latest information? Where is this used?

PR #688, opened 2019-05-30
Add CBMC Stub Library and queue proof infrastructure

mmuesly:
CBMC Stub Library

Description
-----------
While working on the proofs for linear containers,
I figured out that there are a couple of stubs and function models I have to reuse in different harnesses. Therefore, I started a CBMC Stub library for keeping a single model for each of these stubs.
This library is still work in progress but I appreciate some early feedback on my assumptions.

 
Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
+++ b/tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
@@ -1,0 +1,47 @@
+#include "FreeRTOS.h"
+#include "task.h"
+#include "tasksStubs.h"
+
+#ifndef TASK_STUB_COUNTER
+	#define TASK_STUB_COUNTER 0;
+#endif
+
+/* 5 is a magic number, but we need some number here as a default value.
+   This value is used to bound any loop depending on xTaskCheckForTimeOut
+   as a loop bound. It should be overwritten in the Makefile.json adapting
+   to the performance requirements of the harness. */
+#ifndef TASK_STUB_COUNTER_LIMIT
+	#define TASK_STUB_COUNTER_LIMIT 5;
+#endif
+
+
+static BaseType_t xCounter = TASK_STUB_COUNTER;
+static BaseType_t xCounterLimit = TASK_STUB_COUNTER_LIMIT;
+
+BaseType_t xTaskGetSchedulerState( void )
+{
+	return xState;
+}
+
+/* This function is another method apart from overwritting the defines to init the max
+   loop bound. */
+void vInitTaskCheckForTimeOut(BaseType_t maxCounter, BaseType_t maxCounter_limit)
+{
+	xCounter = maxCounter;
+	xCounterLimit = maxCounter_limit;
+}
+
+/* This is mostly called in a loop. For CBMC, we have to bound the loop
+   to a max limits of calls. Therefore this Stub models a nondet timeout in
+   max TASK_STUB_COUNTER_LIMIT iterations.*/
+BaseType_t xTaskCheckForTimeOut( TimeOut_t * const pxTimeOut, TickType_t * const pxTicksToWait ) {
+	++xCounter;
+	if(xCounter == xCounterLimit)
+	{
+		return pdTRUE;
+	}
+	else
+	{
+		return nondet();
+	}
+}
--- a/tools/cbmc/include/queue_init.h
+++ b/tools/cbmc/include/queue_init.h
@@ -1,0 +1,129 @@
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "queue_datastructure.h"
+
+/* Using prvCopyDataToQueue together with prvNotifyQueueSetContainer
+   leads to a problem space explosion. Therefore, we use this stub
+   and a sepearted proof on prvCopyDataToQueue to deal with it.
+   As prvNotifyQueueSetContainer is disabled if configUSE_QUEUE_SETS != 1,
+   in other cases the original implementation should be used. */
+#if( configUSE_QUEUE_SETS == 1 )
+	BaseType_t prvCopyDataToQueue( Queue_t * const pxQueue, const void *pvItemToQueue, const BaseType_t xPosition )
+	{
+		if(pxQueue->uxItemSize > ( UBaseType_t ) 0)
+		{
+			__CPROVER_assert(__CPROVER_r_ok(pvItemToQueue, ( size_t ) pxQueue->uxItemSize), "pvItemToQueue region must be readable");
+			if(xPosition == queueSEND_TO_BACK){
+				__CPROVER_assert(__CPROVER_w_ok(( void * ) pxQueue->pcWriteTo, ( size_t ) pxQueue->uxItemSize), "pxQueue->pcWriteTo region must be writable");
+			}else{
+				__CPROVER_assert(__CPROVER_w_ok(( void * ) pxQueue->u.xQueue.pcReadFrom, ( size_t ) pxQueue->uxItemSize), "pxQueue->u.xQueue.pcReadFrom region must be writable");
+			}
+			return pdFALSE;
+		}else
+		{
+			return nondet_BaseType_t();
+		}
+	}
+#endif
+
+/* xQueueCreateSet is compiled out if configUSE_QUEUE_SETS != 1.*/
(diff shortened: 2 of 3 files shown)


tools/cbmc/proofs/CBMCStubLibrary/portableMemoryAllocation.c
```diff
@@ -0,0 +1,17 @@
...
+/* xWantedSize is not bounded in this function,
+but there might be a need to bound it in the future.
+In theory, CBMC malloc allows to allocate an
+arbitray ammount of data.*/
+void *pvPortMalloc( size_t xWantedSize ){
+	if(xWantedSize == 0){
+		return NULL;
+	}
+	uint8_t byte;
+	return byte ? malloc(xWantedSize) : NULL;
```
feliperodri: Shouldn't you use a version of `malloc` that considers it could fail (i.e., return `NULL`)? I agree with your comment about the need to bound `xWantedSize`, see our version of [bounded_malloc](https://github.com/awslabs/aws-c-common/blob/48995ff6da2a4547430e74e3415b28346956166c/.cbmc-batch/source/proof_allocators.c#L69). Perhaps you could included it in this code base (another PR), and then use it here.

lib/include/FreeRTOS.h
```diff
@@ -241,6 +241,13 @@ extern "C" {
 	#define configASSERT_DEFINED 1
 #endif
 
+#ifndef configPRECONDITION
```
feliperodri: What is this for? I'd add a comment.

tools/cbmc/proofs/CBMCStubLibrary/queue_init.h
```diff
@@ -0,0 +1,119 @@
...
+   leads to a problem space explosion. Therefore, we use this stub
+   and a sepearted proof on prvCopyDataToQueue to deal with it.
+   As prvNotifyQueueSetContainer is disabled if configUSE_QUEUE_SETS != 1,
+   in other cases the original implementation should be used. */
+#if( configUSE_QUEUE_SETS == 1 )
+	BaseType_t prvCopyDataToQueue( Queue_t * const pxQueue, const void *pvItemToQueue, const BaseType_t xPosition )
+	{
+		if(pxQueue->uxItemSize > ( UBaseType_t ) 0)
+		{
+			__CPROVER_assert(__CPROVER_r_ok(pvItemToQueue, ( size_t ) pxQueue->uxItemSize), "Must be readable");
```
feliperodri: `s/"Must be readable"/"prvCopyDataToQueue: pvItemToQueue region must readable"/g` (the point here is to have more meaningful debug messages). It applies for all the other `__CPROVER_assert` statements

tools/cbmc/proofs/CBMCStubLibrary/queue_init.h
```diff
@@ -0,0 +1,119 @@
...
+			if(xPosition == queueSEND_TO_BACK){
+				__CPROVER_assert(__CPROVER_w_ok(( void * ) pxQueue->pcWriteTo, ( size_t ) pxQueue->uxItemSize), "Must be readable");
```
feliperodri: `s/readable/writable/g`

tools/cbmc/proofs/CBMCStubLibrary/queue_init.h
```diff
@@ -0,0 +1,119 @@
...
+			}else{
+				__CPROVER_assert(__CPROVER_w_ok(( void * ) pxQueue->u.xQueue.pcReadFrom, ( size_t ) pxQueue->uxItemSize), "Must be readable");
+			}
+			return pdFALSE;
+		}else
+		{
+			return nondet();
```
feliperodri: `nondet` what? I would renamed this function to explicitly says which nondet object is returning.

tools/cbmc/proofs/CBMCStubLibrary/queue_init.h
```diff
@@ -0,0 +1,119 @@
...
+	}
+#endif
+
+/* Create a mostly unconstrained Queue but bound the max item size.
+   This is required for performance reasons in CBMC at the moment. */
+QueueHandle_t xUnconstrainedQueueBoundedItemSize( UBaseType_t uxItemSizeBound ) {
+	UBaseType_t uxQueueLength;
+	UBaseType_t uxItemSize;
+	uint8_t ucQueueType;
+	__CPROVER_assume(uxQueueLength > 0);
```
feliperodri: Is `uxQueueLength` defined as integer rather than `size_t`?

tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef TASK_STUB_COUNTER
+	#define TASK_STUB_COUNTER 0;
+#endif
+
+/* 5 is a magic number, but we need some number here as a default value.
+   This value is used to bound any loop depending on xTaskCheckForTimeOut
+   as a loop bound. */
+#ifndef TASK_STUB_COUNTER_LIMIT
+	#define TASK_STUB_COUNTER_LIMIT 5;
```
feliperodri: Could you justify 5 by saying the verification time goes to the roof if you use 10 for instance? (justifying by performance)

tools/cbmc/proofs/CBMCStubLibrary/portableMemoryAllocation.c
mmuesly: `pvPortMalloc` is used in the codebase instead of direct calls to malloc. It is up to the third-party code to provide an implementation. Pure malloc is not used in the code base. Therefore, I think as long as `pvPortMalloc` allows to fail, this is a sufficient good model for the FreeRTOS code base.

tools/cbmc/proofs/CBMCStubLibrary/queue_init.h
mmuesly: It depends on the definition of UBaseType_t. But it is close to never defined as size_t. It is always something unsigned.

tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
```diff
@@ -0,0 +1,46 @@
...
+#endif
+
+
+static BaseType_t xCounter = TASK_STUB_COUNTER;
+static BaseType_t xCounterLimit = TASK_STUB_COUNTER_LIMIT;
+
+BaseType_t xTaskGetSchedulerState( void )
+{
+	return xState;
```
Baha: What about `return nondet_bool() ? pdTRUE : pdFALSE` instead? The scheduler is just supposed to return only one of those values, and it seems xState is not used anywhere else.

tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/* This function is another method apart from overwritting the defines to init the max
+   loop bound. */
+void init_TaskCheckForTimeOut(BaseType_t maxCounter, BaseType_t maxCounter_limit)
```
Baha: This function does not follow the naming conventions.

tools/cbmc/proofs/CBMCStubLibrary/tasksStubs.c
mmuesly: Interesting, it is used like this ```xTaskGetSchedulerState() == taskSCHEDULER_SUSPENDED``` in queue.c. I will use xState in proof harnesses later.

Baha: You are right, it is supposed to return 3 different values, and these depends on two different flags (`xSchedulerRunning` and `uxSchedulerSuspended`).

mmuesly:
Dropped pvPortMalloc and vPortFree from this PR in favor or  #774.

karkhaz:
@dcgaws @dan4thewin this PR contains a bunch of utility functions that other proofs call into, to initialize various data structures. I've had a look and the assumptions seem sane to me, but I'd appreciate it if you would confirm that---thank you.

PR #690, opened 2019-05-30
Fix MediaTek demo build break in #650.

dcgaws:
PR #650 broke the MediaTek demo project build. This commit fixes it.


demos/common/devmode_key_provisioning/aws_dev_mode_key_provisioning.c
```diff
@@ -410,7 +410,6 @@ void vDevModeKeyProvisioning( void )
     {
         configPRINTF( ( "ERROR: the vDevModeKeyProvisioning function requires a valid device private key.\r\n" ) );
         configASSERT( pdFALSE );
-        return;
```
cobusve: This looks strange? When ASSERT returns this return would not be reachable, but with a release build the assert will do nothing, this will not return and the code will continue while the check is trying to prevent this?

Are you sure we want this to be removed?

dcgaws: No. Good point. Please see latest.

cobusve: Ok that looks more like what I expected :)

PR #694, opened 2019-05-30
Update to mbedTLS 2.16.0.

dcgaws:
Merge mbedTLS 2.16.0 into lib\third_party\mbedtls.


lib/third_party/mbedtls/library/x509_crt.c
```diff
@@ -2014,7 +2014,7 @@ static int x509_crt_find_parent_in(
 {
     int ret;
     mbedtls_x509_crt *parent, *fallback_parent;
```
cobusve: There is also a path where parent is accessed unitialized, specifically when candidates is passed in as null.

Seems to me the best approach here is to initialize all of these locals here?

alexa-noxon:
We should consider adding the new configs that mbedTLS has added since previous releases (we can keep them commented out, but just put the comments from them in the config file in our repo so that people know they exist)

https://github.com/ARMmbed/mbedtls/commits/development/include/mbedtls/config.h

Skimming the code there seems to be at least MBEDTLS_ECP_RESTARTABLE

PR #697, opened 2019-05-31
Moving some code temporarily to headers for proofs

mmuesly:
Header Extraction Script
-----------
For CBMC proofs it is sometimes required to move some of hidden type declarations into a header to make them accessible for proofs. Due to limitations arising from the way how CBMC interacts with the code base during stubbing, it is not easily possible to use the FreeRTOS test way of injecting code using the include mechanism to the file. As it is explicitly not desired to move these declarations permanently to private headers, I suggest in this PR a script for moving them temporarily to header files. There is also some code provided to reverse the operation for future commits, in case the code in the header has been changed during proof debugging.

Checklist:
----------

- [ ] My code is Linted.


tools/cbmc/proofs/make_create_header_files.py
```diff
@@ -0,0 +1,144 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import os
+import re
+
+CWD = os.getcwd()
+
+CREATE_HEADERS_FOR = ["lib/FreeRTOS/queue.c"]
```
Baha: This list could be obtained from a JSON file.

tools/cbmc/proofs/make_create_header_files.py
```diff
@@ -0,0 +1,144 @@
...
+    with open(absolut_file, "w") as source:
+        source.write(source_modified)
+    if header_name:
+        absolut_header = get_absolute_header(header_name, file)
+        print("Create:", absolut_header)
+        with open(absolut_header, "w") as out:
+            print("#ifndef QUEUE_DATASTRUCTURE_H", file=out)
+            print("#define QUEUE_DATASTRUCTURE_H", file=out)
+            out.write(header_content)
+            print("#endif /* QUEUE_DATASTRUCTURE_H */", file=out)
```
Baha: Could this definition (`QUEUE_DATASTRUCTURE_H`) be obtained automatically from the file? I guess they could (from the start marker) and then it should work for any header in the list.

tools/cbmc/proofs/make_create_header_files.py
```diff
@@ -0,0 +1,144 @@
...
+                         header_name))
+    else:
+        absolut_header = os.path.normpath(
+            os.path.join(FreeRTOS_HEAD,
+                         *FreeRTOS_INCLUDE.split("/"),
+                         header_name))
+    return absolut_header
+
+def create_header(file):
+    absolut_file = os.path.join(FreeRTOS_HEAD, *file.split("/"))
```
Baha: Minor detail, but there are several occurrences of "absolut_*" in the file which can be replaced by "absolute_*".

RichardBarry:
Please also add a comment above the special CBMC comments in the source code to say what they are - I would be left scratching my head as a reader otherwise.

mmuesly:
Depends on a new CBMC flag, proposed in: https://github.com/diffblue/cbmc/pull/4790

tools/cbmc/proofs/make_type_header_files.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
dan4thewin: This desperately needs documentation.  How is this different from using the C preprocessor to dump the macros?

mmuesly: This script doesn't only dump the macros but also includes type declarations extracted from a previously with goto-cc compiled binary. After reading about the C preprocessor flag for gcc, it might be possible to invoke the C preprocessor of some compilers for gathering the macros and merge this with the output of the -dump-c-type-header flag. Using python to grep for the macros makes the script c preprocessor independent.

karkhaz:
@dan4thewin I've added documentation to the new script---it's a string near the top of the file, and which is also dumped when you pass `-h`. Also made other improvements, so now ready for re-review. Thank you!

tools/cbmc/proofs/make_type_header_files.py
karkhaz: I've added documentation in the `epilog()` function below, thanks!

PR #704, opened 2019-06-01
OTA Callback changes for custom and secondary processor jobs - Final

mtalreja-a:
OTA Callback changes for custom and secondary processor jobs
-----------

Modifying the OTA Agent to use callback structure instead of directly calling PAL functions. This allows customers to pass in custom callbacks for the PAL functions.

1. Updated agent to pass file ID in callback signatures that did not have file context.
2. Created default callbacks for the new signatures to keep the code backwards compatible.
3. Saved the file ID in the OTA agent context so that it can be passed to callbacks.
4. Renamed OTA_AgentInit_v2 to OTA_AgentInit_internal
5. Self test check does not check for version numbers if the file ID is not 0 - it just updates the image state to eOTA_ImageState_Testing
6. Added code to check for custom job callback return code and respond back with a success.

Checklist:
----------

- [ ] My code is Linted.


libraries/freertos_plus/aws/ota/include/aws_ota_agent.h
```diff
@@ -271,6 +271,151 @@ typedef struct
...
+typedef int16_t (* pxOTAPALWriteBlockCallback_t)( OTA_FileContext_t * const C, uint32_t iOffset, uint8_t * const pacData, uint32_t iBlockSize );
+
+/**
+ * @brief Custom Job callback function typedef.
+ *
+ * The user may register a callback function when initializing the OTA Agent. This
+ * callback will be called when the OTA agent cannot parse a job document.
+ *
+ * @param[in] pcJSON Pointer to the json document received by the OTA agent
+ * @param[in] ulMsgLen Length of the json document received by the agent
```
ravibhagavandas: Minor: Missing documentation for the return value types for all the callbacks.

libraries/freertos_plus/aws/ota/include/aws_ota_agent.h
```diff
@@ -271,6 +271,151 @@ typedef struct
...
+ */
+typedef OTA_JobParseErr_t (* pxOTACustomJobCallback_t)( const char * pcJSON, uint32_t ulMsgLen );
```
ravibhagavandas: Do we need to also pass the File Context for the callback to populate it ?

mtalreja-a: Since it's a custom job, there is no afr_ota key. The callback just returns the entire json doc without any parsed details. What will we put in the file context if we send it back?

libraries/freertos_plus/aws/ota/include/aws_ota_agent.h
mtalreja-a: Thanks, @pvyawaha has a bunch of other documentation changes planned.

libraries/freertos_plus/aws/ota/include/aws_ota_agent.h
ravibhagavandas: Ok, I was curious as to how this callback will be used, as the line [2606](https://github.com/aws/amazon-freertos/pull/704/files#diff-6778a459de8ac6acebe32ec87c57b11dR2606),  immediately after the callback checks if the context has a valid job name and fails if the job name is null. Shouldn't the job name be populated from callback?

PR #705, opened 2019-06-03
only compile ble if enabled

qiutongs:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/demo_runner/iot_demo_afr.c
+++ b/demos/demo_runner/iot_demo_afr.c
@@ -31,12 +31,16 @@
 #include "aws_clientcredential.h"
 #include "iot_demo_logging.h"
 #include "platform/iot_network_afr.h"
-#include "platform/iot_network_ble.h"
 #include "iot_network_manager_private.h"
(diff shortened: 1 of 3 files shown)


demos/demo_runner/iot_demo_afr.c
```diff
@@ -31,13 +31,17 @@
...
 #include "platform/iot_network_afr.h"
-#include "platform/iot_network_ble.h"
 #include "iot_network_manager_private.h"
 #include "platform/iot_threads.h"
 #include "aws_demo.h"
 #include "iot_init.h"
 #include "iot_mqtt.h"
 
+#if BLE_ENABLED
+    #include "platform/iot_network_ble.h"
```
ravibhagavandas: This could still cause a redefinition error/warning if both BLE and WIFI are included. Can we instead create a new intializer for BLE instead of AWS_IOT_NETWORK_SERVER_INFO_AFR_INITIALIZER in iot_network_ble.h

PR #721, opened 2019-06-06
Tests Update: Removed one wifi test.  Moved TCP thread safety tests to the end of the group.

xuelix:
Checklist:
----------

- [ ] My code is Linted.


tests/common/wifi/aws_test_wifi.c
```diff
@@ -2253,14 +2345,15 @@ static void prvConnectionTask( void * pvParameters )
...
-                      "Task %d failed to connect to the AP %s with error code %d.\r\n",
+                      "FAILURE: Task %d failed to connect to the AP %s with error code %d.\r\n",
                       pxTaskParams->usTaskId,
                       xTestNetworkParams.pcSSID,
                       xWiFiConnectStatus );
             configPRINTF( ( pxTaskParams->cStatusMsg ) );
             break;
         }
 
+        configPRINTF( ( "INFO: Task %d connected to AP %s.\r\n", pxTaskParams->usTaskId, xTestNetworkParams.pcSSID ) );
```
alexa-noxon: do we want different levels of print statement (like OTA has) for log and INFO?

tests/common/wifi/aws_test_wifi.c
```diff
@@ -2446,86 +2543,3 @@ static void prvConnectionTask( void * pvParameters )
...
-        xTaskConnectDisconnectSyncEventGroupHandle = xEventGroupCreate();
-        configASSERT( xTaskConnectDisconnectSyncEventGroupHandle != NULL );
-
-        /* Create the task that connect at the same time. */
-        for( usIndex = 0; usIndex < testwifiNUM_TASKS; usIndex++ )
-        {
-            xTaskParams[ usIndex ].usTaskId = usIndex;
-            xTaskParams[ usIndex ].xWiFiStatus = eWiFiFailure;
-            xTaskCreateResult = xTaskCreate(
-                prvConnectionTask, "WiFiConnectionTask", testwifiTASK_STACK_SIZE,
```
sarenameas: Thank you for removing this function. I have the same comment as the other PR, that prvConnectionTask() function above also needs to be removed since it is used only in this test.

xuelix: ok, will update

tests/common/wifi/aws_test_wifi.c
```diff
@@ -805,6 +810,9 @@ TEST( Full_WiFi, AFQP_WIFI_SetMode_InvalidMode )
     {
         xWiFiStatus = WIFI_SetMode( eWiFiModeNotSupported );
         TEST_WIFI_ASSERT_OPTIONAL_API( xWiFiStatus != eWiFiSuccess, xWiFiStatus );
+    } else
```
sarenameas: The else needs to be on a separate line according to our coding standard.

tests/common/wifi/aws_test_wifi.c
```diff
@@ -1023,6 +1031,9 @@ TEST( Full_WiFi, AFQP_WIFI_GetHostIP_InvalidDomainName )
 
         /* Assert that the IP address is NOT found. */
         TEST_ASSERT_EQUAL_INT32( *( ( uint32_t * ) ucIPAddr ), 0 );
+    } else
```
sarenameas: I now remember why I did not put a `else { TEST_FAIL(); }` here. It is because WIFI_GetHostIP( testwifiTEST_INVALID_DOMAIN_NAME, ucIPAddr ); might have an assert inside for us passing in an invalid domain. Then it was difficult to determine if that was indeed a failure case. I gave the libraries room to use an assert for invalid parameters and pass.

tests/common/wifi/aws_test_wifi.c
xuelix: This was taken from 1.5 release branch

PR #728, opened 2019-06-06
Add CBMC patch to remove volatile qualifier from some variables

Baha:
Description
-----------
The proposed patch removes the volatile qualifier from a few static global variables declared in the tasks.c file. This patch will be necessary for the majority of the incoming task pool management proofs.

### DIFF
--- a/tools/cbmc/patches/0005-Remove-volatile-qualifier-from-tasks-variables.patch
+++ b/tools/cbmc/patches/0005-Remove-volatile-qualifier-from-tasks-variables.patch
@@ -1,0 +1,76 @@
+This patch removes the volatile qualifier from some global variable
+declarations in the tasks.c file.  For task pool proofs, we are running
+`goto-instrument` with the `--nondet-volatile` flag on, which causes reads from
+volatile variable to be nondeterministic (i.e., any possible value can be
+returned).  This is actually the desired behavior for volatile variables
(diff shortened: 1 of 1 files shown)


RichardBarry:
Can we please write a comment in each place that a volatile has been removed to say why it is safe for the volatile to be removed.  For example, uxPendedTicks is incremented in potentially nested interrupts and decremented in multiple tasks - so we need to show an analysis of why it is ok to remove volatile, and if removing volatile is ok, also change the code that takes a non volatile copy in one location for optimization.  We also need to ensure that any future use of now non-volatile variables is not going to introduce issues accidentally as people may assume the variable is volatile.

karkhaz:
Thanks Richard. Just to clarify---this particular PR isn't actually modifying the FreeRTOS kernel in the source tree. It is actually a patch that we apply to the kernel just before we run CBMC over it in continuous integration, but the code in the repository is not affected.

Nevertheless, your suggestion to add comments to the patch is a good one, it will help to document why it's okay to remove `voliatile` in each case.

Baha:
Thanks @RichardBarry and @karkhaz for the suggestions.

I have now updated the patch file with an extensive comment regarding the volatile behavior in CBMC and why it is a problem for task pool proofs. Let me know if you have any questions.

dcgaws:
This fails one of the CBMC checks.

PR #744, opened 2019-06-07
Updated the MQTT endpoint variable name as updated in the latest branch.

HarnishaA:
Updated the MQTT endpoint variable name as updated in the latest branch.

Description
-----------
Updated the MQTT endpoint variable name as updated in the latest branch in aws_credentials.h file. 

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
+++ b/tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
@@ -222,7 +222,7 @@ class OtaAfrProject:
         """
         self.__setIdentifierInFile(
             {
-                '#define clientcredentialMQTT_BROKER_ENDPOINT ': '\"' + awsIotEndpoint + '\"',
+                '#define clientcredentialMQTT_BROKER_ENDPOINT': '\"' + awsIotEndpoint + '\"',
                 '#define clientcredentialMQTT_BROKER_PORT' : awsIotEndpointPort
             },
             os.path.join(self._projectRootDir, OtaAfrProject.CLIENT_CREDENTIAL_PATH)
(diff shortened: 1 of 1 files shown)


tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
```diff
@@ -219,7 +222,7 @@ def setClientCredentialsForAwsIotEndpoint(self, awsIotEndpoint, awsIotEndpointPo
         """
         self.__setIdentifierInFile(
             {
-                'static const char clientcredentialMQTT_BROKER_ENDPOINT[] =': '\"' + awsIotEndpoint + '\";',
+                '#define clientcredentialMQTT_BROKER_ENDPOINT ': '\"' + awsIotEndpoint + '\"',
```
sarenameas: __setIdentifierInFile() already inserts a space after the key, although I think it does not hurt to have one here.

tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
```diff
@@ -134,6 +136,7 @@ def buildProject(self):
             print('====> Command run completed with the return code: ', proc.returncode)
             returnCodes.append(proc.returncode)
 
+        sleep(5)
```
sarenameas: What is this sleep for? Is to give the process time to generate binaries after returning?

HarnishaA: Yeah .. It is causing the test to fail saying could not find output binary even after its actually present. Just added sleep to test. If this works i'll add proper handling for this scenario.

tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
HarnishaA: Will remove it in the next commit to keep the consistency :-)

tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
sarenameas: Maybe we can add a comment if you plan to keep it?

HarnishaA: Added the comment in the below commit :-)

PR #747, opened 2019-06-08
fix IDE project files for adding tcp demos

wnyuiot:
fix IDE project files for adding tcp demos

-----------

re-run the IDE generator to update all the project files

Checklist:
----------

- [ ] My code is Linted.


qiutongs:
We already updated some demo projects for tcp demos. For me, it is NXP(IAR), Renesas, TI (IAR).

Is this the plan agreed by Manasa?

You also update test projects. I feel risky that we update so many projects at this point.

wnyuiot:
@qiutongs  After discussing with Manasa, we agreed that updating all projects at once is risky. Cancel this PR. Thanks

PR #748, opened 2019-06-08
Update tasks.c

KiLuYa:
If you use "taskRESET_READY_PRIORITY", 
there will be a duplicated judgement about whether "pxReadyTasksLists[ pxMutexHolderTCB->uxPriority ]" is empty.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/lib/FreeRTOS/tasks.c
+++ b/lib/FreeRTOS/tasks.c
@@ -3980,7 +3980,7 @@ TCB_t *pxTCB;
 				{
 					if( uxListRemove( &( pxMutexHolderTCB->xStateListItem ) ) == ( UBaseType_t ) 0 )
 					{
-						taskRESET_READY_PRIORITY( pxMutexHolderTCB->uxPriority );
+						portRESET_READY_PRIORITY( pxMutexHolderTCB->uxPriority );
 					}
 					else
 					{


lib/FreeRTOS/tasks.c
```diff
@@ -3980,7 +3980,7 @@ TCB_t *pxTCB;
 				{
 					if( uxListRemove( &( pxMutexHolderTCB->xStateListItem ) ) == ( UBaseType_t ) 0 )
 					{
-						taskRESET_READY_PRIORITY( pxMutexHolderTCB->uxPriority );
+						portRESET_READY_PRIORITY( pxMutexHolderTCB->uxPriority );
```
aggarg: As mentioned on one of your other PRs, a task may have been removed from delayed list and therefore a check of ready list is needed before calling ```portRESET_READY_PRIORITY```. This change, therefore, cannot be done. 

Thanks.

KiLuYa: But in this case, before we call taskRESET_READY_PRIORITY, we can confirm it must be in the ready list, 
image

aggarg: Yes, I missed that and you are right. Thanks for pointing it out. We will make this update in the master copy of the kernel, so it won't be visible here for a while. 

Thanks.

KiLuYa: You're welcome.
By the way, the same issue happen in another place of task.c:
task.c  -- vTaskPriorityDisinheritAfterTimeout  --  Line 4185

aggarg:
This has been change in the SVN repo and will be included in the next release of FreeRTOS: https://sourceforge.net/p/freertos/code/2670/

Thank you for reporting this.

PR #758, opened 2019-06-11
Add CBMC harness for QueueGenericCreate

mmuesly:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/include/cbmc.h
+++ b/tools/cbmc/include/cbmc.h
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
 
 #define __CPROVER_printf_ptr(var) { uint8_t *ValueOf_ ## var = (uint8_t *) var; }
 #define __CPROVER_printf2_ptr(str,exp) { uint8_t *ValueOf_ ## str = (uint8_t *) (exp); }
+
+/* CBMC assert to test pvPortMalloc result when xWantedSize is 0. Mostly used to report
+ * full coverage on pvPortMalloc, but use with caution as it might complicate debugging
+ */
+#define __CPROVER_assert_zero_allocation() __CPROVER_assert( pvPortMalloc(0) == NULL, "pvPortMalloc allows zero-allocated memory.")
+
+/* xWantedSize is not bounded in this function, but there might be a need to bound it in the future.
+ * In theory, CBMC malloc allows to allocate an arbitrary amount of data. This will not be true for
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
+{
+	if ( xWantedSize == 0 )
+	{
+		return NULL;
+	}
+	return nondet_bool() ? malloc( xWantedSize ) : NULL;
+}
+
+void vPortFree( void *pv )
+{
+	free(pv);
+}
--- a/tools/cbmc/proofs/Queue/QueueGenericCreate/Configurations.json
+++ b/tools/cbmc/proofs/Queue/QueueGenericCreate/Configurations.json
@@ -1,0 +1,68 @@
+{
+  "ENTRY": "QueueGenericCreate",
+  "CBMCFLAGS":
+  [
+      "--unwind 1",
+      "--signed-overflow-check",
+      "--pointer-overflow-check",
+      "--unsigned-overflow-check"
+
+  ],
+  "OBJS":
+  [
+    "$(ENTRY)_harness.goto",
+    "$(FREERTOS)/freertos_kernel/list.goto",
+    "$(FREERTOS)/freertos_kernel/queue.goto"
+  ],
+  "DEF":
+  [
+    {
+      "QueueGenericCreate_default": [
+        "configUSE_TRACE_FACILITY=0",
+        "configGENERATE_RUN_TIME_STATS=0",
+        "configUSE_MUTEXES=1",
+        "'mtCOVERAGE_TEST_MARKER()=__CPROVER_assert(1, \"Coverage marker\")'",
+        "configSUPPORT_STATIC_ALLOCATION=1",
+        "configUSE_QUEUE_SETS=0",
+        "configSUPPORT_DYNAMIC_ALLOCATION=1"
+      ]
+    },
+    {
+      "QueueGenericCreate_noMutex": [
+        "configUSE_TRACE_FACILITY=0",
+        "configGENERATE_RUN_TIME_STATS=0",
+        "configUSE_MUTEXES=0",
+        "configUSE_RECURSIVE_MUTEXES=0",
+        "'mtCOVERAGE_TEST_MARKER()=__CPROVER_assert(1, \"Coverage marker\")'",
+        "configSUPPORT_STATIC_ALLOCATION=1",
+        "configUSE_QUEUE_SETS=0",
+        "configSUPPORT_DYNAMIC_ALLOCATION=1"
+      ]
+    },
+    {
+      "QueueGenericCreate_noSTATIC_ALLOCATION": [
+        "configUSE_TRACE_FACILITY=0",
+        "configGENERATE_RUN_TIME_STATS=0",
+        "configUSE_MUTEXES=1",
+        "'mtCOVERAGE_TEST_MARKER()=__CPROVER_assert(1, \"Coverage marker\")'",
+        "configSUPPORT_STATIC_ALLOCATION=0",
+        "configUSE_QUEUE_SETS=0",
+        "configSUPPORT_DYNAMIC_ALLOCATION=1"
+      ]
+    },
+    {
+      "QueueGenericCreate_useQueueSets": [
+        "configUSE_TRACE_FACILITY=0",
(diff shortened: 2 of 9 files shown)


RichardBarry:
Fantastic we are moving onto the kernel with these proofs now, however there are a few issues with these changes:

- Please don't introduce a new header file as the code builds with some 20 odd different compilers and we can't guarantee the header file is going to be there (stdint.h is not provided with all of them).  In this case I think using portMAX_DELAY in place of INT_MAX will work.

Related to:

configASSERT( uxItemSize == 0
			|| (uxQueueLength > 0 && uxItemSize >= 0 
				&& uxItemSize < (INT_MAX - sizeof(Queue_t))/uxQueueLength) );

- 0 is the value used each time a semaphore is created as semaphores have length (counting semaphores for example) but no actual storage, hence 0 is a valid value.  

- Please use a separate assert for each set of values being checked, rather than combing them in one.  This is per the coding standard (maybe not documented) and is something folks were asking about on the forum just last week.

- The statement is quite complex so please add a comment to say the condition it is checking for.  In this case I think the check to ensure the value is not above INT_MAX is probably obsolete in practice (nothing has that much memory in the MCU space), although may be required to claim formal proof.

mmuesly:
Thank you for the comment.

I think the portMAX_DELAY is a good suggestion instead of INT_MAX and does the job without further header dependencies.

Further, I minimized the assert statement to a single line and added a comment as requested. 

The new assert statement ensures that all parameters are set to a range that allow secure algebraic execution of the method without under or overflow errors. Therefore it must hold for a formal analysis.

mmuesly:
This PR depends on: #697 and #774

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
 
 #define __CPROVER_printf_ptr(var) { uint8_t *ValueOf_ ## var = (uint8_t *) var; }
 #define __CPROVER_printf2_ptr(str,exp) { uint8_t *ValueOf_ ## str = (uint8_t *) (exp); }
+
+/* CBMC assert to test pvPortMalloc result when xWantedSize is 0. Mostly used to report
```
danielsn: Is this a guarantee of the FreeRTOS system? If so, make a comment referencing where this is documented to be true? Its not true for regular malloc.

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
...
+ */
+#define __CPROVER_assert_zero_allocation() __CPROVER_assert( pvPortMalloc(0) == NULL, "pvPortMalloc allows zero-allocated memory.")
+
+/* xWantedSize is not bounded in this function, but there might be a need to bound it in the future.
+ * In theory, CBMC malloc allows to allocate an arbitrary amount of data. This will not be true for
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
+{
+	if ( xWantedSize == 0 )
```
danielsn: See above.  This is not true of malloc in general.

tools/cbmc/proofs/Queue/QueueGenericCreate/Configurations.json
```diff
@@ -0,0 +1,68 @@
+{
+  "ENTRY": "QueueGenericCreate",
+  "CBMCFLAGS":
+  [
+      "--unwind 1",
+      "--signed-overflow-check",
+      "--pointer-overflow-check",
+      "--unsigned-overflow-check"
```
danielsn: no memory safety flags?

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
```diff
@@ -0,0 +1,19 @@
...
+#include "queue.h"
+#include "queue_datastructure.h"
+#include "cbmc.h"
+
+
+void harness(){
+	UBaseType_t uxQueueLength;
+	UBaseType_t uxItemSize;
+	uint8_t ucQueueType;
+	__CPROVER_assume(uxQueueLength > ( UBaseType_t ) 0);
```
danielsn: What happens if the user tried to create a queue of length 0?  Would that be legal?

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
```diff
@@ -0,0 +1,19 @@
...
+
+	/* The implicit assumption for the QueueGenericCreate method is,
+	   that there are no overflows in the computation and the inputs are safe.
+	   There is no check for this in the code base */
```
danielsn: Then add some!

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
```diff
@@ -0,0 +1,19 @@
...
+	UBaseType_t upper_bound = portMAX_DELAY - sizeof(Queue_t);
```
danielsn: Why this upper bound? Is this for performance reasons, or required by the spec?

tools/cbmc/proofs/make_type_header_files.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python3
```
danielsn: This does not belong in this PR.  refactor into a seperate PR

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
 
 #define __CPROVER_printf_ptr(var) { uint8_t *ValueOf_ ## var = (uint8_t *) var; }
 #define __CPROVER_printf2_ptr(str,exp) { uint8_t *ValueOf_ ## str = (uint8_t *) (exp); }
+
+/* CBMC assert to test pvPortMalloc result when xWantedSize is 0. Mostly used to report
+ * full coverage on pvPortMalloc, but use with caution as it might complicate debugging
+ */
+#define __CPROVER_assert_zero_allocation() __CPROVER_assert( pvPortMalloc(0) == NULL, "pvPortMalloc allows zero-allocated memory.")
```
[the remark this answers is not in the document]
mmuesly: I think, this discussion should be part of #774.

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
...
+
+/* xWantedSize is not bounded in this function, but there might be a need to bound it in the future.
+ * In theory, CBMC malloc allows to allocate an arbitrary amount of data. This will not be true for
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
```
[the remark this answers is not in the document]
mmuesly: I think, this discussion should be part of #774.

tools/cbmc/proofs/Queue/QueueGenericCreate/Configurations.json
mmuesly: The code is written in a way, that on boot the device uses pvPortMalloc to allocate a couple of data structures for reuse which are never freed while the device is only. Therefore the memory safety flags is not delivering very meaningful results as these data structures are supposed to be not freed.

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
mmuesly: This is not yet specified. The function is supposed to be called with uxQueueLength > 0.

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
mmuesly: portMAX_DELAY is the maximum value a UBaseType can get for a specific system. So this bound is resulting from the specs.

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
danielsn: Is that documented in the mainline code, ideally as an assertion?

tools/cbmc/proofs/Queue/QueueGenericCreate/QueueGenericCreate_harness.c
danielsn: That's a really odd name for the constant.  Is it possible to make a constant with a better name?

markrtuttle:
Closed in favor of https://github.com/aws/amazon-freertos/pull/1324.

PR #759, opened 2019-06-11
Anoxon/pkcs release15

alexa-noxon:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,191 @@
...
- * 
+ *
  * \warn This code is provided for example purposes only, and
  * should not be used in production code.
  */
 
+
 #ifndef _AWS_DEV_MODE_KEY_PROVISIONING_H_
 #define _AWS_DEV_MODE_KEY_PROVISIONING_H_
-
```
sarenameas: Why did you delete this space?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -46,85 +47,500 @@
...
+        {
+            xPrivateKeyType = 0xFFFFFFFF;
+            xResult = CKR_ARGUMENTS_BAD;
+        }
     }
 
-    /* Get the default private key storage ID. */
-    if( ( 0 == xResult ) || ( CKR_CRYPTOKI_ALREADY_INITIALIZED == xResult ) )
+    /* EC Keys. */
+    if( xPrivateKeyType == CKK_EC )
```
sarenameas: Is it ok to keep going to this if if the xResult form the last block is not CKR_OK?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -46,85 +47,500 @@
...
 {
-    CK_RV xResult = 0;
-    CK_C_GetFunctionList xCkGetFunctionList = NULL;
-    CK_ULONG xCount = 1;
+    CK_RV xResult = CKR_OK;
+    CK_OBJECT_CLASS xPrivateKeyClass = CKO_PRIVATE_KEY;
+    CK_FUNCTION_LIST_PTR pxFunctionList;
+    CK_BBOOL xTokenStorage = CK_TRUE;
+    mbedtls_pk_type_t xMbedKeyType;
+    CK_KEY_TYPE xPrivateKeyType;
```
sarenameas: I think it would be nicer if all variables were initialized.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -46,85 +47,500 @@
...
+        pxModulus = pvPortMalloc( MODULUS_LENGTH + 1 );
+        pxE = pvPortMalloc( E_LENGTH + 1 );
+        pxD = pvPortMalloc( D_LENGTH + 1 );
+        pxPrime1 = pvPortMalloc( PRIME_1_LENGTH + 1 );
+        pxPrime2 = pvPortMalloc( PRIME_2_LENGTH + 1 );
+        pxExp1 = pvPortMalloc( EXPONENT_1_LENGTH + 1 );
+        pxExp2 = pvPortMalloc( EXPONENT_2_LENGTH + 1 );
+        pxCoefficient = pvPortMalloc( COEFFICIENT_LENGTH + 1 );
+
+        if( ( pxModulus == NULL ) || ( pxE == NULL ) || ( pxD == NULL ) || ( pxPrime1 == NULL ) || ( pxPrime2 == NULL ) || ( pxExp1 == NULL ) || ( pxExp2 == NULL ) )
```
sarenameas: Looks like we are missing pxCoefficient in this list?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -46,85 +47,500 @@
...
+                           CK_KEY_TYPE xPublicKeyType,
+                           uint8_t * pucPublicKeyLabel,
+                           CK_OBJECT_HANDLE_PTR pxPublicKeyHandle )
+{
+    CK_RV xResult;
+    CK_BBOOL xTrue = CK_TRUE;
+    CK_FUNCTION_LIST_PTR pxFunctionList;
+    CK_OBJECT_CLASS xClass = CKO_PUBLIC_KEY;
+
+    xResult = C_GetFunctionList( &pxFunctionList );
```
sarenameas: Do we need to check the result of this before continuing?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -43,88 +44,503 @@
...
+        if( NULL != pxCoefficient )
+        {
+            vPortFree( pxCoefficient );
+        }
     }
 
+    mbedtls_pk_free( &xMbedPkContext );
+
     return xResult;
 }
```
dan4thewin: Such a long function... Perhaps it could be split into some helpers?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -43,88 +44,503 @@
...
+        CK_BYTE xPublicExponent[] = { 0x01, 0x00, 0x01 };
+        CK_BYTE xModulus[ MODULUS_LENGTH + 1 ] = { 0 };
+
+        lMbedResult = mbedtls_rsa_export_raw( ( mbedtls_rsa_context * ) xMbedPkContext.pk_ctx,
+                                              ( unsigned char * ) &xModulus, MODULUS_LENGTH + 1,
+                                              NULL, 0,
+                                              NULL, 0,
+                                              NULL, 0,
+                                              NULL, 0 );
+        CK_ATTRIBUTE xPublicKeyTemplate[] =
```
dan4thewin: This is easy on the eyes.  Could the xPrivateKeyTemplate in the last function be populated the same way?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -43,88 +44,503 @@
...
+        size_t xLength;
+        CK_BYTE xEcPoint[ 256 ] = { 0 };
+
+        mbedtls_ecdsa_context * pxEcdsaContext;
+        pxEcdsaContext = ( mbedtls_ecdsa_context * ) xMbedPkContext.pk_ctx;
+
+        /* DER encoded EC point. Leave 2 bytes for the tag and length. */
+        lMbedResult = mbedtls_ecp_point_write_binary( &pxEcdsaContext->grp, &pxEcdsaContext->Q,
+                                                      MBEDTLS_ECP_PF_UNCOMPRESSED, &xLength,
+                                                      xEcPoint + 2, sizeof( xEcPoint ) - 2 );
```
dan4thewin: Does this agree with the comment?  Seems like a gap of 4 based on the last two arguments.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -43,88 +44,503 @@
...
+{
+    CK_RV xResult;
+    CK_MECHANISM xMechanism =
+    {
+        CKM_RSA_PKCS_KEY_PAIR_GEN, NULL_PTR, 0
+    };
+    CK_FUNCTION_LIST_PTR pxFunctionList;
+    CK_ULONG xModulusBits = 2048;
+    CK_BYTE xPublicExponent[] = { 0x01, 0x00, 0x01 };
+    CK_BYTE xSubject[] = { 0x01, 0x02 };
```
dan4thewin: For the sake of those who come after, could we use some named constants or macros?  I think it's better to avoid "magic" numbers.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -134,220 +550,297 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
     CK_OBJECT_CLASS xCertificateClass = CKO_CERTIFICATE;
+    CK_CERTIFICATE_TYPE xCertificateType = CKC_X_509;
     CK_FUNCTION_LIST_PTR pxFunctionList;
     CK_RV xResult;
     uint8_t * pucDerObject = NULL;
     int32_t lConversionReturn = 0;
     size_t xDerLen = 0;
+    CK_BBOOL xTokenStorage = CK_TRUE;
 
+    CK_BYTE xSubject[] = "TestSubject";
```
dan4thewin: A dummy value?  Do we really need to initialize the entire template in order to process pucCertificate?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -134,220 +550,297 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
+
+        xResult = xFindObjectWithLabelAndClass( xSession,
+                                                ( const char * ) pxLabel,
+                                                xClass[ uiIndex ],
+                                                &xObjectHandle );
+
+        if( ( xResult == CKR_OK ) && ( xObjectHandle != CK_INVALID_HANDLE ) )
         {
-            configPRINTF( ( "ERROR: Failed to provision certificate %d \r\n", xResult ) );
+            while( ( xResult == CKR_OK ) && ( xObjectHandle != CK_INVALID_HANDLE ) )
```
dan4thewin: Couldn't you just do `uiIndex--` instead of this nested loop?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -134,220 +550,297 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
         xResult = xProvisionDevice( xSession, xParams );
-    }
 
-    pxFunctionList->C_CloseSession( xSession );
+        pxFunctionList->C_CloseSession( xSession );
+    }
 }
 /*-----------------------------------------------------------*/
 
 void vDevModeKeyProvisioning( void )
```
dan4thewin: Silly question... What is "dev mode"?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -134,220 +550,297 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
-                                      &xSession );
+    if( xResult == CKR_OK )
+    {
+        xResult = xInitializePkcs11Session( &xSession );
+    }
 
     if( xResult == CKR_OK )
     {
+        #if ( pkcs11configIMPORT_PRIVATE_KEYS_SUPPORTED == 1 )
+            xDestroyCredentials( xSession );
```
dan4thewin: Why is this here?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1140,20 +2436,48 @@ CK_DEFINE_FUNCTION( CK_RV, C_DigestInit )( CK_SESSION_HANDLE xSession,
...
+ *                                          of the message digest.  If pDigest is NULL,
+ *                                          this value is updated to contain the length
+ *                                          of the buffer needed to hold the digest. Else
+ *                                          it is updated to contain the actual length of
+ *                                          the digest placed in pDigest.
+ *
+ * @return CKR_OK if successful.
+ * Else, see <a href="https://tiny.amazon.com/wtscrttv">PKCS #11 specification</a>
+ * for more information.
+ */
```
dan4thewin: The params above don't seem to match this function.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1415,70 +2981,444 @@ CK_DEFINE_FUNCTION( CK_RV, C_VerifyInit )( CK_SESSION_HANDLE xSession,
...
+                xAttributeMap |= LABEL;
+                break;
+
+            case ( CKA_TOKEN ):
+                memcpy( &xBool, xAttribute.pValue, sizeof( CK_BBOOL ) );
+
+                if( xBool != CK_TRUE )
+                {
+                    PKCS11_PRINT( ( "ERROR: Only token key generation is supported. \r\n" ) );
+                    xResult = CKR_ATTRIBUTE_VALUE_INVALID;
```
dan4thewin: Why is this result value different from the other cases?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1415,70 +2981,444 @@ CK_DEFINE_FUNCTION( CK_RV, C_VerifyInit )( CK_SESSION_HANDLE xSession,
...
+}
+
+/* Checks that the public key template provided for C_GenerateKeyPair
+ * contains all necessary attributes, and does not contain any invalid
+ * attributes. */
+CK_RV prvCheckGenerateKeyPairPublicTemplate( CK_ATTRIBUTE_PTR * ppxLabel,
+                                             CK_ATTRIBUTE_PTR pxTemplate,
+                                             CK_ULONG ulTemplateLength )
+{
+#define LABEL        ( 1U )
```
dan4thewin: Redefined macro?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1511,26 +3448,28 @@ CK_DEFINE_FUNCTION( CK_RV, C_GenerateKeyPair )( CK_SESSION_HANDLE xSession,
...
-    mbedtls_pk_context xCtx;
-    mbedtls_pk_init( &xCtx );
-    xResult = mbedtls_pk_setup( &xCtx, mbedtls_pk_info_from_type( MBEDTLS_PK_ECKEY ) );
-
-    if( xResult == 0 )
+    if( xResult == CKR_OK )
     {
-        /*mbedtls_ecdsa_init( xCtx.pk_ctx ); */
+        mbedtls_pk_init( &xCtx );
+        lMbedResult = mbedtls_pk_setup( &xCtx, mbedtls_pk_info_from_type( MBEDTLS_PK_ECKEY ) );
```
dan4thewin: xResult should be CKR_OK regardless of lMbedResult here?

libraries/freertos_plus/standard/utils/src/iot_pki_utils.c
```diff
@@ -0,0 +1,69 @@
...
+#include "iot_pki_utils.h"
+
+/* CRT includes. */
+#include <stdio.h>
+#include <string.h>
+
+
+/*-----------------------------------------------------------*/
+
+void PKI_mbedTLSSignatureToPkcs11Signature( uint8_t * pxSignaturePKCS,
```
dan4thewin: It would be nice to have a comment describing why we need this function.

libraries/freertos_plus/standard/utils/src/iot_pki_utils.c
```diff
@@ -0,0 +1,69 @@
...
+                                            uint8_t * pxMbedSignature )
+{
+    /* Reconstruct the signature in PKCS #11 format. */
+    uint8_t * pxNextLength;
```
dan4thewin: Null guards?  How about checks that the buffer lengths are correct?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: This comment is accurate.

https://tls.mbed.org/api/ecp_8h.html#af7d35a6631f662cc279eed8c4f0f7ba6

The second to last argument xEcPoint+2 is the pointer to the location where the data is being copied (so leaving 2 bytes at the beginning for tag and length).   The last argument sizeof(xEcPoint)-2 is the length of the data being copied, which is 2 less than the full buffer length to account for the 2 bytes for length and tag which are written on the subsequent 2 lines.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -549,307 +876,1064 @@ CK_DEFINE_FUNCTION( CK_RV, C_OpenSession )( CK_SLOT_ID xSlotID,
...
+        xResult = CKR_TEMPLATE_INCOMPLETE;
+    }
+
+    if( xResult == CKR_OK )
+    {
+        xPalHandle = PKCS11_PAL_SaveObject( pxLabel, pxCertificateValue, xCertificateLength );
+
+        if( xPalHandle == 0 ) /*Invalid handle. */
+        {
+            xResult = CKR_DEVICE_MEMORY;
```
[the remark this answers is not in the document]
alexa-noxon: Host memory is supposed to be when the computer that PKCS is running on runs out of memory, and Device Memory is supposed to be when the token runs out of memory.  

I have been thinking of it as "host memory" if we run out of freertos heap and "device memory" if there is a flash/driver issue, but I am not sure that this is completely accurate as the token/host border is quite blurred.   I will do a review for consistency of usage, as I am not positive that this mental model has not evolved over my time working on this library.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,192 @@
...
-    uint32_t ulClientCertificateLength;    /**< Length of the certificate in bytes. */
+    uint8_t * pucClientPrivateKey;      /**< Pointer to the device private key in PEM format.
+                                         *   See tools/certificate_configuration/PEMfileToCString.html
+                                         *   for help with formatting.*/
+    uint32_t ulClientPrivateKeyLength;  /**< Length of the private key data, in bytes. */
+    uint8_t * pucClientCertificate;     /**< Pointer to the device certificate in PEM format.
+                                         *   See tools/certificate_configuration/PEMfileToCString.html
+                                         *   for help with formatting.*/
+    uint32_t ulClientCertificateLength; /**< Length of the device certificate in bytes. */
+    uint8_t * pucJITPCertificate;       /**< Pointer to the JITP certificate in PEM format.  If no
```
dcgaws: Can you please expand this acronym in the comment? It would be nice to also link to the IoT Core feature in question, since that's important for the customer transition to production.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,192 @@
...
+ * \return CKR_OK if EC key pair generation succeeded.
+ * Otherwise, a positive PKCS #11 error code.
+ */
+CK_RV xProvisionGenerateKeyPairEC( CK_SESSION_HANDLE xSession,
+                                   uint8_t * pucPrivateKeyLabel,
+                                   uint8_t * pucPublicKeyLabel,
+                                   CK_OBJECT_HANDLE_PTR pxPrivateKeyHandle,
+                                   CK_OBJECT_HANDLE_PTR pxPublicKeyHandle );
+
+/** \brief Destroys AFR credentials stored in device PKCS #11 module.
```
dcgaws: Please use Amazon FreeRTOS instead of our various internal shorthands.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,192 @@
...
+*
+* \note Not all ports support the deletion of all objects.  Successful
+* function return only indicates that all objest for which destroy is
```
dcgaws: "Objects". 

Is this behavior per spec? Seems potentially annoying for the app writer, if so.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -549,307 +895,1078 @@ CK_DEFINE_FUNCTION( CK_RV, C_OpenSession )( CK_SLOT_ID xSlotID,
...
-        if( NULL != pxSession->xSignMutex )
+        else
         {
-            vSemaphoreDelete( pxSession->xSignMutex );
+            xPrivateKeyFound = CK_TRUE;
         }
 
-        /* Free the public key context if it exists. */
-        if( NULL != pxSession->xVerifyKey.pk_ctx )
+        xResult = prvCreateECPublicKey( &xMbedContext, &pxLabel, pxTemplate, ulCount, pxObject );
```
alexa-noxon: Note: There is a memory leak in prvCreateECPublicKey that I am going to work out.   Leaving this comment so I won't forget :)

libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -0,0 +1,233 @@
...
+
+#define CK_DEFINE_FUNCTION( returnType, name )             returnType name
+#define CK_DECLARE_FUNCTION( returnType, name )            returnType name
+#define CK_DECLARE_FUNCTION_POINTER( returnType, name )    returnType( CK_PTR name )
+#define CK_CALLBACK_FUNCTION( returnType, name )           returnType( CK_PTR name )
+
+/**
+ *   @brief Length of a SHA256 digest, in bytes.
+ */
+#define pkcs11SHA256_DIGEST_LENGTH           32
```
dcgaws: There's a redundant define in the crypto library header above, I think. Can one of them be eradicated?

libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * @brief Length of a curve P-256 ECDSA signature, in bytes.
+ */
+#define pkcs11ECDSA_P256_SIGNATURE_LENGTH    64
```
dcgaws: Length of just the non-encoded point values, of course. But do we actually use non-encoded points anywhere?

libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -0,0 +1,233 @@
...
+#if defined( __PASTE ) && defined( __CC_ARM )
+    /* ARM RCVT stdint.h has a duplicate definition with PKCS #11. */
+    #undef __PASTE
+#endif
+
+#include "pkcs11.h"
+
+/* Elliptic Curve Private Key Template
+ * The object class must be the first attribute in the array.
+ * https://www.cryptsoft.com/pkcs11doc/v220/group__SEC__12__3__4__ELLIPTIC__CURVE__PRIVATE__KEY__OBJECTS.html */
```
dcgaws: Could we reference the standard directly instead? This is obviously the website of a for-profit company.

libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -0,0 +1,233 @@
...
+typedef struct PKCS11_PrivateEcKeyTemplate
+{
+    CK_ATTRIBUTE xObjectClass; /* CKA_CLASS, set to CKO_PRIVATE_KEY. */
+    CK_ATTRIBUTE xKeyType;     /* CKA_KEY_TYPE, set to CKK_EC. */
+    CK_ATTRIBUTE xLabel;       /* CKA_LABEL. */
+    CK_ATTRIBUTE xEcParams;    /* CKA_EC_PARAMS, the DER encoded OID for the curve. */
+    CK_ATTRIBUTE xValue;       /* CKA_VALUE, the value d, the private value. */
```
dcgaws: You call this (capital) "D, the private exponent" below, which seems better.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,195 @@
...
+                                         *   See tools/certificate_configuration/PEMfileToCString.html
+                                         *   for help with formatting.*/
+    uint32_t ulClientCertificateLength; /**< Length of the device certificate in bytes. */
+    uint8_t * pucJITPCertificate;       /**< Pointer to the Just-In-Time Provisioning (JITP) certificate in 
+                                         *   PEM format.  If no JITP certificate exists,
+                                         *   this should be set to NULL.
+                                         *   See tools/certificate_configuration/PEMfileToCString.html
+                                         *   for help with formatting.
+                                         *   See https://tinyurl.com/y3qf3un5 for more information about 
+                                         *   getting started with J*/
```
dcgaws: Did the last line of this comment get truncated? 

Regarding the previous line, I've gotten feedback that tinyurl makes some people nervous, since you don't know what destination site it's about to take you to (unless you pre-resolve it, in which case what's the point). 

It might also be worth pointing out that this field is required if you want JITP to work, and useless if you're not using JITP.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,195 @@
...
-/** \brief Provisions device with provided credentials. 
- * 
+/** \brief Provisiong a device given a valid PKCS #11 session.
+*
+* \param[in] xSession       A valid PKCS #11 session.
+* \param[in] pxParams       Pointer to an initialized provisioning
+*                           structure.
+*
+* \return CKR_OK upon successful key creation.  PKCS #11 error code on failure.
+* Note that PKCS #11 error codes are positive.
```
dcgaws: The last line of this comment is cruft, in my opinion.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
```diff
@@ -23,87 +23,195 @@
...
+                             CK_OBJECT_HANDLE_PTR pxObjectHandle );
+
+/** \brief Generates an RSA key pair.
+ *
+ * \param[in] xSession              A valid PKCS #11 session handle.
+ * \param[in] pucPrivateKeyLabel    PKCS #11 label attribute value for private key to be created.
+ *                                  This should be a string value. See iot_pkcs11_config.h.
+ * \param[in] pucPublicKeyLabel     PKCS #11 label attribute value for public key to be created.
+ *                                  This should be a string value. See iot_pkcs11_config.h.
+ * \param[in] pxPrivateKeyHandle    Updated to provide the PKCS #11 handle for the private key
```
dcgaws: I.e. which therefore must be freed by the caller?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -43,88 +44,486 @@
...
+                                    CK_OBJECT_HANDLE_PTR pxPrivateKeyHandle,
+                                    CK_OBJECT_HANDLE_PTR pxPublicKeyHandle )
+{
+    CK_RV xResult;
+    CK_MECHANISM xMechanism =
+    {
+        CKM_RSA_PKCS_KEY_PAIR_GEN, NULL_PTR, 0
+    };
+    CK_FUNCTION_LIST_PTR pxFunctionList;
+    CK_ULONG xModulusBits = 2048;
```
dcgaws: Could this instead be based on the #defines you have above for key byte field lengths (i.e. multiplied by eight)? That would make it easier to change the whole implementation over to a different modulus size.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
dcgaws: Should be provided by the caller, then, in my opinion.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -168,186 +589,245 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
 
-            xResult = pxFunctionList->C_CreateObject( xSession,
-                                                      ( CK_ATTRIBUTE_PTR ) &xCertificateTemplate,
-                                                      sizeof( xCertificateTemplate ) / sizeof( CK_ATTRIBUTE ),
-                                                      pxObjectHandle );
-        }
+    /* Create an object using the encoded client certificate. */
+    if( xResult == CKR_OK )
+    {
+        configPRINTF( ( "Write code signing certificate...\r\n" ) );
```
dcgaws: This is for sure handling the code-signing certificate in all cases?

libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -0,0 +1,227 @@
...
+ *      digestAlgorithm DigestAlgorithmIdentifier,
+ *      digest Digest }
+ *
+ * DigestAlgorithmIdentifier :: = AlgorithmIdentifier
+ * Digest :: = OCTET STRING
+ *
+ * This is the DigestInfo sequence, digest algorithm, and the octet string/length
+ * for the digest, without the actual digest itself.
+ */
+#define pkcs11STUFF_APPENDED_TO_RSA_SIG    { 0x30, 0x31, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x01, 0x05, 0x00, 0x04, 0x20 }
```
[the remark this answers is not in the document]
alexa-noxon: As discussed, added additional comments to explain reason for padding here and at its usage in iot_tls.c

libraries/freertos_plus/standard/pkcs11/src/iot_pkcs11.c
```diff
@@ -0,0 +1,247 @@
...
+                                                 pxSession );
+    }
+
+    return xResult;
+}
+
+/*-----------------------------------------------------------*/
+#ifdef CreateMutex
+    #undef CreateMutex
+    #define CreateMutex    CreateMutex /* This is a hack because CreateMutex is redefined to CreateMutexW in synchapi.h in windows. :/ */
```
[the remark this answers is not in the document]
alexa-noxon: fixed.

libraries/freertos_plus/standard/pkcs11/src/iot_pkcs11.c
```diff
@@ -0,0 +1,247 @@
...
+{
+    CK_RV xResult;
+    CK_FUNCTION_LIST_PTR pxFunctionList;
+
+    xResult = C_GetFunctionList( &pxFunctionList );
+
+    if( xResult == CKR_OK )
+    {
+        xResult = pxFunctionList->C_OpenSession( xSlotId,
+                                                 CKF_SERIAL_SESSION | CKF_RW_SESSION,
```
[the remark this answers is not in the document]
alexa-noxon: As discussed in person, this will go on the TODO list.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -376,166 +278,591 @@ CK_RV prvMbedTLS_Initialize( void )
...
 
-    if( NULL == ppxFunctionList )
-    {
-        xResult = CKR_ARGUMENTS_BAD;
-    }
-    else
+    *ppcLabel = NULL;
+    *pxLabelLength = 0;
+
+    if( xP11Context.xObjectList.xObjects[ lIndex ].xHandle != CK_INVALID_HANDLE )
```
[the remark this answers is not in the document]
dan4thewin: So, no loop is needed here.  Is the loop in prvFindObjectInListByLabel() not needed?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -376,166 +278,591 @@ CK_RV prvMbedTLS_Initialize( void )
...
-    ( void ) ( xSlotID );
-    ( void ) ( pvApplication );
-    ( void ) ( xNotify );
+    CK_BBOOL xObjectFound = CK_FALSE;
+    int lInsertIndex = -1;
+    int lSearchIndex = pkcs11configMAX_NUM_OBJECTS - 1;
 
-    /* Check arguments. */
-    if( NULL == pxSession )
+    xGotSemaphore = xSemaphoreTake( xP11Context.xObjectList.xMutex, portMAX_DELAY );
```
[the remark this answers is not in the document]
dan4thewin: "destroy an object in the middle of an operation" - that language seems to suggest intention.  Is there anything to prevent an unintentional case where one task is using an object that another task is destroying?  If not, we might want to provide warnings around the use of destroy.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -376,166 +278,591 @@ CK_RV prvMbedTLS_Initialize( void )
...
+                PKCS11_PRINT( ( "ERROR: JITP Certificate not specified.\r\n" ) );
+            }
+        }
+        else if( 0 == memcmp( pucLabel, pkcs11configLABEL_ROOT_CERTIFICATE, strlen( ( char * ) pkcs11configLABEL_ROOT_CERTIFICATE ) ) )
+        {
+            /* Use either Verisign or Starfield root CA,
+             * depending on whether this is an ATS endpoint. */
+            if( strstr( clientcredentialMQTT_BROKER_ENDPOINT, "-ats.iot" ) == NULL )
+            {
+                *ppucCertData = ( uint8_t * ) tlsVERISIGN_ROOT_CERTIFICATE_PEM;
```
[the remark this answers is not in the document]
dan4thewin: Ah, I understand now that the ATS certs are cross-signed by Starfield.  Still, it seems like we should be using Amazon's own root certs.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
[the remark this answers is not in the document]
alexa-noxon: In this function, we are given the object handle, and the object handle is the index in the array so there is no searching needed.

The label however is not an index in the array, so in the other function I must search the labels to see if I have a match, so a for loop is needed.

This comment makes me realize that I am not checking out of bounds for the xAppHandle, so I'm going to add that now.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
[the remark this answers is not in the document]
alexa-noxon: yes sorry I should have mentioned that.  ATM we cannot always use the Amazon Root Certificate because some platforms do not trust Amazon's Root CA, but do have built in trust for the Starfield one.  I think updating device SDKs in our repo might take care of that issue, but until this has done I will leave as is.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
alexa-noxon: updated.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
alexa-noxon: no which should have a valid CK_OBJECT_HANDLE created by the caller.  I will update the wording to match that of pkcs.

demos/dev_mode_key_provisioning/include/aws_dev_mode_key_provisioning.h
alexa-noxon: updated.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: added a define for 2048.  still have some of the length parameters independently defined... I feel like I need to do more research into the causal relationship between these parameters.

vendors/pc/boards/windows/ports/pkcs11/iot_pkcs11_pal.c
```diff
@@ -156,7 +159,7 @@ CK_OBJECT_HANDLE PKCS11_PAL_SaveObject( CK_ATTRIBUTE_PTR pxLabel,
         /* Create the file. */
         hFile = CreateFileA( pcFileName,
                              GENERIC_WRITE,
-                             0,
+                             FILE_SHARE_READ,
```
dcgaws: This doesn't make sense to me. We're opening the file to blow it away and rewrite here. Sharing it for read won't result in anything good. If two threads can be calling this routine at the same time, it's a bug.

libraries/freertos_plus/standard/utils/src/iot_pki_utils.c
```diff
@@ -0,0 +1,189 @@
...
+}
+
+
+
+int PKI_pkcs11SignatureTombedTLSSignature( uint8_t * pucSig,
+                                           size_t * pxSigLen ) 
+{
+    int xResult = 0;
+    uint8_t * pucSigPtr;
+    uint8_t ucTemp[ 64 ] = { 0 }; /* A temporary buffer for the pre-formatted signature. */
```
dcgaws: This routine is specific to ECDSA?

libraries/freertos_plus/standard/utils/src/iot_pki_utils.c
```diff
@@ -0,0 +1,189 @@
...
+/* CRT includes. */
+#include <stdio.h>
+#include <string.h>
+
+#define FAILURE    ( -1 )
+
+/*-----------------------------------------------------------*/
+/* Reconstruct the signature in PKCS #11 format. */
+int PKI_mbedTLSSignatureToPkcs11Signature( uint8_t * pxSignaturePKCS,
+                                           uint8_t * pxMbedSignature )
```
dcgaws: Could the caller(s) pass in buffer length parameters as well? That would be more defensive as well as future proof for bigger curves.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: Conclusion of in person discussion: post-merge, will update xProvisionCertificate to parse the certificate subject out of the cert and add it to the template properly.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: no it is not.  I have changed the comment to just say write certificate

vendors/pc/boards/windows/ports/pkcs11/iot_pkcs11_pal.c
alexa-noxon: Yes good catch I will change that back to zero.

libraries/freertos_plus/standard/utils/src/iot_pki_utils.c
alexa-noxon: yes.  I have added tiny comments above to reference this fact (and more complete explanation is in header).

PR #761, opened 2019-06-11
Changing the Makefile template to re run --drop-unused-functions

dlohar:
Description
-----------

This scripts re runs --drop-unused-functions.

Checklist:
----------

- [ ] My code is Linted.


tools/cbmc/proofs/Makefile.template
```diff
@@ -90,7 +90,11 @@ $(ENTRY)4.goto: $(ENTRY)3.goto
 	$(GOTO_INSTRUMENT) $(INSTFLAGS) --slice-global-inits @RULE_INPUT@ @RULE_OUTPUT@ \
 		> $(ENTRY)4.txt 2>&1
 
-$(ENTRY).goto: $(ENTRY)4.goto
+$(ENTRY)5.goto: $(ENTRY)4.goto
```
karkhaz: Please write a comment explaining why it's necessary to run --drop-unused-functions twice. Otherwise it looks like a mistake.

dlohar: I have added a comment. Does it look fine or do we need more explanation?

PR #774, opened 2019-06-12
Add common stubs to CBMC header

Baha:
Description
-----------
Added pvPortMalloc, pvFree functions stubs to CBMC header, these will be used across several proofs in the future. Added macro that asserts if zero-allocation is allowed.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/include/cbmc.h
+++ b/tools/cbmc/include/cbmc.h
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
 
 #define __CPROVER_printf_ptr(var) { uint8_t *ValueOf_ ## var = (uint8_t *) var; }
 #define __CPROVER_printf2_ptr(str,exp) { uint8_t *ValueOf_ ## str = (uint8_t *) (exp); }
+
+/* CBMC assert to test pvPortMalloc result when xWantedSize is 0. Mostly used to report
+ * full coverage on pvPortMalloc, but use with caution as it might complicate debugging
+ */
+#define __CPROVER_assert_zero_allocation() __CPROVER_assert( pvPortMalloc(0) == NULL, "pvPortMalloc allows zero-allocated memory.")
+
+/* xWantedSize is not bounded in this function, but there might be a need to bound it in the future.
+ * In theory, CBMC malloc allows to allocate an arbitrary amount of data. This will not be true for
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
+{
+	if ( xWantedSize == 0 )
+	{
+		return NULL;
+	}
+	return nondet_bool() ? malloc( xWantedSize ) : NULL;
+}
+
+void vPortFree( void *pv )
+{
+	free(pv);
+}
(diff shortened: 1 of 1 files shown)


Baha:
Just rebased this PR to current master, after removing a non-related change in infrastructure that will have its own PR soon. @mmuesly @karkhaz can you review it?

Baha:
Many of @mmuesly and my proofs depend on this branch now, so it would be nice to have it merged. @lt72 @dcgaws can you please check it?

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
 
 #define __CPROVER_printf_ptr(var) { uint8_t *ValueOf_ ## var = (uint8_t *) var; }
 #define __CPROVER_printf2_ptr(str,exp) { uint8_t *ValueOf_ ## str = (uint8_t *) (exp); }
+
+/* CBMC assert to test pvPortMalloc result when xWantedSize is 0. Mostly used to report
+ * full coverage on pvPortMalloc, but use with caution as it might complicate debugging
+ */
+#define __CPROVER_assert_zero_allocation() __CPROVER_assert( pvPortMalloc(0) == NULL, "pvPortMalloc allows zero-allocated memory.")
```
danielsn: I reviewed these in some dependent PRs.  
1. The macro name only makes sense once I read the comment. It should be self-explanatory
1. Why does this lead to full coverage? That is not clear. How does it complicate debugging?
1. The string on the assert itself is incorrect.  its allowed under the description of " If size is 0, either a null pointer or a unique pointer that can be successfully passed to free() shall be returned." 
https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
...
+
+/* xWantedSize is not bounded in this function, but there might be a need to bound it in the future.
+ * In theory, CBMC malloc allows to allocate an arbitrary amount of data. This will not be true for
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
+{
+	if ( xWantedSize == 0 )
```
danielsn: Again, this is not the actual spec of malloc.  Why limit it?

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,26 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
...
+	{
+		return NULL;
+	}
+	return nondet_bool() ? malloc( xWantedSize ) : NULL;
+}
+
+void vPortFree( void *pv )
+{
+	free(pv);
```
danielsn: Is ```free(0)``` legal?

tools/cbmc/include/cbmc.h
karkhaz: This seems to be necessary because of the implementation of `pvPortMalloc` below, which always returns `NULL` when passed `0`.

I am going to remove that case from `pvPortMalloc`, so there should be no need for this macro at all. I'll delete from this PR and from any PRs that use it.

karkhaz:
Setting assignee to @Baha until I have contributor access to this fork

tools/cbmc/include/cbmc.h
```diff
@@ -25,3 +25,18 @@ enum CBMC_LOOP_CONDITION { CBMC_LOOP_BREAK, CBMC_LOOP_CONTINUE, CBMC_LOOP_RETURN
...
+ * embedded devices.
+ */
+void *pvPortMalloc( size_t xWantedSize )
+{
+	return nondet_bool() ? malloc( xWantedSize ) : NULL;
+}
+
+void vPortFree( void *pv )
+{
+	(void)pv;
```
danielsn: Why is this necessary?

PR #785, opened 2019-06-14
Add CBMC .gitignores

karkhaz:


tools/cbmc/proofs/.gitignore
```diff
@@ -0,0 +1,9 @@
+Makefile
+Makefile.common
+cbmc-batch.yaml
+**/*.txt
```
danielsn: I didn't know the ```**``` pattern.  What does it do?

PR #787, opened 2019-06-14
Put CBMC proofs of parsing funs in one directory

karkhaz:
These were the first proofs to be committed to a:FR and were put in the
top-level proofs directory. We're now grouping proofs on a per-protocol
basis.


Baha:
I would instead make a specific folder for DHCP-related proofs. That folder would contain proofs for `DHCPProcess` and `IsDHCPSocket`. The latter proof does not have a PR yet, but I can make it after merging this one.

markrtuttle:
I'd close this pull request and open a new one against the current master.  I'd use the directory Parsing in place of parsing to be consistent with the current naming scheme.

PR #805, opened 2019-06-19
Improving ble stack initialization sequence

huguesBouvier:
Improving ble stack initialization sequence

Description
-----------
Before:
In the main during initialization, if BLE test were not being run, memory for BLE would be cleared.
That memory needs to be released to allow memory hungry test like last will and testament to succeed.

The issue with that approach is that is needs to check on flags in test_runner_config.h.
This conflict with our test system and IDT.

Now:
Only BT classic memory is released.
BLE stack is initialized when network is selected.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/common/iot_tests_network.c
+++ b/tests/common/iot_tests_network.c
@@ -97,6 +97,7 @@ static bool bleEnabled = false;
         }
     }
     /*-----------------------------------------------------------*/
+    extern BTStatus_t bleStackInit( void );
 
     static BaseType_t _BLEEnable( void )
     {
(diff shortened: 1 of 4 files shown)


vendors/nordic/boards/nrf52840-dk/aws_tests/application_code/main.c
```diff
@@ -276,6 +276,15 @@ static void prvUartInit( void )
     APP_ERROR_CHECK( xErrCode );
 }
 
+/**
+ * @brief Initialize BLE stask for Nordic.
+ * On Nordic it is just a stub.
```
ravibhagavandas: Dont we have vendor specific stack initialization for Nordic ?

tests/common/iot_tests_network.c
```diff
@@ -107,12 +108,18 @@ static bool bleEnabled = false;
...
         if( bInitBLE == false )
         {
-            xStatus = IotBle_Init();
-
-            if( xStatus == eBTStatusSuccess )
-            {
-                bInitBLE = true;
-            }
+        	/* Initialize low level drivers for BLE */
+        	xStatus = bleStackInit( );
```
ravibhagavandas: Does it also make sense to teardown ble stack after all ble stacks ? That will be a good practice as we dont mandate anywhere multiple testgroups cannot be run together.

vendors/nordic/boards/nrf52840-dk/aws_tests/application_code/main.c
huguesBouvier: No, it comes with softdevice

tests/common/iot_tests_network.c
huguesBouvier: Yes but that would be a separate commit. Right now we don't do it.

PR #828, opened 2019-06-21
Wordsmithing of TI README.txt

cringti:
Adds some details behind the subset of the TI SDK provided with Amazon
FreeRTOS, as well as a link to the 2.10.00.04 SDK.

Signed-off-by: Chris Ring <cring@ti.com>


cringti:
No issue changing "suitable" to "recommended".  I'll fix that and force push an updated commit onto this branch.  (If that's not the policy for PR feedback, let me know.)

Since a specific version of the TI SDK is checked into the repo (currently 2.10.00.04), I think it's more appropriate to provide a link to download exactly that version.  The 2.10.00.04 release is now 15 months old(!), and 4 SDK releases have been made since - none of which will work without code changes in other amazon-freertos sources.  So I'm against adding a link to the "latest" TI SDK release as users will download it and fail.

I agree with your comment that the SDK versions could drift over time.  When we send PRs to update the TI SDK version, we'll need to remember to update this link appropriately.

PR #836, opened 2019-06-24
Update CFLAGS to always include -m32

mmuesly:
Description
-----------

This flag enforces that goto-cc compiles all data structures assuming 32 bits.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/proofs/Makefile.template
+++ b/tools/cbmc/proofs/Makefile.template
@@ -34,6 +34,7 @@ CFLAGS = \
   $(CFLAGS2) \
   $(C_DEF) $(O_DEF) $(H_DEF) $(DEF) \
   $(C_OPT) $(O_OPT) $(H_OPT) $(OPT) \
+  -m32 \
   # empty
 
 CBMCFLAGS = \


tools/cbmc/proofs/Makefile.template
```diff
@@ -34,6 +34,7 @@ CFLAGS = \
   $(CFLAGS2) \
   $(C_DEF) $(O_DEF) $(H_DEF) $(DEF) \
   $(C_OPT) $(O_OPT) $(H_OPT) $(OPT) \
+  -m32 \
```
dan4thewin: What was the default behavior before this change?  Do the outcomes of any of the tests change with this configuration?

markrtuttle: This change ensures that we are both building and testing the code as 32-bit code.  Prior to this, projects without this change were building 64 bit code and testing it was 32 bit code.  This was an issue on some proofs, and the authors added -m32 to the local makefile for their proofs.  This change adds -m32 to the global makefile so that all (future) proofs are checked consistently.

PR #843, opened 2019-06-25
Update compute_patch.py to pick up MakefileCommon.json

mmuesly:
Description
-----------

Enabling the auto patch script to pick up the defines from the `MakefileCommon.json` in the future.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tools/cbmc/patches/compute_patch.py
+++ b/tools/cbmc/patches/compute_patch.py
@@ -63,19 +63,24 @@ def find_all_defines():
     proof_dir = os.path.abspath(os.path.join(PATCHES_DIR, "..", "proofs"))
 
     for fldr, _, fyles in os.walk(proof_dir):
-        if not "Makefile.json" in fyles:
+        if "Makefile.json" in fyles:
+            file = os.path.join(fldr, "Makefile.json")
+            key = "DEF"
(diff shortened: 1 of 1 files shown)


tools/cbmc/patches/compute_patch.py
```diff
@@ -63,19 +63,24 @@ def find_all_defines():
     proof_dir = os.path.abspath(os.path.join(PATCHES_DIR, "..", "proofs"))
 
     for fldr, _, fyles in os.walk(proof_dir):
-        if not "Makefile.json" in fyles:
+        if "Makefile.json" in fyles:
+            file = os.path.join(fldr, "Makefile.json")
+            key = "DEF"
+        elif "MakefileCommon.json" in fyles:
+            file = os.path.join(fldr, "MakefileCommon.json")
+            key = "DEF "
```
karkhaz: Mark says that there's no reason that the key should be "`DEF `" with a trailing space in `MakefileCommon.json`, this seems like an error. (There are several other keys defined that way, including "`INC `". This doesn't actually break anything because Mark strips whitespace anyway.

I'm not asking for changes here. In a separate pull request, I will fix `MakefileCommon.json` and remove the whitespace from the key in this file.

PR #846, opened 2019-06-26
Add support for multiline definitions in patch computing

Baha:
Description
-----------
This PR adds support for multiline definitions in patch computing. This is required for some proofs in TaskPool where macros must be redefined (e.g., TaskSwitchContext).

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/patches/compute_patch.py
+++ b/tools/cbmc/patches/compute_patch.py
@@ -93,11 +93,17 @@ def manipulate_headerfile(defines, header_file):
             if (match and
                     match.group(1) in defines and
                     not last.lstrip().startswith("#ifndef")):
+                full_def = line
+                # this loop deals with multiline definitions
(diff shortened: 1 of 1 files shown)


tools/cbmc/patches/compute_patch.py
```diff
@@ -93,11 +93,18 @@ def manipulate_headerfile(defines, header_file):
             if (match and
                     match.group(1) in defines and
                     not last.lstrip().startswith("#ifndef")):
+                full_def = [line]
```
mmuesly: Why don't you use directly string concatenation instead of collecting lines in a list and joining it with an empty string?
```
full_def = line
while line.rstrip().endswith("\\"):
    full_def += next(source)
```

Baha: I have updated the code with your suggestion, thanks!

PR #854, opened 2019-06-27
Added a check for a failed process job message.

bbergquist0930:
This change adds a check that if the prvProcessOTAJobMsg returns NULL
and the image state is still unknown, the OTA job is aborted with
a job parse error

Description
-----------
If "prvProcessOTAJobMsg" returns NULL, a test is made to see if the image state is in the "self_test" phase.  However "prvProcessOTAJobMsg" will also return NULL if there is an error in parsing the OTA job data.  

This changes checks for that condition by checking if the image state is still in the "unknown" state and if so, aborts the OTA job with a parse error.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -1481,10 +1481,12 @@ static void prvOTAUpdateTask( void * pvUnused )
                                             ( void ) prvResetDevice(); /* Ignore return code since there's nothing we can do if we can't force reset. */
                                         }
 									}
-                                    else if (OTA_GetImageState() == eOTA_ImageState_Unknown)
+                                    else
                                     {
-										/* If the job context returned NULL and the image state is unknown, then an error
-										 * occurred parsing the OTA job message.  Abort the OTA job with a parse error.
+										/* If the job context returned NULL and the image state is not in the self_test state,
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1480,7 +1480,14 @@ static void prvOTAUpdateTask( void * pvUnused )
                                             ( void ) prvSetImageStateWithReason( eOTA_ImageState_Rejected, kOTA_Err_ImageStateMismatch );
                                             ( void ) prvResetDevice(); /* Ignore return code since there's nothing we can do if we can't force reset. */
                                         }
-                                    }
+									}
+                                    else if (OTA_GetImageState() == eOTA_ImageState_Unknown)
```
pvyawaha: Instead of calling prvSetImageStateWithReason after checking for eOTA_ImageState_Unknown we can call it for any other state than eOTA_ImageState_Testing. For example previous abort set the image state to eOTA_ImageState_Aborted and after that we get a job document that fails parsing.

bbergquist0930: Updated the pull request to not check for this specific reason, but rather anything else other than "eOTA_ImageState_Testing" will be handled.

PR #861, opened 2019-06-28
aws_tcp_echo_client_single_task.c : a time-out is not fatal (second version)

htibosch:
aws_tcp_echo_client_single_task.c : a time-out is not fatal (second version)

Description
-----------

The earlier PR #354 had too many conflicts, this one will replace it.

When the reception of data times out, and at least 1 byte has been received, an ASSERT will be thrown because the contents does not match.
When the number of bytes received is less than expected, it would be nicer to just consider this as a failure. The link may have failed, the remote party is too slow, etc.

Only when the expected number of bytes was received, but the received string does not match, this should be a fatal error ( calling configASSERT() ).

Also I replaced a few tabs with spaces, to make it confirm Crustify.cfg.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/tcp/aws_tcp_echo_client_single_task.c
@@ -198,15 +198,19 @@ static void prvEchoClientTask( void * pvParameters )
     char * pcReceivedString;
     TickType_t xTimeOnEntering;
 
-#if ( ipconfigUSE_TCP_WIN == 1 )
-    WinProperties_t xWinProps;
+    #if ( ipconfigUSE_TCP_WIN == 1 )
+        WinProperties_t xWinProps;
+    #endif
 
-    /* Fill in the buffer and window sizes that will be used by the socket. */
-    xWinProps.lTxBufSize = ipconfigTCP_TX_BUFFER_LENGTH;
-    xWinProps.lTxWinSize = configECHO_CLIENT_TX_WINDOW_SIZE;
-    xWinProps.lRxBufSize = ipconfigTCP_RX_BUFFER_LENGTH;
-    xWinProps.lRxWinSize = configECHO_CLIENT_RX_WINDOW_SIZE;
-#endif /* ipconfigUSE_TCP_WIN */
+    #if ( ipconfigUSE_TCP_WIN == 1 )
+        {
+            /* Fill in the buffer and window sizes that will be used by the socket. */
+            xWinProps.lTxBufSize = ipconfigTCP_TX_BUFFER_LENGTH;
+            xWinProps.lTxWinSize = configECHO_CLIENT_TX_WINDOW_SIZE;
+            xWinProps.lRxBufSize = ipconfigTCP_RX_BUFFER_LENGTH;
+            xWinProps.lRxWinSize = configECHO_CLIENT_RX_WINDOW_SIZE;
+        }
+    #endif /* ipconfigUSE_TCP_WIN */
 
     /* This task can be created a number of times.  Each instance is numbered
      * to enable each instance to use a different Rx and Tx buffer.  The number is
--- a/demos/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/tcp/aws_tcp_echo_client_single_task.c
@@ -343,9 +343,9 @@ static void prvEchoClientTask( void * pvParameters )
                         break;
                     }
                 }
-                else if( xReceivedBytes < 0 )
+                else if( xReceivedBytes == -pdFREERTOS_ERRNO_ENOTCONN )
                 {
-                    /* SOCKETS_Recv() returned an error. */
+                    /* The connection got closed. */
                     break;
                 }
                 else
--- a/demos/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/tcp/aws_tcp_echo_client_single_task.c
@@ -198,17 +198,15 @@ static void prvEchoClientTask( void * pvParameters )
     char * pcReceivedString;
     TickType_t xTimeOnEntering;
 
-    #if ( ipconfigUSE_TCP_WIN == 1 )
-        {
-            WinProperties_t xWinProps;
+#if ( ipconfigUSE_TCP_WIN == 1 )
+    WinProperties_t xWinProps;
 
-            /* Fill in the buffer and window sizes that will be used by the socket. */
-            xWinProps.lTxBufSize = ipconfigTCP_TX_BUFFER_LENGTH;
-            xWinProps.lTxWinSize = configECHO_CLIENT_TX_WINDOW_SIZE;
-            xWinProps.lRxBufSize = ipconfigTCP_RX_BUFFER_LENGTH;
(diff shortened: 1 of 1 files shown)


htibosch:
> Code style check

I just ran Uncrustify with [uncrustify.cfg](https://github.com/aws/amazon-freertos/blob/master/tools/uncrustify.cfg), which changed a few lines

demos/tcp/aws_tcp_echo_client_single_task.c
```diff
@@ -350,6 +350,11 @@ static void prvEchoClientTask( void * pvParameters )
                 }
                 else
                 {
+                    /* The received length did not match the transmitted
+                     * length. */
+                    ulTxRxFailures[ xInstance ]++;
```
alexa-noxon: It seems like the handling for this condition (xTransmitted>xReceivedBytes>=0) might be swapped with the case above (xReceivedBytes < 0)

I think we should trigger an error condition on xReceivedBytes being negative because that indicates an error occurred.

I'm not as clear on whether  (xTransmitted>xReceivedBytes>=0) should simply break and try again, or if we consider not receiving all the bytes within the timeout an error and we should increment the failure counter.

htibosch:
I'm not sure how to answer your question, but let me rephrase it:

    if( xReceivedBytes == xTransmitted )
    {
        /* The full packet is received, it makes sense to do a memcmp() */
    }
    else if( xReceivedBytes < 0 )
    {
        /* In most cases, a negative return value means: disconnected,
         * or -pdFREERTOS_ERRNO_ENOTCONN.
         * Another possible error is -pdFREERTOS_ERRNO_ENOMEM when malloc failed. */
    }
    else
    {
        /* xTransmitted > xReceivedBytes >= 0 */
    }

In the existing version, a failing `memcmp()` would lead to an assert. I thought it is reasonable to call `memcmp()` only when enough bytes have been received.

I propose to do the following:

    if( xReceivedBytes == xTransmitted )
    else if( xReceivedBytes == -pdFREERTOS_ERRNO_ENOTCONN )
    else

All other values of `xReceivedBytes` will be logged, including unexpected error codes like `-ENOMEM`.

qiutongs:
I believe this fix is good. But still want Dan to take a look.

demos/tcp/aws_tcp_echo_client_single_task.c
dan4thewin: I'd also like to see negative values treated separately and brought to the user's attention.  (Do we have a perror(3) equivalent?)

demos/tcp/aws_tcp_echo_client_single_task.c
```diff
@@ -323,7 +323,7 @@ static void prvEchoClientTask( void * pvParameters )
                 /* If an error occurred it will be latched in xReceivedBytes,
```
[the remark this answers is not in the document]
htibosch: Apparently, the test and demo projects do not use the TCP sliding windows ( `ipconfigUSE_TCP_WIN` ). Which is a pity, because it would make them run considerably faster.

I will repair it, and re-test it in real hardware, both with TCP sliding enabled and disabled.
( see my commit: "xWinProps undefined when ipconfigUSE_TCP_WIN equals 1" )

demos/tcp/aws_tcp_echo_client_single_task.c
htibosch: The error numbers returned by the API's are quite compatible with errno.h
I sometimes use `strerror_r()` to print them:

    if( xReceivedBytes < 0 )
    {
        char pcBuffer[ 128 ];
        strerror_r( -xReceivedBytes, pcBuffer, sizeof( pcBuffer ) );
        configPRINTF( ( "ERROR: xTransmitted %d errno %d (%s)\r\n",
                         int ) xTransmitted, ( int ) -xReceivedBytes, pcBuffer ) );
    }
    else
    {
        configPRINTF( ( "ERROR: xTransmitted %d xReceivedBytes %d \r\n",
                         int ) xTransmitted, ( int ) xReceivedBytes ) );
    }
    break;

But I'm afraid that `strerror_r()` is too compiler-specific?

demos/tcp/aws_tcp_echo_client_single_task.c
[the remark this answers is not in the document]
htibosch: I re-tested the code, both with and without sliding windows. I had an echo server that can trigger several errors, like:
- Corrupted data
- Bad length ( less than expected )
- Sudden disconnection

That all worked well. Corrupted data leads to an ASSERT.

demos/tcp/aws_tcp_echo_client_single_task.c
dan4thewin: In the mqtt demo I see IotMqtt_strerror() where someone rolled their own.  We could do something like that with a small subset of meaningful errors.

htibosch: > In the mqtt demo I see IotMqtt_strerror()

FreeRTOS+TCP uses 15 different errno values, so it is doable to make a `strerror()` function.
```c
EADDRINUSE          ENOMEM
EADDRNOTAVAIL       ENOPROTOOPT
EBADF               ENOSPC
ECANCELED           ENOTCONN
EINPROGRESS         EOPNOTSUPP
EINTR               ETIMEDOUT
EINVAL              EWOULDBLOCK /* same as EAGAIN */
EISCONN

```
I don't mind making a separate PR to introduce that function.

dan4thewin:
@htibosch , looks like a style check is failing.  Can you run this through uncrustify?

htibosch:
> @htibosch , looks like a style check is failing. Can you run this through uncrustify?

Sure. Now it will pass the uncrustify test.

The current version in [master](https://github.com/aws/amazon-freertos/blob/master/demos/tcp/aws_tcp_echo_client_single_task.c) has this:

~~~c
    #if ( ipconfigUSE_TCP_WIN == 1 )
        {
            WinProperties_t xWinProps;

            /* Fill in the buffer and window sizes that will be used by the socket. */
            xWinProps.lTxBufSize = ipconfigTCP_TX_BUFFER_LENGTH;
            xWinProps.lTxWinSize = configECHO_CLIENT_TX_WINDOW_SIZE;
            xWinProps.lRxBufSize = ipconfigTCP_RX_BUFFER_LENGTH;
            xWinProps.lRxWinSize = configECHO_CLIENT_RX_WINDOW_SIZE;
        }
    #endif /* ipconfigUSE_TCP_WIN */
~~~

This code can not be compiled when `ipconfigUSE_TCP_WIN` is enabled. The compiler complains here that `xWinProps` is not defined:

~~~c
    #if ( ipconfigUSE_TCP_WIN == 1 )
        {
            /* Set the window and buffer sizes. */
            SOCKETS_SetSockOpt( xSocket, 0, FREERTOS_SO_WIN_PROPERTIES, ( void * ) &xWinProps, sizeof( xWinProps ) );
        }
    #endif /* ipconfigUSE_TCP_WIN */
~~~

PR #871, opened 2019-07-01
Length check in usGenerateProtocolChecksum()

htibosch:
Corrected length-check in usGenerateProtocolChecksum()
-----------

The 16-bit IPv4-field `usLength` is equal to the length of the IP-header, plus the protocol header plus the payload.
But it does not have knowledge about the Ethernet header that may precede it.

The the length check should be:

    -	if( uxBufferLength < FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength )
    +	if( uxBufferLength < FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength + ipSIZE_OF_ETH_HEADER )

I tested the above change by adding some logging, which shows that `uxBufferLength` is always 14 bytes bigger than what is indicated in `usLength`.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
@@ -1824,7 +1824,7 @@ uint8_t ucProtocol;
 	{
 		return ipINVALID_LENGTH;
 	}
-	if( uxBufferLength < FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength ) )
+	if( uxBufferLength < ( size_t ) ( ipSIZE_OF_ETH_HEADER + FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength ) ) )
 	{
 		return ipINVALID_LENGTH;
 	}


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1824,7 +1824,7 @@ uint8_t ucProtocol;
 	{
 		return ipINVALID_LENGTH;
 	}
-	if( uxBufferLength < FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength ) )
+	if( uxBufferLength < ( size_t ) ( ipSIZE_OF_ETH_HEADER + FreeRTOS_ntohs( pxIPPacket->xIPHeader.usLength ) ) )
```
dan4thewin: Is the layer 2 header copied to the buffer?  Is it always an ethernet packet - could it be an 802.11 packet?

htibosch: The function `usGenerateProtocolChecksum()` expects a packet that starts with a 14-byte Ethernet header.

PR #872, opened 2019-07-01
DNS calculates protocol checksum with wrong length

htibosch:
DNS calculates protocol-checksum with wrong length
-----------

When the IP-task must calculate checksums:

    #define ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM   0

the function `prvReplyDNSMessage()` passes the length of the UDP-payload, in stead of the length of the entire Ethernet packet to `usGenerateProtocolChecksum()`.

    -	usGenerateProtocolChecksum( ( uint8_t* ) pxUDPPacket, lNetLength, pdTRUE );
    +	usGenerateProtocolChecksum( ( uint8_t* ) pxUDPPacket, pxNetworkBuffer->xDataLength, pdTRUE );

As a result, the unit was not able to reply to LLMNR or NBNS requests.

I tested it by doing a ping to my device, by name. Ping uses the LLMNR protocol to find the IP-address.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -1308,6 +1308,9 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 		pxUDPHeader->usLength			   = FreeRTOS_htons( lNetLength + ipSIZE_OF_UDP_HEADER );
 		vFlip_16( pxUDPPacket->xUDPHeader.usSourcePort, pxUDPPacket->xUDPHeader.usDestinationPort );
 
+		/* Important: tell NIC driver how many bytes must be sent */
+		pxNetworkBuffer->xDataLength = ( size_t ) ( lNetLength + ipSIZE_OF_IPv4_HEADER + ipSIZE_OF_UDP_HEADER + ipSIZE_OF_ETH_HEADER );
+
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1308,6 +1308,9 @@ TickType_t xTimeoutTime = pdMS_TO_TICKS( 200 );
 		pxUDPHeader->usLength			   = FreeRTOS_htons( lNetLength + ipSIZE_OF_UDP_HEADER );
 		vFlip_16( pxUDPPacket->xUDPHeader.usSourcePort, pxUDPPacket->xUDPHeader.usDestinationPort );
 
+		/* Important: tell NIC driver how many bytes must be sent */
+		pxNetworkBuffer->xDataLength = ( size_t ) ( lNetLength + ipSIZE_OF_IPv4_HEADER + ipSIZE_OF_UDP_HEADER + ipSIZE_OF_ETH_HEADER );
```
dan4thewin: As in the PR for usGenerateProtocolChecksum, I have to ask; is this function only used for ethernet frames, and never with any other layer 2 type?

htibosch: It is used for ethernet frames only, just like all other FreeRTOS+TCP functions.

PR #881, opened 2019-07-01
Add CBMC harness for QueueCreateMutex

mmuesly:
Description
-----------

Add a CBMC harness for QueueCreateMutex. The single input variable is completely unrestricted.

Depends on #774

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/cbmc/proofs/Queue/QueueCreateMutex/QueueCreateMutex_harness.c
+++ b/tools/cbmc/proofs/Queue/QueueCreateMutex/QueueCreateMutex_harness.c
@@ -1,0 +1,10 @@
+#include "FreeRTOS.h"
+#include "queue.h"
+
+#include "cbmc.h"
+
+void harness() {
+  uint8_t ucQueueType;
+
+  xQueueCreateMutex(ucQueueType);
+}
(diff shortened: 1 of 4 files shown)


tools/cbmc/proofs/Queue/QueueCreateMutex/QueueCreateMutex_harness.c
```diff
@@ -0,0 +1,10 @@
+#include "FreeRTOS.h"
+#include "queue.h"
+
+#include "cbmc.h"
+
+void harness() {
+  uint8_t ucQueueType;
+
+  xQueueCreateMutex(ucQueueType);
```
danielsn: You get back a ```QueueHandle_t``` here.  Surely there are some properties to assert on it?

mmuesly: It is a pointer to a queue struct. I am not sure which properties you want to assert. The only thing I could imagine is to test that it is a valid pointer. As all relevant fields of the Mutex have been written during the xQueueCreateMutex function calls, there must be allocated memory for them. Hence, I am not sure whether this is a meaningful assertion.

danielsn: The fact that they were written is an implementation detail - its better to check that as a contract. What does a valid queue struct look like?

markrtuttle:
Closed in favor of https://github.com/aws/amazon-freertos/pull/1324.

Issue #897, opened 2019-07-03
OTA gets stuck if the "prvPublishGetStreamMessage" receives an error from "prvPublishMessage"


bbergquist0930:
**Describe the bug**
If the "prvPublishMessage" called within "prvPublishGetStreamMessage" returns an error, the OTA process becomes stuck.  

The “prvPublishGetStreamMessage” is called only on a timer event “prvOTAUpdateTask” which is set by “prvRequestTimer_Callback” which is the itself setup by “prvStartRequestTimer” which is called by both “prvProcessOTAJobMessage” and “prvInjectDataBlock”.

So if there is no timer setup, then “prvPublishGetStreamMessage” will never be called. When itself receives an error from the “prvPublishMessage”, then no timer is being restarted and it will never be called again.

The comment in the code indicates that the author wanted to just let the “momentum” mechanism to handle this issue and not return an error directly in case the publish error was intermittent, but the “momentum” mechanism is only incremented in this “prvPublishGetStreamMessage” function, so without it being called again this will never be updated.

The solution is to just start the timer again in this case. If the problem is intermittent, then the “prvPublishMessage” will not fail again. If it is persistent, then the “prvPublishGetStreamMessage” will be called exactly as if the there was a timeout on reception of the data blocks and after 32 retries, the OTA will be aborted.

**System information**
This issue was found on an earlier version of the software, however, looking at the latest code as of today 6/3/2019, the issue still exists.

**Expected behavior**
The OTA process should not become stuck and if the "prvPublishMessage" were to continue to fail and return an error, the OTA job would be aborted.

**Screenshots or console output**

**To reproduce**
So how cause the "prvPublishMessage" call to return an error.  What I did is to simulate this by having after 10 calls, to not call the "prvPublishMessage", but rather set "eResult" to something other than "eMQTTAgentSuccess".  This will show the symptoms and the OTA will become stuck.
![Screen Shot 2019-07-02 at 8 19 30 AM](https://user-images.githubusercontent.com/28760864/60598861-74586880-9d7b-11e9-9d7f-3faa1aa38eab.png)

Thank you!

sarenameas:
Hello,
Thanks for bringing this up. This intermittent issue we have seen and are actively working on internally. We will get back to you in a week or so after we recreate it internally and try your fix with a few days of soak as well.

pvyawaha:
Hi Brett, 

[...]

PR #898, opened 2019-07-03
Fix for issue#897.

bbergquist0930:
Added starting of the request timer even when the prvPublishMmessage returns a failure.

This is a fix for [Issue#897](https://github.com/aws/amazon-freertos/issues/897)

Description
-----------
Added the starting of the request timer even if the "prvPublishMessage" returns a failure.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -1257,17 +1257,25 @@ static OTA_Err_t prvPublishGetStreamMessage( OTA_FileContext_t * C )
 
                     if( eResult != IOT_MQTT_SUCCESS )
                     {
+                        /* Don't return an error. Let max momentum catch it since this may be intermittent. */
                         OTA_LOG_L1( "[%s] Failed: %s\r\n", OTA_METHOD_NAME, pcTopicBuffer );
-                        /* Don't return an error. Let max momentum catch it since this may be intermittent.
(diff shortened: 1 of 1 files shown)


sarenameas:
Hello,
Thanks for bringing this up. This intermittent issue we have seen and are actively working on internally. We will get back to you in a week or so after we recreate it internally and try your fix with a few days of soak as well.

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1258,7 +1258,9 @@ static OTA_Err_t prvPublishGetStreamMessage( OTA_FileContext_t * C )
                     if( eResult != IOT_MQTT_SUCCESS )
                     {
                         OTA_LOG_L1( "[%s] Failed: %s\r\n", OTA_METHOD_NAME, pcTopicBuffer );
-                        /* Don't return an error. Let max momentum catch it since this may be intermittent. */
+                        /* Don't return an error. Let max momentum catch it since this may be intermittent.
```
pvyawaha: Can we please move the call to prvStartRequestTimer after if-else block as it is being called in both cases and the if-else blocks will be left for logs only.

bbergquist0930: Updated the change to only restart the timer after the if-else block and added a comment on the logic of why it is always restarted.

PR #901, opened 2019-07-03
Change references from v4 to v2

alexa-noxon:
The C-SDK refers to upcoming MQTT API as v4, but for Amazon FreeRTOS they are v2 of these libraries.
Fixing comment references between v4 and v2 in Amazon FreeRTOS-only files to use Amazon FreeRTOS
versioning syntax.

https://amazon-freertos-ci.corp.amazon.com/view/3-Nightly/job/dsl_nightly_cmake/71/

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/aws/shadow/src/aws_shadow.c
+++ b/libraries/c_sdk/aws/shadow/src/aws_shadow.c
@@ -107,7 +107,7 @@ static void prvUpdatedCallbackWrapper( void * pvArgument,
                                        AwsIotShadowCallbackParam_t * const pxUpdatedDocument );
 
 /* Retrieves the MQTT v2 connection from the MQTT v1 connection handle. */
-extern IotMqttConnection_t MQTT_AGENT_Getv4Connection( MQTTAgentHandle_t xMQTTHandle );
+extern IotMqttConnection_t MQTT_AGENT_Getv2Connection( MQTTAgentHandle_t xMQTTHandle );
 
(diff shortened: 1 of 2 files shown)


libraries/c_sdk/aws/shadow/src/aws_shadow.c
```diff
@@ -80,33 +80,33 @@ typedef struct ShadowClient
...
  * @param[in] pvArgument The associated #ShadowClient_t.
- * @param[in] pxUpdatedDocument Shadow v4 updated document info.
+ * @param[in] pxUpdatedDocument Shadow v2 updated document info.
  */
 static void prvUpdatedCallbackWrapper( void * pvArgument,
                                        AwsIotShadowCallbackParam_t * const pxUpdatedDocument );
 
-/* Retrieves the MQTT v4 connection from the MQTT v1 connection handle. */
+/* Retrieves the MQTT v2 connection from the MQTT v1 connection handle. */
 extern IotMqttConnection_t MQTT_AGENT_Getv4Connection( MQTTAgentHandle_t xMQTTHandle );
```
alexa-noxon: Note: I didn't change the function name here

gordonwang0: This can be changed as it's a function internal to the conversion layer. (The only reason the function isn't static is because it is used in both the MQTT and Shadow conversion layers.)

PR #915, opened 2019-07-09
CBMC memory-safety proof for DNSgetHostByName

dlohar:
Description
-----------
This PR includes the CBMC proof for memory safety of the DNSgetHostByName function, which prepares and sends a message to a DNS server.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -420,7 +420,6 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
-if(pcHostName) {
 
 
 	/* If the supplied hostname is IP address, convert it to uint32_t
@@ -482,8 +481,9 @@ if(pcHostName) {
 	{
 		ulIPAddress = prvGetHostByName( pcHostName, xIdentifier, xReadTimeOut_ms );
 	}
+
 	return ulIPAddress;
-  }
+
 }
 /*-----------------------------------------------------------*/
 
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -420,7 +420,7 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
-configASSERT(pcHostName);
+if(pcHostName) {
 
 
 	/* If the supplied hostname is IP address, convert it to uint32_t
@@ -482,8 +482,8 @@ configASSERT(pcHostName);
 	{
 		ulIPAddress = prvGetHostByName( pcHostName, xIdentifier, xReadTimeOut_ms );
 	}
-
 	return ulIPAddress;
+  }
 }
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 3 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -420,6 +420,8 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
+configASSERT(pcHostName);
```
dan4thewin: same comment as for DNSgetHostByName_a

tools/cbmc/proofs/DNS/DNSgetHostByName/DNSgetHostByName_harness.c
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/* Abstraction of FreeRTOS_socket. This abstraction allocates a memory of size Socket_t. */
+Socket_t FreeRTOS_socket( BaseType_t xDomain, BaseType_t xType, BaseType_t xProtocol ) {
+	Socket_t xSocket = safeMalloc(sizeof(Socket_t)); /* Replaced malloc by safeMalloc */
+	return xSocket;
+}
+
+/* This function only uses the return value of prvParseDNSReply. Hence it returns an unconstrained uint32 value */
+uint32_t prvParseDNSReply( uint8_t *pucUDPPayloadBuffer, size_t xBufferLength, TickType_t xIdentifier) { }
```
dan4thewin: There seems to be a lot of duplication among the DNS harness files.  For the sake of future maintainers, can we strive for DRY, and refactor so this code is shared?

tools/cbmc/proofs/DNS/DNSgetHostByName/DNSgetHostByName_harness.c
```diff
@@ -0,0 +1,47 @@
...
+
+void harness() {
+	size_t len;
+	__CPROVER_assume(len >= 0 && len <= MAX_HOSTNAME_LEN);
+	char *pcHostName = safeMalloc(len); /* Replaced malloc by safeMalloc */
+	if (len && pcHostName) {
+		pcHostName[len-1] = NULL;
```
dan4thewin: I've seen this in a couple of these now.  Is this accurately expressing an invariant of the original source, or is this introducing an assumption?  Seems to me it would be desirable to catch the case when the sentinel character is not present.

dlohar: This is required as there are `strlen`, `strcpy` functions in the code which require `NULL` termination. In this harness we are assuming any `NULL` terminated  `hostname` of length up to `MAX_HOSTNAME_LEN`. If we want to catch the case where the `hostname` is not `NULL` terminated, we need to assert it in `FreeRTOS_DNS.c`.

tools/cbmc/proofs/DNS/DNSgetHostByName/DNSgetHostByName_harness.c
karkhaz: @dan4thewin May I request that I do the deduplication work myself, later in a separate PR?

There has certainly been a lot of duplicate stubs in the proofs that we've been adding this summer. I was planning to re-review them myself once everything has settled down a few weeks from now, and decide on the most sensible way to unify the stubs.

In AWS C-Common (another project that we've been contributing proofs to), there is a [proof helpers directory](https://github.com/awslabs/aws-c-common/tree/master/.cbmc-batch/include/proof_helpers) containing code that is common to all proofs. It's certainly a good idea to have a similar thing in a:FR, but I'd like to get an overview of the commonality in the current and incoming proofs first.

Please let me know if that's okay with you.

dan4thewin: That'd be great.  Thanks.

tools/cbmc/proofs/DNS/DNSgetHostByName/DNSgetHostByName_harness.c
dan4thewin: I see.  So, the proof is only valid when this precondition is met?

dlohar: Yes.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -420,6 +420,8 @@ uint32_t FreeRTOS_gethostbyname_a( const char *pcHostName, FOnDNSEvent pCallback
 uint32_t ulIPAddress = 0UL;
 TickType_t xReadTimeOut_ms = ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME;
 TickType_t xIdentifier = 0;
+if(pcHostName) {
```
dan4thewin: Is this the exact same change from #663?  Shouldn't only one of these PR's contain this change?

dlohar: Yes, I removed the `if` and `configASSERT`.

PR #916, opened 2019-07-09
Feat:Add support in OTA agent for controlling number of data blocks.

pvyawaha:
Add support in OTA agent for controlling number of data blocks requested from streaming service.

Description
-----------

This change adds support in OTA agent for controlling number of data blocks requested from streaming service. A new parameter representing number of data blocks is added as a config that is sent with every data block request. For example if this parameter is set as 16 then OTA streaming service will send 16 Data Blocks in response. 

This change also restarts the timer as soon as the requested number of data blocks are received instead of waiting for the request timer timeout. This helps in speeding up the OTA process by avoiding unnecessary wait for timer to expire.

Checklist:
----------

- [x] My code is Linted.


sarenameas:
Looks like you will need to uncrustify those files.

PR #918, opened 2019-07-09
Update JSON deserializer to parse nested objects

danhorner:
Description
-----------
Update JSON deserializer to parse nested objects. 
 - Add unit tests for deserializer. 
 - update parser to make `find_nested_key_array_of_objects_value` pass

Checklist:
----------

- [X] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/serializer/CMakeLists.txt
+++ b/libraries/c_sdk/standard/serializer/CMakeLists.txt
@@ -38,6 +38,7 @@ afr_module_sources(
     INTERFACE
         "${test_dir}/iot_tests_serializer_cbor.c"
         "${test_dir}/iot_tests_serializer_json.c"
+	"${test_dir}/iot_tests_deserializer_json.c"
 )
 afr_module_dependencies(
     ${AFR_CURRENT_MODULE}
--- a/tests/common/aws_test_runner.c
+++ b/tests/common/aws_test_runner.c
@@ -71,7 +71,7 @@ static void RunTests( void )
     #endif
 
     #if ( testrunnerFULL_TASKPOOL_ENABLED == 1 )
-            RUN_TEST_GROUP( Common_Unit_Task_Pool );
+        RUN_TEST_GROUP( Common_Unit_Task_Pool );
     #endif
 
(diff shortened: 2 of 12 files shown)


tests/common/aws_test_runner.c
```diff
@@ -188,6 +188,7 @@ static void RunTests( void )
     #if ( testrunnerFULL_SERIALIZER_ENABLED == 1 )
         RUN_TEST_GROUP( Full_Serializer_CBOR );
         RUN_TEST_GROUP( Full_Serializer_JSON );
+        RUN_TEST_GROUP( Deserializer_JSON );
```
qiutongs: Our test infra needs to be updated correspondingly. I am not sure where "Serializer Tests" are defined..

libraries/c_sdk/standard/serializer/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ afr_module_sources(
     INTERFACE
         "${test_dir}/iot_tests_serializer_cbor.c"
         "${test_dir}/iot_tests_serializer_json.c"
+	"${test_dir}/iot_tests_deserializer_json.c"
```
qiutongs: Whenever we add a new file, we are supposed to update all project files (test project in your case). You don't need to make the change in this PR. I just raise this concern.

tests/common/aws_test_runner.c
ravibhagavandas: Test rules are defined here : https://github.com/amazon-freertos/afr-host-tools-config/blob/40e545ace0c2885330235b49c4a806415359d002/test_rules.yml

I think its better to prefix with Full_Serializer so that it runs with all other tests.

danhorner: @ravibhagavandas, I updated the PR to put the new tests behind the Full_Serializer prefix.

libraries/c_sdk/standard/serializer/CMakeLists.txt
danhorner: @qiutongs I updated the PR with the revised project files.

qiutongs: Looks good

PR #919, opened 2019-07-09
BLE - Nimble and blob create service

huguesBouvier:
BLE - Nimble and blob create service

Description
-----------
Adding Nimble drivers for esp - deactivated for now
Adding blob create service support. 
     Mode supported on Nordic only
Adding versionning for BLE
Adding lengh parameter for short local name on ble hal

Nordic now support both blob option and separate attribute creation. 

Tested:
MQTT demos and ESP and Nordic
BLE tests on ESP and Nordic

Checklist:
----------

- [ ] My code is Linted.


libraries/c_sdk/standard/ble/src/iot_ble_gap.c
```diff
@@ -72,7 +72,7 @@ static const BTUuid_t _serverUUID =
 static IotBleAdvertisementParams_t _scanRespParams =
 {
     .includeTxPower    = true,
-    .nameType          = BTGattAdvNameNone,
+    .name              = { BTGattAdvNameNone,                  0},
```
ravibhagavandas: Indentation seems to be off due to uncrustify. Maybe break into multiple lines ?

huguesBouvier: fixed

PR #941, opened 2019-07-15
Add ota over ble test case

bxpan:
Add ota over ble test case

Description
-----------
Update the ota test script to add ota e2e test over ble.

Checklist:
----------

- [x] My code is Linted.


tools/ota_e2e_tests/aws_ota_test/aws_ota_project.py
```diff
@@ -302,6 +304,55 @@ def setOtaBlockSize(self, blockSize):
             {' #define otaconfigMAX_NUM_BLOCKS_REQUEST': str(blockSize) + 'U'},
             os.path.join(self._projectRootDir, self._boardProjectPath, 'config_files', 'aws_ota_agent_config.h')
         )
+ 
+    def __insertTexts(self, prefix, texts, filePath):
```
tgsong: double leading underscore means reserved by system in python, use single leading underscore for private method.

HarnishaA: Originally the whole scripts was written in this way with double leading underscore. Just trying to keep consistency now :-)

tgsong: I see, let's correct the wrong consistency with a separate PR.

PR #947, opened 2019-07-16
Espressif: Add NimBLE Host Stack in ESP-IDF

dhrishi:
Espressif: Add NimBLE Host Stack in ESP-IDF

Description
-----------
1. Update the upstream host stack to 1.1.0 release
2. Miscellaneous bug fixes
3. Feature to disable and deinit NimBLE host as well as FreeRTOS port

Enabled NimBLE in both aws_demos and aws_tests

Checklist:
----------

- [ ] I have tested my changes. No regression in existing tests. - Tested compilation using cmake as well as make.
- [ ] My code is Linted.


huguesBouvier:
Thank you for the PR! 

There is only 3 potential findings:
1.	It seems there is a regression in remove bonding, I can’t remove the bond anymore:
See the following tests failure:
TEST(Full_BLE, BLE_Connection_RemoveBonding)/home/ANT.AMAZON.COM/hbouvier/Desktop/Nimble/amazon-freertos/libraries/abstractions/ble_hal/test/iot_test_ble_hal.c:829::FAIL: Expected 0 Was 1 GAP procedure initiated: advertise; disc_mode=2 adv_channel_map=0 own_addr_type=0 adv_filter_policy=0 adv_itvl_min=51 adv_itvl_max=102
TEST(Full_BLE, BLE_Connection_RemoveAllBonds)/home/ANT.AMAZON.COM/hbouvier/Desktop/Nimble/amazon-freertos/libraries/abstractions/ble_hal/test/iot_test_ble_hal.c:829::FAIL: Expected 0 Was 1

2.	It is able to pair in Just works when secure connection only is on.
reconnectWhileNotBonded test: Able to pair in just Works mode TEST(Full_BLE, RaspberryPI_reconnectWhileNotBonded) FAIL
I suspect that issue is because bonding was not removed. In that case it would still pair (because all keys are already accepted and in memory)

3.	I was not was to revert back to bluedroid with cmake. I was able to achieve that only with the makefile. If it is possible, could you tell me how to do it?

PR #968, opened 2019-07-23
Add ltorg to port for gcc arm cm3+4f+7

josesimoes:
Description
-----------
- Add LTORG to `prvPortStartFirstTask` and `vPortEnableVFP` to enable compile with GCC and LTO optimization.

This was spotted when migrating the CI toolchain of [nanoFramework NXP port](https://github.com/nanoframework/nf-interpreter/pull/1411) to compile with GCC 8 2019 Q3 update. The optimization includes `-Os -flto -fuse-linker-plugin -fno-fat-lto-objects`. 

The build failed with:
```text
(…)
ccw7TTqT.s:154: Error: offset out of range
(…)
arm-none-eabi/bin/ld.exe: error: lto-wrapper failed
``` 
Removing LTO optimization options above allow the build to pass. Not the best option.

The problem seems to be related to the ldr generating a too long jump to the literal pool. A simple solution is to add a .ltorg at the end of each of `prvPortStartFirstTask` and `vPortEnableVFP` functions to ensure that the pool gets placed there.

See [here](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0041c/Chedgddh.html) about LTORG directive. 
And [here](https://bugs.launchpad.net/gcc-arm-embedded/+bug/1763050) about the proposed fix.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/freertos_kernel/portable/GCC/ARM_CM3/port.c
+++ b/freertos_kernel/portable/GCC/ARM_CM3/port.c
@@ -247,6 +247,7 @@ static void prvPortStartFirstTask( void )
 					" isb					\n"
 					" svc 0					\n" /* System call to start first task. */
 					" nop					\n"
+					" .ltorg 				\n" /* make sure the pool is placed here, so ldr doesn't generate a too long jump */
 				);
 }
 /*-----------------------------------------------------------*/
--- a/freertos_kernel/portable/GCC/ARM_CM4F/port.c
+++ b/freertos_kernel/portable/GCC/ARM_CM4F/port.c
@@ -277,6 +277,7 @@ static void prvPortStartFirstTask( void )
 					" isb					\n"
 					" svc 0					\n" /* System call to start first task. */
 					" nop					\n"
+					" .ltorg 				\n" /* make sure the pool is placed here, so ldr doesn't generate a too long jump */
 				);
 }
 /*-----------------------------------------------------------*/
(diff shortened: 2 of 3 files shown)


RichardBarry:
Thanks for this - although we don't use LTO in our tests yet we have added a few updates recently that should make it possible and this seems to fit that category.  We will check with several compiler versions, and if all ok, there will be several other GCC port that can have the same update.

josesimoes:
@RichardBarry got you!

FYI: this build error has just showed up after updating to the new version.
The LTO optimization options have been there for a couple of years now and have been working fine with the previous GCC versions.

networkfusion:
@RichardBarry is there any update on this?

cobusve:
This change affects only the FreeRTOS Kernel.

The Kernel is now a submodule of the Amazon-FreeRTOS repo so we are not able to merge this change in from this location. I have confirmed that we are in the process of updating this in the FreeRTOS.org source repo, which will then come to Amazon-FreeRTOS downstream.

yuhui-zheng:
@josesimoes I will open another PR in https://github.com/FreeRTOS/FreeRTOS-Kernel repo, and validate it there. Apologies for the unreasonable delay.

networkfusion:
@yuhui-zheng any reason why it was never created in the kernel repo / never linked here?

networkfusion:
@RichardBarry @yuhui-zheng an update would still be great?! thanks.

yuhui-zheng:
https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/71 A colleague of mine has reopened the PR. 

Apologies... (<-- I seem to do that a lot recently...) This item has been in our backlog for a while. 

Minor note... when an issue/PR is closed, it's better to get attention with another issue/PR since that's better monitored. (And I agree, theoretically you don't have to if we do things fast enough... )

PR #996, opened 2019-07-30
Anoxon/remove dup defines

alexa-noxon:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
+++ b/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
@@ -813,7 +813,7 @@ CK_RV xInitializePkcs11Token()
 
     if( ( CKR_OK == xResult ) && !( CKF_TOKEN_INITIALIZED & xTokenFlags ) )
     {
-        /* If token is not initialized - initialize it */
+        /* Initialize the token if it is not already. */
         xResult = pxFunctionList->C_InitToken( pxSlotId[ 0 ],
                                                configPKCS11_DEFAULT_USER_PIN,
                                                sizeof( configPKCS11_DEFAULT_USER_PIN ) - 1,
--- a/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
+++ b/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
@@ -764,6 +764,67 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
+CK_RV xInitializePkcs11Token()
+{
+    CK_RV xResult;
+
+    CK_FUNCTION_LIST_PTR pxFunctionList;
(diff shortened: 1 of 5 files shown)


demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -764,6 +764,67 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
...
+
+    if( ( xResult == CKR_OK ) || ( xResult == CKR_CRYPTOKI_ALREADY_INITIALIZED ) )
+    {
+        xResult = xGetSlotList( &pxSlotId, &xSlotCount );
+    }
+
+    if( xResult == CKR_OK )
+    {
+        /* Check if token is initialized */
+        pxTokenInfo = pvPortMalloc( sizeof( CK_TOKEN_INFO ) );
```
dan4thewin: Do we need to check this for failure?

alexa-noxon: Yes good catch updated the code now.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -764,6 +764,75 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
...
+    }
+
+    if( CKR_OK == xResult )
+    {
+        xTokenFlags = pxTokenInfo->flags;
+    }
+
+    if( ( CKR_OK == xResult ) && !( CKF_TOKEN_INITIALIZED & xTokenFlags ) )
+    {
+        /* If token is not initialized - initialize it */
```
dcgaws: "If the token is not initialized, initialize it."

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -1497,9 +1492,9 @@ TEST( Full_PKCS11_EC, AFQP_Sign )
     CK_OBJECT_HANDLE xPublicKeyHandle;
     CK_OBJECT_HANDLE xCertificateHandle;
     /* Note that mbedTLS does not permit a signature on all 0's. */
```
dcgaws: I don't think that's accurate. Rather, the issue is in trying to do the DSA computation with a zero exponent.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: fixed.

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
alexa-noxon: thanks for pointing that out- fixed the comment here and other places.

PR #1001, opened 2019-07-30
Modify PKCS #11 tests to destroy test credentials rather than default…

alexa-noxon:
… credentials

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
+++ b/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
@@ -627,33 +627,22 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
-CK_RV xDestroyCredentials( CK_SESSION_HANDLE xSession )
+CK_RV xDestroyProvidedCredentials( CK_SESSION_HANDLE xSession,
+                                   CK_BYTE_PTR * ppxPkcsLabels,
+                                   CK_OBJECT_CLASS * xClass,
+                                   CK_ULONG ulCount )
 {
     CK_RV xResult;
     CK_FUNCTION_LIST_PTR pxFunctionList;
     CK_OBJECT_HANDLE xObjectHandle;
     CK_BYTE * pxLabel;
     uint32_t uiIndex = 0;
-    CK_BYTE * pxPkcsLabels[] =
-    {
-        ( CK_BYTE * ) pkcs11configLABEL_DEVICE_CERTIFICATE_FOR_TLS,
-        ( CK_BYTE * ) pkcs11configLABEL_CODE_VERIFICATION_KEY,
-        ( CK_BYTE * ) pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS,
-        ( CK_BYTE * ) pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS
-    };
-    CK_OBJECT_CLASS xClass[] =
-    {
-        CKO_CERTIFICATE,
-        CKO_PUBLIC_KEY,
-        CKO_PRIVATE_KEY,
-        CKO_PUBLIC_KEY
-    };
 
     xResult = C_GetFunctionList( &pxFunctionList );
 
-    for( uiIndex = 0; uiIndex < sizeof( pxPkcsLabels ) / sizeof( pxPkcsLabels[ 0 ] ); uiIndex++ )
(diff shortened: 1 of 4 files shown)


demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -627,33 +627,22 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
-CK_RV xDestroyCredentials( CK_SESSION_HANDLE xSession )
+CK_RV xDestroyProvidedCredentials( CK_SESSION_HANDLE xSession,
```
dcgaws: I don't think that credentials is the right term. For one thing, the only type of credential we directly support with this code is X.509 certificates and private keys. For another thing, this helper routine is sufficiently abstract that it could be used to delete any type of object by label + class, including those that may not be used for authentication.

alexa-noxon: I will rename this to xDestroyProvidedObjects.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -627,33 +627,22 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
...
-        CKO_PUBLIC_KEY
-    };
 
     xResult = C_GetFunctionList( &pxFunctionList );
 
-    for( uiIndex = 0; uiIndex < sizeof( pxPkcsLabels ) / sizeof( pxPkcsLabels[ 0 ] ); uiIndex++ )
+    for( uiIndex = 0; uiIndex < ulCount; uiIndex++ )
     {
-        pxLabel = pxPkcsLabels[ uiIndex ];
+        pxLabel = ppxPkcsLabels[ uiIndex ];
```
dan4thewin: Below, there's a while loop surrounded by an if statement, both checking the same expression.  The if statement is redundant.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -627,33 +627,22 @@ CK_RV xProvisionCertificate( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
-CK_RV xDestroyCredentials( CK_SESSION_HANDLE xSession )
+CK_RV xDestroyProvidedCredentials( CK_SESSION_HANDLE xSession,
+                                   CK_BYTE_PTR * ppxPkcsLabels,
+                                   CK_OBJECT_CLASS * xClass,
+                                   CK_ULONG ulCount )
```
dan4thewin: Is there a reason that the types of ulCount and uiIndex are different?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -693,6 +682,33 @@ CK_RV xDestroyCredentials( CK_SESSION_HANDLE xSession )
     return xResult;
 }
 
+CK_RV xDestroyCredentials( CK_SESSION_HANDLE xSession )
+{
+    CK_RV xResult;
+    uint32_t uiIndex = 0;
```
dan4thewin: unused?

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -327,6 +327,33 @@ static MultithreadTaskParams_t xGlobalTaskParams[ pkcs11testMULTI_THREAD_TASK_CO
 #define pkcs11testALL_BITS    ( ( 1 << pkcs11testMULTI_THREAD_TASK_COUNT ) - 1 )
 
 
+static CK_RV prvDestroyTestCredentials( void )
```
dan4thewin: Could you just call `xDestroyCredentials(xGlobalSession)` ?

alexa-noxon: No- I need to distinguish between destroying the test credentials and the regular ones for secure element testing.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
alexa-noxon: yes, fixed

PR #1017, opened 2019-08-02
lwIP 2.1.2 redux

dan4thewin:
This PR replaces lwIP 2.0.3 with lwIP 2.1.2, along with fixes for breakage in the related Cypress code.  I've also added code to wake a task blocked on an mbox fetch if that mbox is freed.


cobusve:
Looks good

jainamt:
Dan, can you please update Changelog file as well. 

Also, for the changes done on top of the new LWIP, do we plan to send those patches upstream ?

PR #1018, opened 2019-08-05
Retrieve the JITR/JITP certificate from PKCS11

bryan-hunt:
Update TLS implementation to retrieve the JITR/JITP certificate from PKCS11

Description
-----------
iot_tls.c uses keyJITR_DEVICE_CERTIFICATE_AUTHORITY_PEM as a constant rather than retrieving the certificate from the PKCS11 interface where is was previously written (and where it would naturally reside permanently).

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/tls/src/iot_tls.c
+++ b/libraries/freertos_plus/standard/tls/src/iot_tls.c
@@ -562,12 +562,12 @@ static int prvInitializeClientCredential( TLSContext_t * pxCtx )
     }
 
     /* Free certificate buffer since the parsing is complete */
-    if (NULL != pxCertificate)
+    if( NULL != pxCertificate )
     {
-        vPortFree(pxCertificate);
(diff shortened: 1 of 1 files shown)


bryan-hunt:
\bot run checks

alexa-noxon:
The spirit of this change was brought into master here with the merge of ECC608a + WinSim feature branch.
https://github.com/aws/amazon-freertos/commit/8a0968bc2c6699c5c8c0dc0e383466cfe3dd7783#diff-5ddbe5703615970ce07164e69e8e5010R622, where JITP certificate is retrieved from PKCS #11 module if the JITR certificate is blank in aws_clientcredential_keys.h. 

There is still work to be done to always retrieve a JITP/JITR certificate from PKCS #11, rather than only when it is not present in the header. 

Closing the PR.

Thanks for your help.

PR #1029, opened 2019-08-06
Fix MQTT demo calculation for counting incoming messages

gordonwang0:
Description
-----------

Checklist:
----------

- [x] My code is Linted.


demos/mqtt/iot_demo_mqtt.c
```diff
@@ -719,30 +719,6 @@ static int _publishAllMessages( IotMqttConnection_t mqttConnection,
...
-                break;
-            }
-        }
-
-        if( i == IOT_DEMO_MQTT_PUBLISH_BURST_SIZE )
-        {
-            IotLogInfo( "All publishes received." );
-        }
-    }
-
```
jainamt: Is this not needed?

gordonwang0: Correct, the change to the calculation above means this block is no longer needed.

PR #1034, opened 2019-08-07
Feature HTTPS Client Library v1.0.0

sarenameas:
Feature HTTPS Client Library 

Description
-----------
This PR encompasses both the Synchronous and Asynchronous API.
This first release of the HTTPS Client Library supports only the GET and HEAD requests.
Please feat/http_test for the unit and integration tests for this feature.

Checklist:
----------

- [ ] My code is Linted.


demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * @brief HTTP standard header field "Range".
+ */
+#define RANGE_HEADER_FIELD                   "Range"
+#define RANGE_HEADER_FIELD_LENGTH            ( 5 )
+
+/**
+ * @brief HTTP standard header value for requesting a range of bytes from 0 to 0.
+ *
+ * This is used to get the size of the file from S3. Performing a HEAD request with S3 requires generating a Sigv4
```
dcgaws: But not for a pre-signed URL right?

demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,227 @@
...
+    respSyncInfo.bodyLen = 0;  /* Since we are not saving any response body the length is 0. */
+
+    /* Set the request configurations. */
+    fileSizeReqConfig.pPath = pPath;
+    fileSizeReqConfig.pathLen = pathLen;
+    fileSizeReqConfig.pHost = pAddress;
+    fileSizeReqConfig.hostLen = addressLen;
+    fileSizeReqConfig.method = IOT_HTTPS_METHOD_GET; /* Performing a HEAD request with S3 requires generating a Sigv4 signature
+                                                      * in an Authorization header field. We work around this by performing a GET
+                                                      * on Range: bytes=0-0. Then extracting the size of the file from the
```
dcgaws: Could these last two sentences be clarified?

demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,227 @@
...
+
+    /* String to store the Content-Range header field value. This header field as we are requesting in this demo is of
+     * the form: "Content-Range: bytes 0-0/FILESIZE", where file size would be the length of the maximum 32 bit integer
+     * which is 10.  Since the header field value "bytes 0-0/FILESIZE" is less than the maximum possible Range header
+     * field value, we size this string to the Range header field value.*/
+    char contentRangeValStr[ RANGE_VALUE_MAX_LENGTH ] = { 0 };
+    /* The location of the file size in the contentRangeValStr. */
+    char * pFileSizeStr = NULL;
+    /* Size in bytes of a single character. */
+    uint8_t sizeOfOneChar = 1;
```
dcgaws: Seems like this could instead be sizeof ( uint8_t ) inline below.

demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,227 @@
...
+    /* Get the file size by parsing the "bytes 0-0/FILESIZE" Content-Range header value string. */
+    httpsClientStatus = IotHttpsClient_ReadHeader( fileSizeRespHandle, CONTENT_RANGE_HEADER_FIELD, CONTENT_RANGE_HEADER_FIELD_LENGTH, contentRangeValStr, sizeof( contentRangeValStr ) );
+
+    if( httpsClientStatus != IOT_HTTPS_OK )
+    {
+        IotLogError( "Could find the Content-Range header in the response. Error code %d", httpsClientStatus );
+        return EXIT_FAILURE;
+    }
+
+    pFileSizeStr = strstr( contentRangeValStr, "/" ) + sizeOfOneChar;
```
dcgaws: What if the server didn't return a '/', or a '\0', before we read off the end of contentRangeValStr?

demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,227 @@
...
+    *pFileSize = ( uint32_t ) strtoul( pFileSizeStr, NULL, 10 );
```
dcgaws: What if there's a conversion error?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,804 @@
...
+#ifndef IOT_HTTPS_DEMO_ASYNC_TIMEOUT_MS
+    #define IOT_HTTPS_DEMO_ASYNC_TIMEOUT_MS    ( ( uint32_t ) 300000 )   /* 5 minute timeout for this demo. */
+#endif
+
+/** @endcond */
+
+/**
+ * @brief HTTP standard header field "Range".
+ */
+#define RANGE_HEADER_FIELD                     "Range"
```
dcgaws: Could these be in a common HTTP helper header?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,804 @@
...
+/* Size in bytes of the buffer used to store the response body (parts of it). This should be greater than or equal to
+ * the size of the file we want to download.*/
+#ifndef IOT_DEMO_HTTPS_RESP_BODY_BUFFER_SIZE
+    #define IOT_DEMO_HTTPS_RESP_BODY_BUFFER_SIZE    ( 512 )
+#endif
+
+/* Timeout in milliseconds to wait for all asynchronous requests to finish. This timeout starts when the last
+ * IOT_HTTPS_DEMO_MAX_ASYNC_REQUESTS sent has been scheduled. */
+#ifndef IOT_HTTPS_DEMO_ASYNC_TIMEOUT_MS
+    #define IOT_HTTPS_DEMO_ASYNC_TIMEOUT_MS    ( ( uint32_t ) 300000 )   /* 5 minute timeout for this demo. */
```
dcgaws: This would be a pretty long time for a customer to want to wait for one of our demos to complete. Have you encountered this kind of delay in testing?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,804 @@
...
+
+    IotLogInfo( "Inside of the append header callback for part %s", rangeValueStr );
+    /* The length of the range value string. */
+    uint32_t rangeValueLen = strlen( rangeValueStr );
+    /* Set the header for a range request and check the HTTPS Client library return code. */
+    IotHttpsReturnCode_t status = IotHttpsClient_AddHeader( reqHandle, RANGE_HEADER_FIELD, RANGE_HEADER_FIELD_LENGTH, rangeValueStr, rangeValueLen );
+
+    if( status != IOT_HTTPS_OK )
+    {
+        IotLogError( "Failed to write the header Range: %.*s into the request. With error code: %d", rangeValueLen, rangeValueStr, status );
```
dcgaws: If this fails, and the async caller doesn't know it, could the caller end up getting back a different amount of format of data than what they were expecting?

demos/https/iot_demo_https_common.c
sarenameas: Yea can't do a HEAD request with a pre-signed URL.

demos/https/iot_demo_https_common.c
sarenameas: /* Performing a HEAD request with S3 requires generating a Sigv4 signature
                                                      * in an Authorization header field. We work around this by performing a GET
                                                      * on Range: bytes=0-0. The size of the file is extracted from the Content-Range 
                                                      * header field in the response. */

demos/https/iot_demo_https_common.c
sarenameas: Does this make things more clear?
 * This is used to get the size of the file from S3. A HEAD request is forbidden with a presigned URL. Performing a 
 * HEAD request with S3 requires generating a Sigv4 signature in an Authorization header field. We work around this by 
 * performing a GET on Range: bytes=0-0. Then extracting the size of the file from the Content-Range header field in 
 * the response. */

demos/https/iot_demo_https_common.c
sarenameas: Thanks

demos/https/iot_demo_https_common.c
sarenameas: Good point. I will need to check for NULL returned.

demos/https/iot_demo_https_common.c
sarenameas: Thanks for pointing this out. I will check for 0 or ULONG_MAX returned.

demos/https/iot_demo_https_s3_download_async.c
sarenameas: Moved to the iot_demo_https_common.h.

demos/https/iot_demo_https_s3_download_async.c
sarenameas: No I have not encountered that time limit. I just wanted to make it this long just in case the server take a long time to respond. The demo WILL finish before 5 minutes.

demos/https/iot_demo_https_s3_download_async.c
sarenameas: I will add cancellation for failures.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3326 @@
...
+            pHttpsResponse->pBodyInHeaderBuf = ( uint8_t * ) ( pLoc );
+            pHttpsResponse->pBodyCurInHeaderBuf = pHttpsResponse->pBodyInHeaderBuf;
+        }
+
+        /* If there is a chunk encoded body in the header buffer, we will want to overwrite the chunk headers with the
+         * actual body. This is so that when the application calls IotHttpsClient_ReadResponseBody(), in the
+         * readReadyCallback(), we can pass the body into the body buffer provided right away. */
+        if( pHttpsResponse->pBodyCurInHeaderBuf != ( uint8_t * ) pLoc )
+        {
+            memcpy( pHttpsResponse->pBodyCurInHeaderBuf, pLoc, length );
```
dcgaws: This block would happen when we just got the remainder, or all, of the head, plus a chunk of the body? If so, could that be explained in more detail in the comments?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3326 @@
...
+    }
+    else
+    {
+        pHttpsConnection->timeout = pConnInfo->timeout;
+    }
+
+    /* pNetworkInterface contains all the routines to be able to send/receive data on the network. */
+    pHttpsConnection->pNetworkInterface = pConnInfo->pNetworkInterface;
+
+    /* The address from the connection conf
```
dcgaws: What's the rationale behind this macro?

dcgaws: "parsing the"

dcgaws: If both of these fields should include the NULL, then why not just compare that they're equal? This line is confusing.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3326 @@
...
+{
+    HTTPS_FUNCTION_ENTRY( IOT_HTTPS_OK );
+
+    IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;
+
+    /* The maximum string length of the ALPN protocols is configured in IOT_HTTPS_MAX_ALPN_PROTOCOLS_LENGTH.
+     * The +1 is for the NULL terminator needed by IotNetworkCredentials_t.pAlpnProtos. */
+    char pAlpnProtos[ IOT_HTTPS_MAX_ALPN_PROTOCOLS_LENGTH + 1 ] = { 0 };
+
+    /* The maximum string length of the Server host name is configured in IOT_HTTPS_MAX_HOST_NAME_LENGTH.
```
dcgaws: There is a max value defined by the DNS standard. Can we just use that?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3326 @@
...
+     * This +1 is for the NULL terminator needed by IotNetworkServerInfo_t.pHostName. */
+    char pHostName[ IOT_HTTPS_MAX_HOST_NAME_LENGTH + 1 ] = { 0 };
```
dcgaws: What's the rationale for making this based on static configuration, and sticking it on the stack? If our copy isn't persisted beyond the scope of this function anyway, why not pass the caller's hostname pointer through to the network stack as-is?

libraries/c_sdk/standard/https/include/iot_https_client.h
```diff
@@ -0,0 +1,890 @@
...
+ * connInfo.port = 443;
+ * connInfo.flags = 0;
+ * connInfo.pAlpnProtocols = "alpnproto0,alpnproto1"
+ * connInfo.pCaCert = HTTPS_TRUSTED_ROOT_CA; // defined elsewhere
+ * connInfo.caCertLen = sizeof( HTTPS_TRUSTED_ROOT_CA );
+ * connInfo.userBuffer.pBuffer = pConnUserBuffer;
+ * connInfo.userBuffer.bufferLen = connectionUserBufferMinimumSize;
+ * connInfo.pClientCert = TLS_CLIENT_CERT;
+ * connInfo.clientCertLen = sizeof( TLS_CLIENT_CERT );
+ * connInfo.pPrivateKey = TLS_CLIENT_PRIV_KEY;
```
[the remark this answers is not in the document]
sarenameas: These values are not used. The private key and client certificate provisioned at the start of the application are used by the network layer. These parameters are here because these parameters still exist in the network interface. In the future we are going to want to allow the user to customize their key into the application while at the same time still allowing usage of secure element keys. But how these parameters look and are passed into applications still needs to be decided upon.

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,782 @@
...
+        IotLogError( "An error occurred in IotHttpsClient_GetUrlAddress() with error code %d.", httpsClientStatus );
+        IOT_SET_AND_GOTO_CLEANUP( EXIT_FAILURE );
+    }
+
+    /* Set the connection configurations. */
+    connConfig.pAddress = pAddress;
+    connConfig.addressLen = addressLen;
+    connConfig.port = IOT_DEMO_HTTPS_PORT;
+    /* We disable SNI here because the address specified includes the S3 bucket name. */
+    connConfig.flags |= IOT_HTTPS_DISABLE_SNI;
```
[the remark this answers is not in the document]
sarenameas: Yes, because the subject of the certificate must be just "s3.amazonaws.com" or "amazonaws.com", but the address that I parse from the URL is "my-bucket-name.s3.amazonaws.com" which is not part of the certificate.

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,782 @@
...
+            IotClock_SleepMs( GET_FREE_REQUEST_RETRY_WAIT_TIME_MS );
+            continue;
+        }
+
+        /* Set the HTTP request configurations. */
+        _pReqConfigs[ reqIndex ].pPath = pPath;
+
+        /* The path is everything that is not the address. It also includes the query. So we get the strlen( pPath ) to
+         * acquire everything following in IOT_DEMO_HTTPS_PRESIGNED_GET_URL. */
+        _pReqConfigs[ reqIndex ].pathLen = strlen( pPath );
```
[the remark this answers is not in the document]
sarenameas: Yes moved to after parsing the path from the URL in line 563. 
```
/* The path is everything that is not the address. It also includes the query. So we get the strlen( pPath ) to
 * acquire everything following in IOT_DEMO_HTTPS_PRESIGNED_GET_URL. */
pathLen = strlen( pPath );
```

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -0,0 +1,782 @@
...
+        _pRespConfigs[ reqIndex ].pSyncInfo = NULL;
+
+        /* Get the Range header value string. */
+        int numWritten = snprintf( _pDownloadDatas[ reqIndex ].rangeValueStr,
+                                   sizeof( _pDownloadDatas[ reqIndex ].rangeValueStr ),
+                                   "bytes=%u-%u",
+                                   ( unsigned int ) curByte,
+                                   ( unsigned int ) ( curByte + numReqBytes - 1 ) );
+
+        if( numWritten < 0 )
```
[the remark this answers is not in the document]
sarenameas: Will update thanks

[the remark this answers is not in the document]
sarenameas: from cplusplus.com:
>The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.
If an encoding error occurs, a negative number is returned.
Notice that only when this returned value is non-negative and less than n, the string has been completely written.
>

So I would need to check for a failure if:
(numWritten < 0 ) and (numWritten >= sizeof(buffer))

libraries/c_sdk/standard/https/include/types/iot_https_types.h
```diff
@@ -0,0 +1,866 @@
...
+ * This buffer size is calculated to fit the HTTP request line and the default headers. It does not account for the
+ * length of the path in the request line nor does it account for the length of the host name. It also does not account
+ * for extra headers that the application may add. These sizes need to be accounted for by the application when
+ * assigning a buffer.
+ *
+ * By the application providing the memory for the internal context, no memory is needed to be allocated internally to
+ * the library for the internal context. The application has control over the memory allocation related to the request,
+ * response, and connection.
+ */
+extern const uint32_t requestUserBufferMinimumSize;
```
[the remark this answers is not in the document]
sarenameas:
```
 * A typical value for sizing the request user buffer for the request context is 512 bytes. See the example below.
 * @code{c}
 * uint8_t requestUserBuffer[512] = { 0 };
 * IotHttpsRequestInfo_t requestInfo = IOT_HTTPS_REQUEST_INFO_INITIALIZER;
 * requestInfo.userBuffer.pBuffer = requestUserBuffer;
 * @endcode
```

libraries/c_sdk/standard/https/include/types/iot_https_types.h
```diff
@@ -0,0 +1,866 @@
...
+ */
+typedef struct IotHttpsSyncRequestInfo
+{
+    /**
+     * Pointer to the HTTP message body.
+     *
+     * For a request this is the file or data we want to send.  The data is separated from the headers for the
+     * flexibility to point to an already established file elsewhere in memory.
+     *
+     * For a response this where to receive the response entity body.
```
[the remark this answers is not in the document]
sarenameas: Thanks.

libraries/c_sdk/standard/https/include/types/iot_https_types.h
```diff
@@ -0,0 +1,866 @@
...
+#define IOT_HTTPS_ASYNC_INFO_INITIALIZER           { 0 }
+/** @brief Initializer for #IotHttpsConnectionInfo_t. */
+#define IOT_HTTPS_CONNECTION_INFO_INITIALIZER      { 0 }
+/** @brief Initializer for #IotHttpsRequestInfo_t. */
+#define IOT_HTTPS_REQUEST_INFO_INITIALIZER         { 0 }
+/** @brief Initializer for #IotHttpsResponseInfo_t. */
+#define IOT_HTTPS_RESPONSE_INFO_INITIALIZER        { 0 }
+/* @[define_https_initializers] */
+
+/* Amazon FreeRTOS network include for the network types below. */
```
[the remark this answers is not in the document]
sarenameas: Update to be more generic.

libraries/c_sdk/standard/https/include/types/iot_https_types.h
```diff
@@ -0,0 +1,866 @@
...
+typedef struct IotHttpsSyncRequestInfo
+{
+    /**
+     * Pointer to the HTTP message body.
+     *
+     * For a request this is the file or data we want to send.  The data is separated from the headers for the
+     * flexibility to point to an already established file elsewhere in memory.
+     *
+     * For a response this where to receive the response entity body.
+     * If the length of the buffer provided to store response body is smaller than the amount of body received,
```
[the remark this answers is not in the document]
sarenameas: Thanks.

libraries/c_sdk/standard/https/include/types/iot_https_types.h
```diff
@@ -0,0 +1,866 @@
...
+ *
+ * This parameter is used to connection in @ref https_client_function_connect.
+ *
+ * @note The lengths of the strings in this struct should not include the NULL
+ * terminator. Strings in this struct do not need to be NULL-terminated.
+ */
+typedef struct IotHttpsConnectionInfo
+{
+    /**
+     * @brief Remote server address that is DNS discoverable.
```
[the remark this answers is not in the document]
sarenameas: I have not tried that.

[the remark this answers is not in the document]
sarenameas: We support IP addresses as much as iot_network.h does, which it looks like it does not..

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3326 @@
...
+    }
+
+    /* If the IotHttpsClient_ReadHeader() was called, then we check for the header field of interest. */
+    if( pHttpsResponse->bufferProcessingState == PROCESSING_STATE_SEARCHING_HEADER_BUFFER )
+    {
+        if( pHttpsResponse->readHeaderFieldLength != length )
+        {
+            pHttpsResponse->foundHeaderField = false;
+        }
+        else if( strncmp( pHttpsResponse->pReadHeaderField, pLoc, length ) == 0 )
```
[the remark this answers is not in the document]
sarenameas: No because we check the lengths matching above:
```
        if( pHttpsResponse->readHeaderFieldLength != length )
        {
            pHttpsResponse->foundHeaderField = false;
        }
```

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: This macro uses sizeof to get the length of the string for the common string literal macros that I use throughout. 
I don't want to make assumptions about the size of char so to save code I wrote:
`#define FAST_MACRO_STRLEN( x )    ( ( sizeof( x ) / sizeof( char ) ) - 1 )`
This macros saves some time by replacing some excessive strlen() usage in the past.

sarenameas: respHandle->readHeaderValueLength will not include a NULL.

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: It is defaulted to that in iot_https_internal.h:

```
#ifndef IOT_HTTPS_MAX_HOST_NAME_LENGTH
    #define IOT_HTTPS_MAX_HOST_NAME_LENGTH         ( 255 ) /* Per FQDN, the maximum host name length is 255 bytes. */
#endif
```

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: I create an array here because in IotHttpsConnectionInfo_t I do not necessitate that pAddress storing the host name needs a NULL terminator. There is an accompanying addressLen parameter. I also made the parameter a const as it must not change anyways.
```
 */
typedef struct IotHttpsConnectionInfo
{
    /**
     * @brief Remote server address that is DNS discoverable.
     *
     * For example: avs-alexa-na.amazon.com.
     */
    const char * pAddress;
    uint32_t addressLen; /**< @brief remote address length. */
    . . .
```
Because of this I cannot guarantee that the IotHttpsConnectionInfo_t.pAddress passed in by the user will be NULL terminated. 
The network interface structure that consumes the host name needs a NULL terminator, see iot_network.h:
```
typedef struct IotNetworkServerInfo
{
    const char * pHostName; /**< @brief Server host name. Must be NULL-terminated. */
    uint16_t port;          /**< @brief Server port in host-order. */
} IotNetworkServerInfo_t;
```

Because IotHttpsConnectionInfo_t.pAddress is a const I cannot simply append a NULL terminator, I need to allocate an array to append the NULL needed by the network interface. I did not want to use ANY mallocs in my implementation, so I created an array.

Please see: https://issues.amazon.com/issues/TS-8462

libraries/c_sdk/standard/https/src/private/iot_https_internal.h
```diff
@@ -0,0 +1,506 @@
...
+     * Post connection, this is set to false only after an implicit disconnect with a non-persistent request, an implicit
+     * disconnect with a network error, or an explicit disconnect with a call to @ref https_client_function_disconnect.
+     */
+    bool isConnected;
+    bool isDestroyed; /**< @brief true if the connection is already destroyed and we should call anymore  */
+
+    /**
+     * @brief true if the a connection is in the process of disconnecting.
+     *
+     * Mutexes are large and expensive, so it is better to just check a variable then return.
```
dcgaws: This comment is confusing because the it implies that thread safety is required, whereas this approach is not thread safe.

sarenameas: Just realized that a thread can context switch in the middle of:
```
    /* If this routine is currently is progress by another thread, for instance the taskpool worker that received a
     * network error after sending, then return right away because connection resources are being used. */
    if( connHandle->isDisconnecting )
    {
        HTTPS_SET_AND_GOTO_CLEANUP( IOT_HTTPS_BUSY );
    }
    else
    {
        connHandle->isDisconnecting = true;
    }
```
And see isDisconnecting as false twice.
So I will switch to a disconnection mutex.

libraries/c_sdk/standard/https/include/types/iot_https_types.h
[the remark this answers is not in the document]
sarenameas: SOCKETS_GetHostByName() should be OK with an IP address, so and IP address should still be good here.

demos/https/iot_demo_https_s3_download_async.c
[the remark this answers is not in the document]
sarenameas: Updated to allow SNI. Not sure what I was confusing this with before. Thanks.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -0,0 +1,3362 @@
...
+
+    pHttpsConnection = ( _httpsConnection_t * ) ( pConnInfo->userBuffer.pBuffer );
+
+    /* Start with the disconnected state. */
+    pHttpsConnection->isConnected = false;
+
+    /* Initialize disconnection state keeper. */
+    pHttpsConnection->isDestroyed = false;
+
+    /* Init
```
dcgaws: At this point we'll be holding three locks. That's a design that's tough to thoroughly test, maintain, and support. Could we instead use a single mutex to protect all fields on the context handle structure?

demos/https/iot_demo_https_common.c
```diff
@@ -0,0 +1,225 @@
...
+
+    pFileSizeStr = strstr( contentRangeValStr, "/" );
+
+    if( pFileSizeStr == NULL )
+    {
+        IotLogError( "'/' not present in Content-Range header value: %s.", contentRangeValStr );
+        return EXIT_FAILURE;
+    }
+
+    pFileSizeStr += sizeOfOneChar;
```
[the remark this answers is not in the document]
sarenameas: Yes

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: Yes changing.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2694,7 +2659,7 @@ IotHttpsReturnCode_t IotHttpsClient_Disconnect( IotHttpsConnectionHandle_t connH
      * on this function that it can proceed with the disconnecting activities. */
     if( connHandle != NULL )
```
dcgaws: How could this suddenly be NULL?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2667,15 +2636,11 @@ IotHttpsReturnCode_t IotHttpsClient_Disconnect( IotHttpsConnectionHandle_t connH
...
     /* Remove all pending requests. If this routine is called from the application context and there is a
      * network receive callback in process, this routine will wait in _networkDestroy until that routine returns.
      * If this is routine is called from the network receive callback context, then the destroy happens after the
      * network receive callback context returns. */
-    IotMutex_Lock( &( connHandle->reqQMutex ) );
     IotDeQueue_RemoveAll( &( connHandle->reqQ ), NULL, 0 );
-    IotMutex_Unlock( &( connHandle->reqQMutex ) );
 
     /* Do not attempt to destroy an already destroyed connection. This can happen when the user calls this function and
      * IOT_HTTPS_BUSY is returned. */
```
dcgaws: Is there another routine that can be used to destroy the connection handle?

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: If the user passes in a NULL connHandle to the routine. This is within the cleanup goto statements.

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: Automatic disconnect happens on parsing and network errors during the taskpool job that sends the request and the network receive task that receives the response.

PR #1039, opened 2019-08-08
Escape configASSERT in FreeRTOSConfig.h for CBMC proof checking

markrtuttle:
This is a two-line modification of a core FreeRTOS include file that makes it easy for us to test all configASSERTs in FreeRTOS (and HTTP) code.

This patch just encloses the definition of configASSERT within #ifndef ... #endif so
it can be redefined from the command line during CBMC proof checking
to invoke __CPROVER_assert to check the assertions with CBMC.

### DIFF
--- a/vendors/pc/boards/windows/aws_demos/config_files/FreeRTOSConfig.h
+++ b/vendors/pc/boards/windows/aws_demos/config_files/FreeRTOSConfig.h
@@ -124,7 +124,9 @@ void vConfigureTimerForRunTimeStats( void );
 /* Assert call defined for debug builds. */
 extern void vAssertCalled( const char * pcFile,
                            uint32_t ulLine );
+#ifndef configASSERT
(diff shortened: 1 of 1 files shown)


sarenameas:
> Looks good to me, thank you.
> 
> There are 35 different `FreeRTOSConfig.h` in the codebase. You're changing only the `pc/.../aws_demos` version because that's the one we `#include` from CBMC proofs.
> 
> I wonder whether we should make this change to _every_ `FreeRTOSConfig.h`, so that future changes to the build system don't cause inconsistent behavior on the `pc` port compared to all the others. But this is just idle speculation, I'll let a FreeRTOS reviewer comment on whether this is likely.

I don't forsee any inconsistent behavior with this.

markrtuttle:
As usual, I'd like to merge this now and put this idea on the back log for after the release.

PR #1053, opened 2019-08-11
Improve CBMC coverage on HTTP synchronous proofs to all but one line.

markrtuttle:
The only uncovered lines are blocks of code guarded by a test for the
asynchronous case, plus one line in the cleanup code for SendSync.

### DIFF
--- a/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
+++ b/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
@@ -12,8 +12,11 @@
 void harness() {
   IotHttpsResponseHandle_t respHandle = allocate_IotResponseHandle();
   initialize_IotResponseHandle(respHandle);
-  if (respHandle)
+  if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
+    // ???: replace assumption above with initialization
+    respHandle->httpParserInfo.readHeaderParser.data = respHandle;
+  }
   uint32_t pContentLength;
   IotHttpsClient_ReadContentLength(respHandle, &pContentLength);
 }
--- a/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
+++ b/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
@@ -14,7 +14,9 @@ void harness() {
   initialize_IotResponseHandle(respHandle);
   if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
-    // ???: replace assumption above with initialization
+    // Until a CBMC issue is addressed
+    // https://github.com/diffblue/cbmc/issues/5004
+    // copy an assumption from is_valid above as a simple assignment below.
     respHandle->httpParserInfo.readHeaderParser.data = respHandle;
   }
   uint32_t pContentLength;
--- a/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
+++ b/tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
@@ -18,7 +18,9 @@ void harness() {
   initialize_IotResponseHandle(respHandle);
   if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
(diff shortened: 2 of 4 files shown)


tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -55,10 +64,21 @@ size_t http_parser_execute (http_parser *parser,
...
+  if (_httpsResponse->foundHeaderField) {
+    size_t valueLen;
+    __CPROVER_assume(valueLen <= len);
+    _httpsResponse->pReadHeaderValue = malloc(valueLen+1);
+    _httpsResponse->pReadHeaderValue[valueLen] = 0;
+    _httpsResponse->readHeaderValueLength = valueLen;
+  }
+
+  // ???: Should this be valueLen?
+  // ???: Does it suggest a problem that returning any length works?
```
sarenameas: This should not be valueLen, the amount parsed will always be <= len. So the return value of this routine is <= len.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -23,6 +29,9 @@ void *safeMalloc(size_t xWantedSize) {
  * responseHandle, requestHandle and connectionHandle similarly.
  */
 
+// TODO: Replace this with a model allowing unconstrained sizes,
+// or at least allowing a small set of differing sizes.
```
sarenameas: That is a good idea for the data part of the structure.

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
```diff
@@ -12,8 +12,11 @@
 void harness() {
   IotHttpsResponseHandle_t respHandle = allocate_IotResponseHandle();
   initialize_IotResponseHandle(respHandle);
-  if (respHandle)
+  if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
+    // ???: replace assumption above with initialization
+    respHandle->httpParserInfo.readHeaderParser.data = respHandle;
```
sarenameas: Would this be better in `allocate_IotResponseHandle()`?

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
```diff
@@ -18,6 +18,9 @@ void harness() {
   initialize_IotResponseHandle(respHandle);
   if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
+    // ???: replace assumption above with initialization
+    respHandle->httpParserInfo.readHeaderParser.data = respHandle;
```
sarenameas: Would this be better in `allocate_IotResponseHandle()`?

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadContentLength/IotHttpsClient_ReadContentLength_harness.c
markrtuttle: Yes, this is the CBMC issue I had to work around: I took the assumption from is_valid_IoTResponseHandle and copied it here an assignment in the test harness.  I could put it in allocate_IotResponseHandle as you suggest, but I'd like to keep it here so we remember to delete it when CBMC is fixed.

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
markrtuttle: Same as above.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
markrtuttle: That is what I meant.  I propose we do this after the release as part of the next package.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
markrtuttle: ValueLen is constrained to be between 0 <= valueLen <= len.   So that is the value you are asking for, and I will make that change.

markrtuttle: I'm changing the name to valueLength.  I think the name valueLen (a name you use a parameter to other functions) was causing confusion in this discussion.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
markrtuttle: I've written a better comment.

PR #1055, opened 2019-08-12
Let Socket_t and SocketSet_t point to a struct

htibosch:
FreeRTOS+TCP : Let the `Socket_t` and `SocketSet_t` point to a struct in stead of void

Description
-----------
In earlier days, the actual contents of objects types was often obfuscated by making them void pointers.
Within the libraries, the `void *` would be casted to the real object type. That is where lint starts complaining, e.g. here:

~~~c
void FreeRTOS_FD_SET( Socket_t xSocket, SocketSet_t xSocketSet, EventBits_t xSelectBits )
{
FreeRTOS_Socket_t *pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
SocketSelect_t *pxSocketSet = ( SocketSelect_t * ) xSocketSet;
~~~

Lint doesn't like the cast from `void *` to 'struct *`.

It would be nice when the anonymous types `Socket_t` and `SocketSet_t` point to "a struct".

I propose the following changes:

~~~c
 /* The socket type itself. */
-typedef void *Socket_t;
+struct xSOCKET;
+typedef struct xSOCKET *Socket_t;

 /* The SocketSet_t type is the equivalent to the fd_set type used by the
 Berkeley API. */
-typedef void *SocketSet_t;
+struct xSOCKET_SET;
+typedef struct xSOCKET_SET *SocketSet_t;
~~~

Note that the higher-level header file `iot_secure_sockets.h` has copied the definition of Socket_t, so it should be adapted as well:

~~~c
-typedef void * Socket_t;
+struct xSOCKET;
+typedef struct xSOCKET *Socket_t;
~~~

After the change, the compiler showed me that in a few locations, a cast had an unnecessary asterisk:

~~~c
-pxSocket->u.xTCP.pxHandleSent( ( Socket_t * )pxSocket, ulCount );
+pxSocket->u.xTCP.pxHandleSent( ( Socket_t )pxSocket, ulCount );
~~~

The compiler doesn't care about the difference between a `void*` and a `void**`.

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
I tested using the Microchip demo as well as the demo on Windows.
- [ ] My code is Linted.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1056, opened 2019-08-12
iot_demo_mqtt.c: test if the client identifier points to a non-empty …

htibosch:
iot_demo_mqtt.c: test if the client identifier points to a non-empty string

Description
-----------
In `iot_demo_mqtt.c` : when a client identifier is not provided by the user, one will be created as e.g. `iotdemo123`:

~~~c
    /* Use the parameter client identifier if provided. Otherwise, generate a
     * unique client identifier. */
    if( pIdentifier != NULL )
    {
        /* Use it. */
    }
    else
    {
        /* Generate a client identifier. */
    }
~~~

Looking back where `pIdentifier` is set, I find the macro `clientcredentialIOT_THING_NAME` which, by default, is defined as an empty string:

~~~c
#define clientcredentialIOT_THING_NAME               ""
~~~

The demo will fail, and it is hard to find the cause of the error message.

To avoid this little annoyance, I would like to propose this small change:

~~~c
-    if( pIdentifier != NULL )
+    if( ( pIdentifier != NULL ) && ( pIdentifier[0] != '\0' ) )
~~~

Checklist:
----------

i tested this together with PR #1055 on a Microchip board and in a Windows project.
- [ ] My code is Linted.

### DIFF
--- a/demos/mqtt/iot_demo_mqtt.c
+++ b/demos/mqtt/iot_demo_mqtt.c
@@ -431,7 +431,7 @@ static int _establishMqttConnection( bool awsIotMqttMode,
 
     /* Use the parameter client identifier if provided. Otherwise, generate a
      * unique client identifier. */
-    if( ( pIdentifier != NULL ) && ( pIdentifier[0] != '\0' ) )
+    if( ( pIdentifier != NULL ) && ( pIdentifier[ 0 ] != '\0' ) )
     {
         connectInfo.pClientIdentifier = pIdentifier;
         connectInfo.clientIdentifierLength = ( uint16_t ) strlen( pIdentifier );
--- a/demos/mqtt/iot_demo_mqtt.c
+++ b/demos/mqtt/iot_demo_mqtt.c
@@ -431,7 +431,7 @@ static int _establishMqttConnection( bool awsIotMqttMode,
 
     /* Use the parameter client identifier if provided. Otherwise, generate a
      * unique client identifier. */
-    if( pIdentifier != NULL )
+    if( ( pIdentifier != NULL ) && ( pIdentifier[0] != '\0' ) )
     {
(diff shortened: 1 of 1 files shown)


gordonwang0:
Hi @htibosch,

The MQTT spec [allows servers to support zero-length client identifiers](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349242), which is why this demo allows an empty client identifier.

> A Server MAY allow a Client to supply a ClientId that has a length of zero bytes...

It's probably better to change the default value of `clientcredentialIOT_THING_NAME` instead of excluding zero-length client identifiers.  Will get back to you on this.

htibosch:
Hello Gordon, the thing is that first-time users are not aware that they must set `clientcredentialIOT_THING_NAME` before running a demo. It would be nice if you can have the compiler warn for all incomplete `#define`'s (also those defining the URL and certificates), Thanks, Hein

dcgaws:
Hein, thank you again for the above. Would you be willing to submit the same change to the same file name in https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/v4_beta?

htibosch:
Hello Dan,
 
> Would you be willing to submit the same change to the same file
> name in https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/v4_beta?

The change has been merged to "mainline". Do you want me to submit the same change to mentioned tree?

I just cloned 'aws-iot-device-sdk-embedded-C', but there is no file called 'iot_demo_mqtt.c'?

htibosch:
> I just cloned 'aws-iot-device-sdk-embedded-C', but there is no file called 'iot_demo_mqtt.c'?

Found it and created a new [PR 543](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/543)

PR #1058, opened 2019-08-12
Fix mqtt example documentation

ArturBa:
Mqtt documentation

Description
-----------
Changed example documentation in mqtt

Checklist:
----------
- [x] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/mqtt/include/iot_mqtt.h
+++ b/libraries/c_sdk/standard/mqtt/include/iot_mqtt.h
@@ -256,7 +256,7 @@ void IotMqtt_ReceiveCallback( void * pNetworkConnection,
  *
  * // Example network abstraction types.
  * IotNetworkServerInfo_t serverInfo = { ... };
- * cIotNetworkCredentials_t credentialInfo = { ... };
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/mqtt/include/iot_mqtt.h
```diff
@@ -256,13 +256,13 @@ void IotMqtt_ReceiveCallback( void * pNetworkConnection,
  *
  * // Example network abstraction types.
  * IotNetworkServerInfo_t serverInfo = { ... };
- * IotNetworkCredentialInfo_t credentialInfo = { ... };
+ * cIotNetworkCredentials_t credentialInfo = { ... };
```
gordonwang0: I don't think this change is needed.  Could you revert this line?  Thanks.

ArturBa: I've corrected misspelling (removing first c letter). 
Also cannot revert cause there is no such struct in AWS FreeRTOS. 
[Here](https://github.com/aws/amazon-freertos/blob/master/demos/defender/aws_iot_demo_defender.c#L207) you can check sample use of this structure

PR #1066, opened 2019-08-13
Feat: Add support for sending number of blocks with OTA data request.

pvyawaha:
Add support in OTA agent for controlling number of data blocks requested from streaming service.

Description
-----------

This change adds support in OTA agent for controlling number of data blocks requested from streaming service. A new parameter representing number of data blocks is added as a config that is sent with every data block request. For example if this parameter is set as 16 then OTA streaming service will send 16 Data Blocks in response.
This change also restarts the timer as soon as the requested number of data blocks are received instead of waiting for the request timer timeout. This helps in speeding up the OTA process by avoiding unnecessary wait for timer to expire.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_ota_cbor.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_ota_cbor.c
@@ -37,7 +37,7 @@
  * @brief Message field definitions, per the server specification.
  */
 
-#define OTA_CBOR_GETSTREAMREQUEST_ITEM_COUNT    5
+#define OTA_CBOR_GETSTREAMREQUEST_ITEM_COUNT    6
 
 /**
  * @brief Internal context structure for decoding CBOR arrays.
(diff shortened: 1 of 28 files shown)


libraries/freertos_plus/aws/ota/src/aws_ota_cbor.c
```diff
@@ -37,7 +37,7 @@
  * @brief Message field definitions, per the server specification.
  */
 
-#define OTA_CBOR_GETSTREAMREQUEST_ITEM_COUNT    5
+#define OTA_CBOR_GETSTREAMREQUEST_ITEM_COUNT    6
```
huguesBouvier: Could that parameter automatically scale?

ravibhagavandas: Unfortunately, the tiny cbor library needs to specify number of parameters upfront before creating the map. If needed otherwise we might need to use the non-finite length maps, which require changes in server side I believe.

PR #1068, opened 2019-08-13
Add network ReceiveUpto

gordonwang0:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -514,6 +514,9 @@ size_t IotNetworkAfr_Receive( void * pConnection,
     /* Cast network connection to the correct type. */
     _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
 
+    /* Caller should never request zero bytes. */
+    configASSERT( bytesRequested > 0 );
+
     /* Write the buffered byte. THIS IS A TEMPORARY WORKAROUND AND ASSUMES THIS
      * FUNCTION IS ALWAYS CALLED FROM THE RECEIVE CALLBACK. */
     if( pNetworkConnection->bufferedByteValid == true )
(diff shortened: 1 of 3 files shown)


libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -569,6 +573,49 @@ size_t IotNetworkAfr_Receive( void * pConnection,
...
+    {
+        *pBuffer = pNetworkConnection->bufferedByte;
+        bytesReceived = 1;
+        pNetworkConnection->bufferedByteValid = false;
+    }
+
+    /* Block and wait for incoming data. */
+    socketStatus = SOCKETS_Recv( pNetworkConnection->socket,
+                                 pBuffer + bytesReceived,
+                                 bufferSize - bytesReceived,
```
cobusve: If we call this funciton with 1 here we will end up calling SOCKETS_Recv with 0. Actually why do we have the special logic/workaround above for receiving the first byte?

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -569,6 +573,49 @@ size_t IotNetworkAfr_Receive( void * pConnection,
...
+                                 0 );
+
+    if( socketStatus <= 0 )
+    {
+        IotLogError( "Error %ld while receiving data.", ( long int ) socketStatus );
+    }
+    else
+    {
+        IotLogDebug( "Successfully received %lu bytes.",
+                     ( unsigned long ) bytesReceived );
```
cobusve: It looks like if bytes were received by the SOCKETS_REcv call they are not included in this count or the count returned at the end of this function? Looks like it always returns 0 or one regardless how many bytes more than 1 you got? Is this the intended behavior? Does not look correct to me ...

libraries/abstractions/platform/freertos/iot_network_freertos.c
sarenameas: That is because polling is not implemented yet.

libraries/abstractions/platform/freertos/iot_network_freertos.c
sarenameas: It is not the intended behavior. 
https://github.com/aws/amazon-freertos/pull/1078

PR #1071, opened 2019-08-13
Doxygen documentation update

lundinc2:
Doxygen minor updates to documentation

Description
-----------
Added missing libraries to doxygen guide. Added missing step of creating a tag folder.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/doc/config/common
+++ b/doc/config/common
@@ -91,7 +91,7 @@ ALIASES += constants_page{1}="@page \1_constants Constants"
 ALIASES += constants_brief{1}="@brief Defined constants of the \1.<br><br>Libraries may @c #`define` constants in their headers with special meanings. This page describes the meanings and uses of any constants defined by the \1. Related constants are shown in a single section on this page.<br>"
 
 # Alias for starting a functions page.
-ALIASES += functions_page{1}="@page \1_functions Functions"
+ALIASES += functions_page{2}="@page \1_functions Functions"
 ALIASES += functions_brief{1}="@brief Functions of the \1 library.<br><br>The \1 library consists of the following functions."
 
 # Alias for listing a single function on a functions page.
--- a/doc/guidance.md
+++ b/doc/guidance.md
@@ -14,13 +14,22 @@ This is about how to use Doxygen to maintain the API documents. There are three 
 	- on mac, run "brew install graphviz"
 - goto the root directory of "amazon-freertos" (this will be referenced as $ROOT_AFR_DIR)
 	- cd $ROOT_AFR_DIR
-- create a "tag" folder in "ROOT/doc
-- run Doxygen command a first time to generate the tags.
-- run Doxygen command a second time to generate the documents.
-	- doxygen doc/config/main
+- run Doxygen commands to generate the documents. Note: These files have circular dependencies so make sure to run each command twice.
+	- doxygen doc/config/linear_containers
+	- doxygen doc/config/static_memory
+	- doxygen doc/config/logging
 	- doxygen doc/config/secure_sockets
-
-- the entry doc is $ROOT_AFR_DIR/doc/output/main/html, open it with browser and verify it looks good
+	- doxygen doc/config/wifi
+	- doxygen doc/config/atomic
+	- doxygen doc/config/taskpool
+	- doxygen doc/config/platform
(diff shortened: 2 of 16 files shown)


doc/guidance.md
```diff
@@ -14,9 +14,22 @@ This is about how to use Doxygen to maintain the API documents. There are three
 	- on mac, run "brew install graphviz"
 - goto the root directory of "amazon-freertos" (this will be referenced as $ROOT_AFR_DIR)
 	- cd $ROOT_AFR_DIR
-- run Doxygen command to generate the documents 
-	- doxygen doc/config/main
+- create a tag folder to store doxygen generated tags
+	- mkdir $ROOT_AFR_DIR/doc/tag
```
lundinc2: Would it be better to just check in a tag folder?

doc/guidance.md
```diff
@@ -14,9 +14,22 @@ This is about how to use Doxygen to maintain the API documents. There are three
...
+- run Doxygen commands to generate the documents
```
lundinc2: Does order matter? Nothing seems broken but doxygen complains it can't find some tag files. Although there are circular dependencies for a lot of these tags

doc/guidance.md
```diff
@@ -14,9 +14,22 @@ This is about how to use Doxygen to maintain the API documents. There are three
...
+	- doxygen doc/config/linear_containers
+	- doxygen doc/config/static_memory
+	- doxygen doc/config/logging
 	- doxygen doc/config/secure_sockets
+	- doxygen doc/config/wifi
+	- doxygen doc/config/atomic
+	- doxygen doc/config/taskpool
+	- doxygen doc/config/platform
+	- doxygen doc/config/mqtt
```
huguesBouvier: missing posix

doc/guidance.md
cobusve: You have to run them in order to avoid the errors, but I beleive that there are circular links, which means you have to run them all twice to be sure all the links are good

doc/guidance.md
cobusve: What would be better is to change the tags to go to the output folder instead. Git does not allow empty folders, so just checking in an empty tag folder is not possible, you would have to place a dummy file which is messy.

I do not see the value of a separate tag folder, I would rather just place the tag files in the output folder and then it would all work without any issues always.

doc/guidance.md
lundinc2: The config file for posix appears to be missing

doc/config/common
```diff
@@ -91,7 +91,7 @@ ALIASES += constants_page{1}="@page \1_constants Constants"
 ALIASES += constants_brief{1}="@brief Defined constants of the \1.<br><br>Libraries may @c #`define` constants in their headers with special meanings. This page describes the meanings and uses of any constants defined by the \1. Related constants are shown in a single section on this page.<br>"
 
 # Alias for starting a functions page.
-ALIASES += functions_page{1}="@page \1_functions Functions"
+ALIASES += functions_page{2}="@page \1_functions Functions"
```
lundinc2: Fix for previous PR. Broke functions page, this will make it look normal again. [link](https://imgur.com/a/gzvfBNd)

doc/guidance.md
lundinc2: Done

doc/guidance.md
lundinc2: Done

doc/guidance.md
```diff
@@ -14,13 +14,22 @@ This is about how to use Doxygen to maintain the API documents. There are three
 	- on mac, run "brew install graphviz"
 - goto the root directory of "amazon-freertos" (this will be referenced as $ROOT_AFR_DIR)
 	- cd $ROOT_AFR_DIR
-- create a "tag" folder in "ROOT/doc
```
huguesBouvier: Why remove this?

lundinc2: Per feedback from @cobusve. Seems like it was an unnecessary folder. Now tag files just get created in output folder.

sarenameas: In the past (a few weeks ago lol), I had to create that folder otherwise I could not generate.

PR #1074, opened 2019-08-13
Increase stack size for Shadow demo on TI

gordonwang0:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/ti/boards/cc3220_launchpad/aws_demos/config_files/aws_demo_config.h
+++ b/vendors/ti/boards/cc3220_launchpad/aws_demos/config_files/aws_demo_config.h
@@ -52,7 +52,7 @@
 #endif
 
 #define democonfigSHADOW_DEMO_NUM_TASKS                ( 1 )
-#define democonfigSHADOW_DEMO_TASK_STACK_SIZE          ( configMINIMAL_STACK_SIZE * 4 )
+#define democonfigSHADOW_DEMO_TASK_STACK_SIZE          ( 1024 )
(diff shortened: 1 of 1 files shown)


vendors/ti/boards/cc3220_launchpad/aws_demos/config_files/aws_demo_config.h
```diff
@@ -52,7 +52,7 @@
 #endif
 
 #define democonfigSHADOW_DEMO_NUM_TASKS                ( 1 )
-#define democonfigSHADOW_DEMO_TASK_STACK_SIZE          ( configMINIMAL_STACK_SIZE * 4 )
+#define democonfigSHADOW_DEMO_TASK_STACK_SIZE          ( 1024 )
```
abhidixi11: To be consistent, can we keep the size in multiples of configMINIMAL_STACK_SIZE  rather than using constants ?

gordonwang0: We have been changing stack sizes to constants, as configMINIMAL_STACK_SIZE is different for every board.

sarenameas: Was this amount chosen because it just a large enough nice power of 2 number?

gordonwang0: Yes

PR #1083, opened 2019-08-15
Increase stack size of ST Shadow demo

gordonwang0:
Description
-----------
As in title.

Checklist:
----------

- [ ] My code is Linted.


leegeth:
> There is a discussion going on about the need of a stack size change here. We would need to analyse/discuss more to find out why this increase is required before we merge the changes in.

Discussed with Gordon today morning and got a confirmation that the debugging clearly showed that stack was overflowing and stack size change is required to address the problem

PR #1084, opened 2019-08-15
Merge the HTTPS Client asynchronous and synchronous demos into one demo

sarenameas:
The HTTPS Client asynchronous and synchronous S3 download demos were merged into one demo.
A user configuration IOT_DEMO_HTTPS_ASYNC switches between the demos and CONFIG_HTTPS_S3_DEMO_ENABLED
was defined instead of two separate configs. This is done because the demos share similarities and an advanced user would know to which detail they desired.

This was also done to better support the Online Configuration download in AWS IoT.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/https/CMakeLists.txt
+++ b/demos/https/CMakeLists.txt
@@ -1,7 +1,7 @@
 # AFR HTTPS demo
 afr_demo_module(https)
 
-afr_set_demo_metadata(ID "HTTPS_DEMO")
+afr_set_demo_metadata(ID "HTTPS_S3_DEMO")
 afr_set_demo_metadata(DESCRIPTION "Examples that demonstrate the HTTPS Client")
 afr_set_demo_metadata(DISPLAY_NAME "HTTPS Client S3 download")
 
--- a/demos/https/CMakeLists.txt
+++ b/demos/https/CMakeLists.txt
@@ -1,9 +1,9 @@
 # AFR HTTPS demo
 afr_demo_module(https)
 
-afr_set_demo_metadata(ID "HTTPS_S3_DEMO")
+afr_set_demo_metadata(ID "HTTPS_DEMO")
 afr_set_demo_metadata(DESCRIPTION "Examples that demonstrate the HTTPS Client")
-afr_set_demo_metadata(DISPLAY_NAME "HTTPS Client S3 download")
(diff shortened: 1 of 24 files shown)


demos/https/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # AFR HTTPS demo
 afr_demo_module(https)
 
-afr_set_demo_metadata(ID "HTTPS_DEMO")
+afr_set_demo_metadata(ID "HTTPS_S3_DEMO")
```
jainamt: I am concerned with this naming. I dont want the customers to even accidentally think that our http library can work only with S3.

demos/include/iot_demo_runner.h
```diff
@@ -94,10 +94,8 @@
     #endif
 #elif defined( CONFIG_BLE_GATT_SERVER_DEMO_ENABLED )
     #define DEMO_entryFUNCTION             vGattDemoSvcInit
-#elif defined( CONFIG_HTTPS_SYNC_DOWNLOAD_DEMO_ENABLED )
-    #define DEMO_entryFUNCTION             RunHttpsSyncDownloadDemo
-#elif defined( CONFIG_HTTPS_ASYNC_DOWNLOAD_DEMO_ENABLED )
-    #define DEMO_entryFUNCTION             RunHttpsAsyncDownloadDemo
+#elif defined( CONFIG_HTTPS_S3_DEMO_ENABLED )
+    #define DEMO_entryFUNCTION             RunHttpsS3Demo
```
jainamt: How about a generic name for this demo

demos/https/CMakeLists.txt
sarenameas: I named it it this because when we support upload next, I also want it it to be encompassed under this same demo. Perhaps we should be rid of the S3 part? I mean the MQTT demo is just called MQTT_DEMO even thought it connects to AWS IoT, but it can connect to any MQTT server.

demos/include/iot_demo_runner.h
sarenameas: Sure.

sarenameas:
Closing this in favor of current configuration of two separate demos.

PR #1098, opened 2019-08-16
Fix log message in ReceiveUpto

gordonwang0:
Description
-----------
An incorrect error log message would have been printed if bufferSize was 1.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -595,22 +595,26 @@ size_t IotNetworkAfr_ReceiveUpto( void * pConnection,
         pNetworkConnection->bufferedByteValid = false;
     }
 
-    /* Block and wait for incoming data. */
-    socketStatus = SOCKETS_Recv( pNetworkConnection->socket,
-                                 pBuffer + bytesReceived,
-                                 bufferSize - bytesReceived,
-                                 0 );
+    if( bufferSize - bytesReceived > 0 )
+    {
+        /* Block and wait for incoming data. */
+        socketStatus = SOCKETS_Recv( pNetworkConnection->socket,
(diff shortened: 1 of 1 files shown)


libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -595,23 +595,27 @@ size_t IotNetworkAfr_ReceiveUpto( void * pConnection,
...
-                                 pBuffer + bytesReceived,
-                                 bufferSize - bytesReceived,
-                                 0 );
-
-    if( socketStatus <= 0 )
-    {
-        IotLogError( "Error %ld while receiving data.", ( long int ) socketStatus );
-    }
-    else
+    if( bufferSize - bytesReceived > 0 )
```
sarenameas: Will something bad happen with the past logic of asking for zero bytes from SOCKETS_Recv?

gordonwang0: It worked in all past cases.  Though it may have depended on the network stack.

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -595,23 +595,27 @@ size_t IotNetworkAfr_ReceiveUpto( void * pConnection,
...
     {
-        bytesReceived += ( size_t ) socketStatus;
-        IotLogDebug( "Successfully received %lu bytes.",
-                     ( unsigned long ) bytesReceived );
+        /* Block and wait for incoming data. */
```
sarenameas: This comment assumes the socket is a blocking one?

gordonwang0: That's correct, the sockets are blocking.

PR #1118, opened 2019-08-20
fix stack garbage return and drop ineffectual yield

dan4thewin:
Full_DEFENDER tests all passing for cypress54

This PR is the same as #1117, but targeted at the release candidate.

### DIFF
--- a/libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
+++ b/libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
@@ -129,11 +129,9 @@ void sys_mbox_free( volatile sys_mbox_t *pxMailBox )
         pxMailBox->xMbox = NULL;
         taskEXIT_CRITICAL();
 
-        while( xTask != NULL )
+        if( xTask != NULL )
         {
             xTaskAbortDelay( xTask );
-            taskYIELD();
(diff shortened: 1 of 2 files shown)


libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
```diff
@@ -129,11 +129,9 @@ void sys_mbox_free( volatile sys_mbox_t *pxMailBox )
         pxMailBox->xMbox = NULL;
         taskEXIT_CRITICAL();
 
-        while( xTask != NULL )
+        if( xTask != NULL )
         {
             xTaskAbortDelay( xTask );
```
abhidixi11: I'm trying to understand how this code is suppose to work. What happens if xTaskAbortDelay() unblocks  xTask, but pxMailBox->xTask is not set to NULL; i.e.;  xTask gets blocked again on a receive message.

dan4thewin: In the fetching task, the next step after the blocking operation is to reread the mbox pointer.  If the fetching task was unblocked by this xTaskAbortDelay(), then the mbox pointer has been set to NULL, which prevents another read on the mbox.

PR #1121, opened 2019-08-20
Remove integer overflow from "room in buffer" calculation.

markrtuttle:
This patch removes from the HTTP function _addHeader a buffer overflow that results from an integer overflow in a calculation to test whether there is room in the buffer to add a header to the http request under construction.

This patch replaces a comparison of the form "space_needed + buffer_position > buffer_end" with "space_needed > buffer_end - buffer_position".

### DIFF
--- a/libraries/c_sdk/standard/https/src/iot_https_client.c
+++ b/libraries/c_sdk/standard/https/src/iot_https_client.c
@@ -1544,7 +1544,7 @@ static IotHttpsReturnCode_t _addHeader( _httpsRequest_t * pHttpsRequest,
      * (name:value\r\n). We need to add a "\r\n" at the end of headers. The use of
      * possibleLastHeaderAdditionalLength is to make sure that there is always
      * space for the last "\r\n". */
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -1544,7 +1544,7 @@ static IotHttpsReturnCode_t _addHeader( _httpsRequest_t * pHttpsRequest,
      * (name:value\r\n). We need to add a "\r\n" at the end of headers. The use of
      * possibleLastHeaderAdditionalLength is to make sure that there is always
      * space for the last "\r\n". */
-    if( ( additionalLength + possibleLastHeaderAdditionalLength + pHttpsRequest->pHeadersCur ) > ( pHttpsRequest->pHeadersEnd ) )
+    if( ( additionalLength + possibleLastHeaderAdditionalLength ) > ( pHttpsRequest->pHeadersEnd -  pHttpsRequest->pHeadersCur ) )
```
sarenameas: Looks like the extra space makes the code style check fail.

PR #1122, opened 2019-08-20
Remove buffer length assumptions from HTTP memory safety proofs (depends on #1121)

markrtuttle:
This patch removes all assumptions about buffer sizes from the CBMC proofs of synchronous HTTP memory safety.

The key idea is that the handle structs for requests/responses include pointers to buffers for the request/response headers and body.  In practice these buffers follow the handle structs in memory. The memory safety proofs hold, however, with these buffers located anywhere in memory.  So we now just allocate arbitrary buffers of arbitrary sizes for the handle structs to point to.

### DIFF
--- a/tools/cbmc/proofs/HTTP/global_state_HTTP.c
+++ b/tools/cbmc/proofs/HTTP/global_state_HTTP.c
@@ -12,44 +12,8 @@ IotHttpsRequestHandle_t allocate_IotRequestHandle();
  * without allocating memory.
  */
 void *safeMalloc(size_t xWantedSize) {
-  if(xWantedSize == 0) {
-    return NULL;
-  }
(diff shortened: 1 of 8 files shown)


tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -352,7 +341,7 @@ int is_valid_IotRequestHandle(IotHttpsRequestHandle_t pRequestHandle) {
   if (!required) return 0;
 
   int valid_headers =
```
karkhaz: Suggest to keep this on one lie

markrtuttle:
These changes were merged as part of https://github.com/aws/amazon-freertos/pull/1251

PR #1132, opened 2019-08-22
Add ISR (MSP) stack checking

DRNadler:
Add ISR (MSP) stack use checking to assure adequate stack space (GCC M4F)

Description
-----------
For ISP stack used by ISRs, add initialization and usage checking. Initialization is performed during FreeRTOS scheduler start as this memory area is used for general stack purposes before scheduler starts. Also comments clarified.

Checklist:
----------

- [ ] My code is Linted.


RichardBarry:
Thanks for your input - Let's continue to discuss here: https://sourceforge.net/p/freertos/discussion/382005/thread/f6773853ae/?limit=25#5e1e

Note that, depending on the outcome of that discussion, any update to a Cortex-M port needs to be done in all Cortex-M3, M4F, M7 and M0 ports for all five (?) supported compilers, many more files than in this pull request.

yuhui-zheng:
Two main reasons I'm closing this PR:
- This repo consumes FreeRTOS kernel official releases. So that no change is accepted for kernel in this repo.
- Quoting from Richard, if this change is to be accepted, all ARM ports need to be changed. Changing CM4F alone for GCC compiler is not sufficient. (Acknowledged you've volunteered to help in forum post.)

Though this PR will no longer be open, forum post is active and we have internal ticket to track this item. 

(Due to the forum migration, the discussion thread Richard mentioned above is moved to here -- https://forums.freertos.org/t/suggested-improvement-isr-stack-checking-gcc-m4/8020)

PR #1148, opened 2019-08-26
Add pkcs11 submodule from oasis-tcs/pkcs11

pavanmr94:
Description
-----------
Add pkcs11 submodule from oasis-tcs/pkcs11, and modify project files to point to new location. 

Checklist:
----------

- [ ] My code is Linted.


alexa-noxon:
:( I don't really like pointing towards a path called `/published/2-40-errata-1/` in our repo.  Can't we just maintain our own copy in a repo?  Regardless of whether we maintain the repo or they do, it iwll be work to do an update since we would have to change the file path in all the cmake and project folders.  

To me it seems more convenient to change the files in our repo than to change the path every time we do a update.  And our file paths will be prettier.

pavanmr94:
Not relevant anymore. Merged in https://github.com/aws/amazon-freertos/pull/1378

Issue #1131, opened 2019-08-22
ESP32: linking fails due to both Nimble and Bluedroid being built


matteoscordino:
**Describe the bug**
I have a CMake project that used to build before commit 3949bfa. 
Since then, the NimBLE stack has been introduced in the build and now linking fails due to multiple definitions of functions that are present in both NimBLE and Bluedroid. It seems for some reason both stack are built, despite only NimBLE being enabled in sdkconfig.

The problem does not occur if I build the demos, only if I build my own project, which consumes amazon-freertos externally as described in issue #832 

**System information**
- Which hardware board or part numbers? ESP32
- IDE used: none, CMake CLI build
- Operating System: Linux
- Version of Amazon FreeRTOS: master branch since commit 3949bfa
- Project: Custom Application 

**Expected behavior**
Linking succeeds

**Screenshots or console output**
Build output:
https://gist.github.com/matteoscordino/b0fbc27184f08a17a35bb3afc4b7db21

**To reproduce**
Build with the project below

**Code to reproduce the bug**
This is the CMakeList.txt:
https://gist.github.com/matteoscordino/5a1b0f984305f480f25ec88f8e187c17

Full project structure tree:
https://gist.github.com/matteoscordino/3bc62c94d6c93cfbbc836275a6a947e5

huguesBouvier:
Hello,

I looked at the cmake file but I am not seeing where "set(afr_ports_dir "${CMAKE_CURRENT_LIST_DIR}/ports")" gets included. 
I think the issue stems from this:
vendors\espressif\boards\esp32\ports\ble\nimble\iot_ble_hal_common_gap.c
Has the following code:
#if CONFIG_NIMBLE_ENABLED == 1
    #include "nimble/iot_ble_hal_common_gap.c"
#else
    #include "bluedroid/iot_ble_hal_common_gap.c"
#endif

So you need to include the nimble/iot_ble_hal_common_gap.c" and bluedroid/iot_ble_hal_common_gap.c as include files, not source files.

matteoscordino:
Hello @huguesBouvier ,

I added `set(afr_ports_dir "${CMAKE_CURRENT_LIST_DIR}/amazon-freertos/ports")` at the beginning of my `CMakeLists.txt` but it didn't make a difference. I was under the impression that it would be brought in anyway by dependencies of the root amazon-freertos `CMakeLists.txt`?

[...]

Issue #832, opened 2019-06-23
Document how to create a new cmake project outside the freertos source tree


jrr:
**Is your feature request related to a problem? Please describe.**  
The Getting Started Guides describe how to run the demo apps, but I can't find anything about creating a _new_ project.

**Describe the solution you would like.**  
Documentation describing how to create a new project outside the amazon-freertos source tree, and guidance about how to consume freertos from outside. 

**Additional context**  
With the old Make-based system, I was able to create a new ESP32 project that references amazon-freertos from outside, including this repo as an unmodified git submodule.

(This wasn't documented before, either, so it took significant effort!)

Now with the new cmake system I feel like I have to start all over.

**I'm looking for something like:**
1) Copy files X, Y, and Z into a new directory
2) Export an `AMAZON_FREERTOS_PATH` environment variable pointing the root directory of the amazon-freertos repo
3) Issue command `cmake -D...` to build your application image.

Is this documented anywhere? Am I missing something?

matteoscordino:
Agreed, this is a big gap to user adoption in the current state of affairs. 
Took me quite some work, as @jrr said, to get this (extremely common) usage scenario to work with make before, and now we have to start over.

tgsong:
Hi,

We haven't establish a standard way of consuming Amazon FreeRTOS outside of its source tree, and we're still evaluating several approaches for this. Generally speaking, with CMake you have 2 ways of doing this,

1. Add the external library (in your case, it's amazon freertos) as a subfolder in your project, and use CMake `add_subdirectory` command to consume it. The downside is that you have to compile external code every time.
1. Compile the external library and "install" its header files and build artifacts to a directory. Then in your CMake project, use the `find_package` command to load this dependency. This way you don't have to bring the external source code into your application, and you're just using header files and linking to static libraries. However, this needs to be supported by the external project as CMake does not magically provide this.

Normally option 2 is better than option 1 as it provides best separation of the application code and its dependencies. It's also the standard way of consuming an external library in C and C++. However, this is not implemented in Amazon FreeRTOS yet, our team spent most of the effort to support CMake build first on as many boards as possible in the last release.

Having said that, you can still use option 1 today with our CMake build support. Assuming you have the following directory structure,

```
- amazon-freertos
- main.c
- CMakeLists.txt
```

Here's a small CMake file you can try to start with,

```cmake
cmake_minimum_required(VERSION 3.13)
 
project(my_app)
 
# AFR_BOARD tells which board we need to target.
[...]

PR #1160, opened 2019-08-29
Fix cmake scope issue for source file properties

tgsong:
Description
-----------

This PR addresses issues in #832 and #1131,

- Use `include` for vendor CMake file instead of `add_subdirectory`.
- BLE hal layer is now defined as static library.

Some use cases require vendor to set some custom settings at source file level. Specifically in issue #1131, some source files in BLE hal porting layer needs to be treated as header files. However, CMake does not support passing these settings outside of current directory scope. `add_subdirectory` will create a new scope which makes this impossible. So we need to use `include` which do not create a new scope.

Test results are pending at this moment.

Checklist:
----------

- [ ] ~~My code is Linted.~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,11 +44,16 @@ elseif(DEFINED AFR_MANIFEST_BOARD_DIR)
 else()
     message(FATAL_ERROR "Could not import board CMakeLists.txt.")
 endif()
-add_subdirectory("${AFR_BOARD_PATH}")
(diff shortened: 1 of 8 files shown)


matteoscordino:
For what it's worth, I tested this PR and it solved issue #1131 for me

tjirsch:
I have tested the pr and it works for me. If you need another reviewer, please add me.

jnorgan:
This PR solved this issue for me as well. Please merge this when possible.

PR #1161, opened 2019-08-29
Allow demos to be turned off in CMake

tgsong:
Description
-----------

Addressing an issue from #832. Demos can now be turned off by setting `AFR_ENABLE_DEMOS` to 0 either from commandline or inside a CMake file.

Checklist:
----------

- [ ] ~~My code is Linted.~~


alexa-noxon:
Recording some of the discussion topics that came up from this PR

- how should we specify config file paths (for flexibility, and so that neither test nor demo configs need to be modified/included in projects consuming)?
- what arguments should be used to specify that demos should be turned off (keeping in mind backward compatibility issues and double-negatives in our cmake options :) )

qiutongs:
> Recording some of the discussion topics that came up from this PR
> 
>     * how should we specify config file paths (for flexibility, and so that neither test nor demo configs need to be modified/included in projects consuming)?
> 
>     * what arguments should be used to specify that demos should be turned off (keeping in mind backward compatibility issues and double-negatives in our cmake options :) )

I would suggest to merge this PR before everything can be finalized, as long as it is harmless to our current cmake build system. There are customers who want to start build their prototype. All they need initially is to compile and run it. We can state our current limitation: configuration files are under "demo" folder.

PR #1162, opened 2019-08-29
Add unit tests for mbox functions in lwIP

dan4thewin:
These are unit tests for the mbox functions from sys_arch.c.
They demonstrate a cpp-based approach to writing tests.
A write-up of the approach and these tests may be found at
https://git.io/fjxVo

### DIFF
--- a/libraries/3rdparty/lwip/src/portable/arch/unit-test/tap.h
+++ b/libraries/3rdparty/lwip/src/portable/arch/unit-test/tap.h
@@ -1,0 +1,46 @@
+/* minimal TAP: test anything protocol */
+/* https://testanything.org/ */
+
+#include <stdio.h>
+#include <stdlib.h>
+
+int _plan;      /* number of tests */
+int _ok_cnt;    /* number of passing tests */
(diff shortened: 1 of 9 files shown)


libraries/3rdparty/lwip/src/portable/arch/unit-test/tap.h
```diff
@@ -0,0 +1,46 @@
...
+int _plan;      /* number of tests */
+int _ok_cnt;    /* number of passing tests */
+int _index;     /* ordinal of current test */
+int _verbose;   /* flag: show all expressions */
+
+/* how many tests should run? */
+/* set verbose flag if environment var VERBOSE exists */
+#define plan(x)  do {                               \
+    _plan = (x);                                    \
+    if (_plan)                                      \
```
gkwicker: I'd prefer if these `if`,`else` blocks had `{`,`}` around the statements.  Same comment at 32-36 and 41/42.

dan4thewin:
I will reintroduce this at a later time.

PR #1168, opened 2019-08-30
Fix BLE testcase setup and teardown

danhorner:
Fix BLE testcase setup and teardown

Description
-----------
Miscellaneous fixes for testcases encountered during testing with ESP-IDF FreeRTOS. Check return values, initialize BLE as needed, and use some older APIs during assertions.

Checklist:
----------

- [ X] My code is Linted.

### DIFF
--- a/libraries/abstractions/platform/test/iot_test_platform_threads.c
+++ b/libraries/abstractions/platform/test/iot_test_platform_threads.c
@@ -119,8 +119,7 @@ TEST( UTIL_Platform_Threads, IotThreads_CreateDetachedThread )
 #if ( configUSE_TRACE_FACILITY == 1 )
     void threadPriorityTestFunction( void * param )
     {
-
-        *(int32_t*) param = uxTaskPriorityGet( NULL );
+        *( int32_t * ) param = uxTaskPriorityGet( NULL );
     }
 
     TEST( UTIL_Platform_Threads, IotThreads_ThreadPriority )
@@ -171,7 +170,6 @@ TEST( UTIL_Platform_Threads, IotThreads_CreateDetachedThread )
  */
     void threadStackSizeTestFunction( void * param )
     {
-
         *( int32_t * ) param = 300 + uxTaskGetStackHighWaterMark( NULL );
     }
 #endif /* if ( configUSE_TRACE_FACILITY == 1 ) */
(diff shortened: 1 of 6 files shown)


libraries/abstractions/platform/test/iot_test_platform_threads.c
```diff
@@ -119,23 +119,7 @@ TEST( UTIL_Platform_Threads, IotThreads_CreateDetachedThread )
...
-            &xTaskDetails,
-
-            /* Include the stack high water mark value in the
-             * TaskStatus_t structure. */
-            pdTRUE,
-            /* Include the task state in the TaskStatus_t structure. */
-            eInvalid );
-
-        *( int32_t * ) param = xTaskDetails.uxCurrentPriority;
+        *( int32_t * ) param = uxTaskPriorityGet( NULL );
```
ravibhagavandas: Will this be always enabled ? Checking freertos documentation: https://www.freertos.org/a00128.html INCLUDE_uxTaskPriorityGet  should be set to 1.
Previous code had a guard for configUSE_TRACE_FACILITY which was required to be enabled for `vTaskGetInfo()`

libraries/abstractions/platform/test/iot_test_platform_threads.c
```diff
@@ -186,23 +170,7 @@ TEST( UTIL_Platform_Threads, IotThreads_CreateDetachedThread )
...
-            &xTaskDetails,
-
-            /* Include the stack high water mark value in the
-             * TaskStatus_t structure. */
-            pdTRUE,
-            /* Include the task state in the TaskStatus_t structure. */
-            eInvalid );
-
-        *( int32_t * ) param = 300 + xTaskDetails.usStackHighWaterMark;
+        *( int32_t * ) param = 300 + uxTaskGetStackHighWaterMark( NULL );
```
ravibhagavandas: same as comment  `uxTaskPriorityGet()`

libraries/abstractions/platform/test/iot_test_platform_threads.c
danhorner: Thanks. Good catch. It's enabled in all of our test configurations, but I've gone ahead and updated the guard.

libraries/abstractions/platform/test/iot_test_platform_threads.c
danhorner: I went ahead and removed this function. It's unused.

danhorner:
Rebase to master and merge.

PR #1173, opened 2019-08-31
Fix integer overflow in HTTP _addHeaders buffer offset calculations.

markrtuttle:
Fix an integer overflow in pointer arithmetic.


markrtuttle:
The requests

* https://github.com/aws/amazon-freertos/pull/1173
* https://github.com/aws/amazon-freertos/pull/1174

are failing CBMC tests because both are required for all tests to pass.

PR #1174, opened 2019-08-31
Fix integer overflow in CBMC proofs caused by long buffer lengths.

markrtuttle:
Fix integer overflow in CBMC proofs caused by choosing buffer lengths longer than the 
maximum object size allowed by CBMC.

In CBMC, a pointer is an object id followed by an offset into the object.  The number of bits of the pointer used for the object id limits the number of bits remaining for the offset, and hence limits the size of the object.  If the top 7 bits are used for the object id, then offsets into the object must fit into the bottom 32-7 bits.


markrtuttle:
The requests

* https://github.com/aws/amazon-freertos/pull/1173
* https://github.com/aws/amazon-freertos/pull/1174

are failing CBMC tests because both are required for all tests to pass.

PR #1183, opened 2019-09-04
UDP support for Amazon FreeRTOS Sockets

ericyeh1995:
Description
-----------

UDP for afr sockets, available on boards with FreeRTOS+TCP and LWIP stack.
Enabling project build for Windows Simulator and Cypress CYW943907AEVAL1F. 
Support for other network stacks/boards are currently work in progress.

Checklist:
----------

- [x] My code is Linted.


libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+/*-----------------------------------------------------------*/
+
+IotNetworkError_t IotNetworkUdp_Close( void * pConnection )
+{
+    int32_t socketStatus = SOCKETS_ERROR_NONE;
+
+    /* Cast network connection to the correct type. */
+    _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
+
+    if( pNetworkConnection->socket != SOCKETS_INVALID_SOCKET )
```
jainamt: ```pConnection``` can be ```NULL```

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+    }
+    
+
+    return bytesReceived;
+}
+/*-----------------------------------------------------------*/
+
+IotNetworkError_t IotNetworkUdp_Close( void * pConnection )
+{
+    int32_t socketStatus = SOCKETS_ERROR_NONE;
```
jainamt: ```IOT_FUNCTION_ENTRY```

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+
+    /* Set the receive callback and context. */
+    pNetworkConnection->receiveCallback = receiveCallback;
+    pNetworkConnection->pReceiveContext = pContext;
+
+    /* No flags should be set. */
+    configASSERT( xEventGroupGetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ) ) == 0 );
+
+    /* Create task that waits for incoming data. */
+    if( xTaskCreate( _networkReceiveTask,
```
jainamt: what if ```IotNetworkUdp_SetReceiveCallback``` is called twice? Would it end up creating 2 tasks?

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+        {
+            IotLogWarn( "Failed to close connection." );
+        }
+
+        /* If receive task was started , indicate SHUT_DOWN */
+        if( pNetworkConnection->receiveCallback != NULL )
+        {
+            /* Set the shutdown flag. */
+            ( void ) xEventGroupSetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ),
+                                         _FLAG_SHUTDOWN );
```
jainamt: the order here is different than order in ```IoTNetworkUdp_Destroy```. In that function, the task is signaled and we wait for the task to exit, before we destroy the connection. Here we shutdown the socket, and then signal the task to shutdown. 

Why the difference?

libraries/abstractions/platform/freertos/include/platform/iot_network_udp.h
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @file iot_network_udp.h
+ * @brief Declares the network stack functions specified in platform/iot_network.h for
+ * Amazon FreeRTOS UDP Sockets.
+ */
+
+#ifndef _IOT_NETWORK_UDP_H_
+#define _IOT_NETWORK_UDP_H_
+
```
jainamt: Should all this be under a check for ```configPLATFORM_SOCKET_UDP_SUPPORT ```

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file iot_network_udp.c
+ * @brief Implementation of the network-related functions from iot_network_udp.h
+ * for Amazon FreeRTOS secure sockets.
+ */
+
```
jainamt: should all the code in this file be under a check for ```configPLATFORM_SOCKET_UDP_SUPPORT ```  ?

libraries/abstractions/secure_sockets/freertos_plus_tcp/iot_secure_sockets.c
```diff
@@ -496,8 +582,8 @@ Socket_t SOCKETS_Socket( int32_t lDomain,
 
     /* Ensure that only supported values are supplied. */
     configASSERT( lDomain == SOCKETS_AF_INET );
-    configASSERT( lType == SOCKETS_SOCK_STREAM );
-    configASSERT( lProtocol == SOCKETS_IPPROTO_TCP );
+    configASSERT( ( lType == SOCKETS_SOCK_STREAM ) || ( lType == SOCKETS_SOCK_DGRAM ) );
+    configASSERT( ( lProtocol == SOCKETS_IPPROTO_TCP ) || ( lProtocol == SOCKETS_IPPROTO_UDP ) );
```
jainamt: with the combination of values for ```configPLATFORM_SOCKET_UDP_SUPPORT ```, are these checks correct?

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+        /* If receive task was started , indicate SHUT_DOWN */
+        if( pNetworkConnection->receiveCallback != NULL )
+        {
+            /* Set the shutdown flag. */
+            ( void ) xEventGroupSetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ),
+                                         _FLAG_SHUTDOWN );
+        }
+    }
+
+    return IOT_NETWORK_SUCCESS;
```
leegeth: Does the api need to return success even if the call failed?

tests/include/aws_test_tcp.h
```diff
@@ -26,6 +26,16 @@
...
 
+/* UDP Echo Server. TODO update address
+ * Update udptestECHO_SERVER_ADDR# and
+ * udptestECHO_PORT with IP address
+ * and port of unencrypted TCP echo server. */
+#define udptestECHO_SERVER_ADDR0         10
+#define udptestECHO_SERVER_ADDR1         184
+#define udptestECHO_SERVER_ADDR2         166
+#define udptestECHO_SERVER_ADDR3         154
+#define udptestECHO_PORT                 ( 8883 )
```
leegeth: I think we will  have to either rename the file to indicate it is the config file for both tcp and udp or move these constants out to a new udp file

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+/*-----------------------------------------------------------*/
+
+size_t IotNetworkUdp_Send( void * pConnection,
+                           const uint8_t * pMessage,
+                           size_t messageLength )
+{
+    size_t bytesSent = 0;
+    _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
+
+    bytesSent = SOCKETS_SendTo( pNetworkConnection->socket,
```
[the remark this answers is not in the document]
ericyeh1995: return 0
https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/platform/include/platform/iot_network.h

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+                                                    IotNetworkReceiveCallback_t receiveCallback,
+                                                    void * pContext )
+{
+    IotNetworkError_t status = IOT_NETWORK_SUCCESS;
+
+    /* Cast network connection to the correct type. */
+    _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
+
+    /* Set the receive callback and context. */
+    pNetworkConnection->receiveCallback = receiveCallback;
```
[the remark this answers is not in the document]
ericyeh1995: Return IOT_NETWORK_BAD_PARAMETER and log error

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+{
+    size_t bytesReceived = 0, bytesRemaining = bytesRequested;
+    int32_t socketStatus = 0;
+    _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
+    SocketsSockaddr_t serverAddress = { 0 };
+    size_t addrLength = 0;
+
+
+    /* if Receive was called as a result of asynch callback, check if receive had failed.*/
+    if( ( xTaskGetCurrentTaskHandle() == pNetworkConnection->receiveTask ) && ( pNetworkConnection->bufferedByteValid == false ) )
```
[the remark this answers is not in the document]
ericyeh1995: log error and return 0

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    udpSocket = SOCKETS_Socket( SOCKETS_AF_INET,
+                                SOCKETS_SOCK_DGRAM,
+                                SOCKETS_IPPROTO_UDP );
+
+    if( udpSocket == SOCKETS_INVALID_SOCKET )
+    {
+        IotLogError( "Failed to create new socket." );
+        IOT_SET_AND_GOTO_CLEANUP( IOT_NETWORK_BAD_PARAMETER );
```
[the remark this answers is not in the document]
ericyeh1995: changed to IOT_NETWORK_SYSTEM_ERROR

tests/include/aws_test_tcp.h
ericyeh1995: yes

libraries/abstractions/platform/freertos/iot_network_udp.c
ericyeh1995: No, check is not needed since project will include the file if the its board is supported.

tests/include/aws_test_tcp.h
abhidixi11: Note: renaming may need bunch of project file updates for various platforms.

libraries/abstractions/platform/freertos/iot_network_udp.c
abhidixi11: Here the goal is to shutdown the socket in both direction and wait for recv task to exit .  Destroy call destroys the connection, hence receive task must exit (if it is running). UDP can only have destroy and does not need close(), but since API provides the call, it is intended  for 'two step' close. (A destroy call is still needed)

libraries/abstractions/platform/freertos/iot_network_udp.c
jainamt: But that means, that ```receiveCallback``` and ```pReceiveContext``` from 1st application will be overwritten by the call from 2nd application. How does this work when 2 applications are using this interface?

libraries/abstractions/platform/freertos/iot_network_udp.c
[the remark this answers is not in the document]
jainamt: Is this resolved in the next commit?

libraries/abstractions/platform/freertos/iot_network_udp.c
jainamt: "Here the goal is to shutdown the socket in both direction and wait for recv task to exit ." Thats exactly my point. I dont see the logic to wait for the task to exit, it seems we only set the bit and return from this function, without waiting to confirm that the task exited.

RichardBarry:
Not looked at this in great detail, but as an overall comment, I would expect UDP socket functions to be called sendto() and recvfrom(), and TCP socket functions to be called send() and recv(), and for their prototypes to be in the same header file as the are both sockets functions.

Additionally, when using FreeRTOS+TCP the TCP/IP stack is already running in a task, and if the callback mechanism is used (which is outside the scope of Berkeley sockets behavior) then they run in the context of that task.  Please don't create additional tasks.

libraries/abstractions/platform/freertos/iot_network_udp.c
RichardBarry: Please see my comments below regarding creating new tasks - in my opinion no tasks should be created at all if the FreeRTOS+TCP stack is used - can't comment for lwIP though.

abhidixi11: > Please see my comments below regarding creating new tasks - in my opinion no tasks should be created at all if the FreeRTOS+TCP stack is used - can't comment for lwIP though.

Network abstraction for TCP currently creates task for asynchronous receive if callback is set.  UDP abstraction code was written to be consistent with TCP network abstraction code:  https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/platform/freertos/iot_network_freertos.c#L458
Network abstraction is common for all TCP/IP stacks, it uses AFR secure sockets interface.

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+
+    /* No flags should be set. */
+    configASSERT( xEventGroupGetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ) ) == 0 );
+
+    /* Create task that waits for incoming data. */
+    if( xTaskCreate( _networkReceiveTask,
+                     "NetUdpRecv",
+                     IOT_NETWORK_RECEIVE_TASK_STACK_SIZE,
+                     pNetworkConnection,
+                     IOT_NETWORK_RECEIVE_TASK_PRIORITY + 5,
```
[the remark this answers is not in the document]
ericyeh1995: test passed after removing +5

libraries/abstractions/platform/freertos/iot_network_udp.c
abhidixi11: > But that means, that `receiveCallback` and `pReceiveContext` from 1st application will be overwritten by the call from 2nd application. How does this work when 2 applications are using this interface?

Each application will use different pConnection. IotNetworkUdp_Create() allocates new connection, so there no overwriting. If same application calls it twice (with same pConnection), it will be overwritten and that is intentional.

libraries/abstractions/platform/freertos/iot_network_udp.c
abhidixi11: IotNetworkUdp_Close() has to be followed by IoTNetworkUdp_Destroy(). _FLAG_SHUTDOWN will break the while loop in the recv task.  oTNetworkUdp_Destroy() will then make sure task has exited.

libraries/abstractions/secure_sockets/freertos_plus_tcp/iot_secure_sockets.c
abhidixi11: Good Point. Here if application developer uses UDP flags by mistake, the check will pass. But other udp functions like sendTo and recvFrom will not compile.

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,534 @@
...
+            bytesRemaining -= socketStatus;
+            bytesReceived += socketStatus;
+            configASSERT( bytesReceived + bytesRemaining == bytesRequested );
+        }
+    }
+
+    if( ( bytesReceived > 0 ) && ( pNetworkConnection->strictServerAddrCheck ) )
+    {
+        /* if strict server address check is enabled, only accept packets from the server that was used for Send */
+        if( serverAddress.ulAddress != pNetworkConnection->serverAddr.ulAddress )
```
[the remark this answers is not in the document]
ericyeh1995: while loop removed.

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,556 @@
...
+    int32_t socketStatus = SOCKETS_ERROR_NONE;
+
+    /* Cast network connection to the correct type. */
+    _networkConnection_t * pNetworkConnection = ( _networkConnection_t * ) pConnection;
+
+    if( pConnection == NULL )
+    {
+        IotLogWarn( "Invalid connection parameter" );
+    }
+    else if( pNetworkConnection->socket != SOCKETS_INVALID_SOCKET )
```
ericyeh1995: should this be pNetworkConnection->socket == SOCKETS_INVALID_SOCKET

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,544 @@
...
+        configASSERT( bytesReceived == bytesRequested );
+    }
+
+    if( ( bytesReceived > 0 ) && ( pNetworkConnection->strictServerAddrCheck ) )
+    {
+        /* if strict server address check is enabled, only accept packets from the server that was used for Send */
+        if( serverAddress.ulAddress != pNetworkConnection->serverAddr.ulAddress )
+        {
+            IotLogError( "Server address does not match and strict server address check enabled." );
+            bytesReceived = 0;
```
jainamt: Although ```bytesReceived``` is set to ```0```, but ```pBuffer``` has the data in it. Should it be cleared off as well?

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,544 @@
...
+
+    return bytesSent;
+}
+/*-----------------------------------------------------------*/
+
+size_t IotNetworkUdp_ReceiveFrom( void * pConnection,
+                              uint8_t * pBuffer,
+                              size_t bytesRequested )
+{
+    size_t bytesReceived = 0;
```
jainamt: This is not needed after the ```while``` loop is removed. I think ```socketStatus``` has all the info thats needed.

libraries/abstractions/platform/freertos/iot_network_udp.c
ericyeh1995: After calling IotNetworkUdp_Close() the receive task may not exit, to ensure receive task is exited we call IotNetworkUdp_Destroy() to prevent any further connection.

libraries/abstractions/platform/freertos/iot_network_udp.c
ericyeh1995: Each ( _networkConnection_t * ) pConnection has a delicate socket and send/recv task. Additional call for this function with different pConnection will be on different socket and have new recv task.

tests/include/aws_test_tcp.h
ericyeh1995: Adding a new file also needs additional update, but is not as a big amount of work all at once: we should decide while still in development.

PR #1184, opened 2019-09-04
AFR: NimBLE - Add support to disable and deinit NimBLE

dhrishi:
AFR: NimBLE - Add support to disable and deinit NimBLE

Description
-----------
1. Implement .pxDisable for NimBLE

2. Also, remove `esp_bt_controller_mem_release( ESP_BT_MODE_BTDM )` from
pxBtManagerCleanup as it is not required. That's because, controller
memory for classic BT is already freed in case of NimBLE. So, we just
need to free controller memory for BLE in case of cleanup

Checklist:
----------

- [ ] My code is Linted.


dhrishi:
@huguesBouvier Please review

PR #1203, opened 2019-09-10
AFR NimBLE: Fix pairing request GAP event handling for SC only mode

prasad-alatkar:
- Additional check added to confirm the `max_enc_key_size` in pairing request is not less than 16
  Bytes in case of secure connection only mode.

Description
-----------

1. Modifies `iot_ble_hal_common_gap.c` to reject pairing request if the max encryption key size is less than 16 Bytes with reason `BLE_SM_ERR_ENC_KEY_SZ`.
2. Adds `max_enc_key_size` to structure `ble_gap_pairing_req`, so that can be used while handling GAP pairing request event.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_common_gap.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_common_gap.c
@@ -246,25 +246,39 @@ int prvGAPeventHandler( struct ble_gap_event * event,
                                              0 );
             }
 
-            if( ble_hs_cfg.sm_sc && ( event->pairing_req.authreq & BLE_SM_PAIR_AUTHREQ_SC ) &&
-                ( ble_sm_sc_ioa[ ble_hs_cfg.sm_io_cap ][ event->pairing_req.io_cap ] == BLE_SM_IOACT_NONE ) )
+            if( ble_hs_cfg.sm_sc && ( event->pairing_req.authreq & BLE_SM_PAIR_AUTHREQ_SC ))
             {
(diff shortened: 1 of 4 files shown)


vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_common_gap.c
```diff
@@ -246,26 +246,40 @@ int prvGAPeventHandler( struct ble_gap_event * event,
...
+                        xBTCallbacks.pxPairingStateChangedCb( eBTStatusFail, ( BTBdaddr_t * ) desc.peer_id_addr.val,
+                                                              eBTbondStateNone,
+                                                              eBTSecLevelNoSecurity,
+                                                              eBTauthFailInsuffSecurity );
+                    }
+
+                    return BLE_SM_ERR_AUTHREQ;
+
                 }
+                else if (event->pairing_req.max_enc_key_size < BLE_SM_PAIR_KEY_SZ_MAX)
```
ravibhagavandas: The AFQP tests for pairing failed. After debugging the issue, I think max_enc_key_size parameter is not set within the event [here]( https://github.com/prasad-alatkar/amazon-freertos/blob/ble/NimBLE_fix_pairing_req_sc_only/vendors/espressif/esp-idf/components/nimble/nimble/nimble/host/src/ble_gap.c#L4504)

After adding the line `event.pairing_req.max_enc_key_size = req->max_enc_key_size;` , the test passes.
Can you fix the same ?

PR #1205, opened 2019-09-10
PKCS #11 private key import capability check

dcgaws:
Check consistency of static and runtime support for private key generation and import. This has been tested on the SoftHSM port as well.

### DIFF
--- a/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
+++ b/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
@@ -758,31 +758,31 @@ TEST( Full_PKCS11_Capabilities, AFQP_Capabilities )
     TEST_ASSERT_EQUAL_MESSAGE( CKR_OK, xResult, "Failed to get slot count" );
 
     /* Check for RSA PKCS #1 signing support. */
-    xResult = pxGlobalFunctionList->C_GetMechanismInfo( pxSlotId[0], CKM_RSA_PKCS, &MechanismInfo );
-    TEST_ASSERT_TRUE(CKR_OK == xResult || CKR_MECHANISM_INVALID == xResult);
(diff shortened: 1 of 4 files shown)


libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -814,6 +816,26 @@ TEST( Full_PKCS11_Capabilities, AFQP_Capabilities )
...
+     * generation settings. */
+    if( CK_TRUE == xSupportsKeyGen )
+    {
+        #ifndef pkcs11testGENERATE_KEYPAIR_SUPPORT
+            TEST_FAIL_MESSAGE( "Static and runtime configuration for key generation support are inconsistent." );
+        #endif
+    }
+    else
+    {
+        #ifdef pkcs11testGENERATE_KEYPAIR_SUPPORT
```
sarenameas: This is the same as the if block. Is this intended?

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -814,6 +816,26 @@ TEST( Full_PKCS11_Capabilities, AFQP_Capabilities )
     xResult = pxGlobalFunctionList->C_GetMechanismInfo( pxSlotId[ 0 ], CKM_SHA256, &MechanismInfo );
     TEST_ASSERT_TRUE( CKR_OK == xResult );
     TEST_ASSERT_TRUE( 0 != ( CKF_DIGEST & MechanismInfo.flags ) );
+
+    /* Check for consistency between static configuration and runtime key
```
alexa-noxon: This is a good idea. 

Could we also do checks on pkcs11testRSA_KEY_SUPPORT and pkcs11testEC_KEY_SUPPORT?

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
dcgaws: Yes, the intent is to check whether they're conflicting.

PR #1208, opened 2019-09-11
Generate individual proof makefiles instead of using Makefile.common

karkhaz:
This PR makes the build system generate a complete custom Makefile for every CBMC proof, rather than using a common Makefile for all proofs.

Our proofs have previously been controlled through a common list of recipes written in `Makefile.common`, while proof-specific definitions have been written to a minimal `Makefile` in the proof's directory that includes `Makefile.common`. This is no longer sufficient, since the recipes needed for various proofs are becoming more and more divergent from the common Makefile. We cannot deal with this divergence using conditional constructs in the Makefile, as we need to support both Windows and Linux Makefiles, and the conditional syntax differs incompatibly between GNU Make and NMAKE.

This PR therefore does all conditional evaluation in Python, writing one self-contained Makefile for each proof.

**Note to reviewers**. The changes to all the `Makefile.json`s were done semi-automatically, by running the following script and then re-inserting the comments in the correct places by hand by comparing the diff. So you can ignore these files in the review, unless you'd like to run the script yourself. I would advise using the "file filter" menu to avoid reading JSON files.
```python3
#!/usr/bin/env python3

import collections
import json
import logging
import os
import re

for root, _, fyles in os.walk("."):
    if "Makefile.json" in fyles:
        path = os.path.join(root, "Makefile.json")
        lines = []
        with open(path) as handle:
            for line in handle:
                if not line.strip().startswith("#"):
                    lines.append(line)
        data = json.loads(" ".join(lines),
                          # Preserve order of keys, since later ones
                          # might refer to earlier ones
                          object_pairs_hook=collections.OrderedDict)
        new_data = {
            "variables": collections.OrderedDict(),
            "lists": collections.OrderedDict(),
            "makefile-control": collections.OrderedDict(),
        }

        for k, v in data.items():
            # We're not using this any more
            if k == "EXPECTED":
                continue
            if k in ["INC", "DEF", "OBJS", "CBMCFLAGS", "INSTFLAGS"]:
                new_data["lists"][k] = v
            else:
                new_data["variables"][k] = v

        # Sanity check that when we dump the JSON and reload it, all
        # variables are defined before they are used–i.e. that the
        # reshuffling of variables really does preserve their original
        # order. Of course, this will also complain if their original
        # order was incorrect...
        blob = json.dumps(new_data)
        new_data = json.loads(blob, object_pairs_hook=collections.OrderedDict)

        seen = {}
        ok = True
        for k, v in new_data["variables"].items():
            try:
                evaluated = str(v).format(**seen)
            except KeyError:
                ok = False
                logging.error("File '%s' referenced a variable in value "
                              "'%s' for key '%s', but the variable was "
                              "not previously declared", path, v, k)
            else:
                seen[k] = evaluated

        if not ok:
            exit(1)

        with open(path, "w") as handle:
            print(json.dumps(new_data, indent=2), file=handle)
```


tools/cbmc/proofs/settings.json
```diff
@@ -0,0 +1,58 @@
...
+      "__free_rtos__",
+      "_CONSOLE",
+      "_WIN32_WINNT=0x0500",
+      "WINVER=0x400",
+      "_CRT_SECURE_NO_WARNINGS",
+      "__PRETTY_FUNCTION__=__FUNCTION__",
+      "'configASSERT(X)=__CPROVER_assert(X,\"Assertion Error\")'",
+      "'configPRECONDITION(X)=__CPROVER_assume(X)'",
+      "CBMC_MAX_OBJECT_SIZE={CBMC_MAX_OBJECT_SIZE}",
+      "CBMC_OBJECTID_BITS={CBMC_OBJECTID_BITS}"
```
markrtuttle: I think -m32 is missing.

tools/cbmc/proofs/proof_specific_makefile.py
```diff
@@ -376,16 +425,32 @@ def set_up_logging(args):
...
+            # Deal with the case of a harness being nested several levels under
+            # the top-level proof directory, where the common Makefile lives
+            common_dir_path = "..%s" % _platform_choices[args.system]["path-sep"]
+            common_dir_path = common_dir_path * len(root.split(os.path.sep)[1:])
+
+            with open(os.path.join(root, "Makefile"), "w") as handle:
+                handle.write("""\
+{contents}
+
+{include} {common_dir_path}Makefile.common""".format(
```
markrtuttle: Why does this makefile refer to makefile.common?  I thought the whole point was to generate just a single makefile.

tools/cbmc/proofs/settings.json
```diff
@@ -0,0 +1,58 @@
+{
+  "variables": {
+    "UNWIND_LIMIT": 1,
+    "FREERTOS": "../../..",
+    "PROOFS": ".",
+    "CBMC_MAX_OBJECT_SIZE": "\"(UINT32_MAX>>CBMC_OBJECTID_BITS)\"",
+    "CBMC_OBJECTID_BITS": 7,
```
markrtuttle: We should default to a smaller number like 4, since 32 bits gives very little room for offsets.

We should use better names.  How about
CBMC_OBJECT_BITS
CBMC_OBJECT_MAX_SIZE

tools/cbmc/proofs/Makefile.template
```diff
@@ -93,15 +58,20 @@ $(ENTRY)3.goto: $(ENTRY)2.goto
 $(ENTRY)4.goto: $(ENTRY)3.goto
 	$(GOTO_INSTRUMENT) $(INSTFLAGS) --slice-global-inits @RULE_INPUT@ @RULE_OUTPUT@ \
 		> $(ENTRY)4.txt 2>&1
+
+$(ENTRY)5.goto: $(ENTRY)4.goto
+	{nondet_static} \
+	  > $(ENTRY)5.txt 2>&1
```
markrtuttle: How do I define the exceptions to nondet-static?

karkhaz:
Closing, this has been superceeded by a CMake build system proposal

PR #1224, opened 2019-09-16
Fix coding style for HTTPS Client Sync and Async demos

sarenameas:
A post commit to master code review finds the "p" missing from char arrays and char pointer variables, in the HTTPS Client synchronous and asynchronous download demos.

This PR also elaborates on the difference between the path in IotHttpsRequest_t and returned from IotHttpsClient_GetUrlPath().
IotHttpsClient_GetUrlPath() returns the path length without a query.
The path with a query is defined to be the \"abs_path\" according to the HTTP/1.1 RFC and this is the path needed in IotHttpsRequest_t. This path is the one in the Request-Line.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/https/src/iot_https_client.c
+++ b/libraries/c_sdk/standard/https/src/iot_https_client.c
@@ -3294,7 +3294,7 @@ IotHttpsReturnCode_t IotHttpsClient_ReadHeader( IotHttpsResponseHandle_t respHan
     }
     else
     {
-        IotLogError( "IotHttpsClient_ReadHeader(): The header field %s was not found.", pName );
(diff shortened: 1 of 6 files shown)


libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -3294,7 +3294,7 @@ IotHttpsReturnCode_t IotHttpsClient_ReadHeader( IotHttpsResponseHandle_t respHan
     }
     else
     {
-        IotLogError( "IotHttpsClient_ReadHeader(): The header field %s was not found.", pName );
+        IotLogWarn( "IotHttpsClient_ReadHeader(): The header field %s was not found.", pName );
```
sarenameas: This is changed to warning for ease in debugging.

PR #1230, opened 2019-09-17
UDP support for Amazon FreeRTOS Sockets on ST

ericyeh1995:
Description
-----------

Porting on top of Inventek Systems Wi-Fi Module, some AFR Sockets UDP specification is not supported by default. The following explain how UDP AFR Sockets is implemented using the Inventek Wi-Fi API.

0. Documented Inventek Wi-Fi API UDP setup:
```
P0=x        (Set communication socket)
P1=1        (Set protocol to UDP)
P3=x.x.x.x  (Set remote server IP address)
P4=x        (Set remote UDP port number)
P6=1        (Enable UDP client mode)
```
The Inventek Wi-Fi API uses the above command in ES_WIFI_StartClientConnection. SendTo and ReceiveFrom require such connection enabled to work.

1. SOCKETS_Bind
Bind uses a modified version of ES_WIFI_StartClientConnection. The modified portion disregard setting P4 if port 0 is provided, this will allow system to self assign a port to be used. 

2. SOCKETS_SendTo and SOCKETS_RecvFrom
These two function require connection (ES_WIFI_StartClientConnection) succeed to work. Therefore, each function has a check whether or not connection had been made, if not, it will create a connection. Checking the connection may cause functions to behave differently from the standard. 

Tests to be quarantined:
Peek is not supported in Inventek Wi-Fi API nor the AT command set.

2. Threadsafe tests
Should be investigated since AFQP_SOCKETS_Threadsafe_SameSocketDifferentTasks does not work for secure sockets TCP as well. 

Other Notes:
Modification to the Inventek Wi-Fi API to allow setting remote address for UDP connection. A dummy IP is used to allow successful connection. 

Checklist:
----------

- [x] My code is Linted.


vendors/st/boards/stm32l475_discovery/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -148,6 +156,9 @@ typedef struct STSecureSocket
     void * pvTLSContext;                /**< The TLS Context. */
     char * pcServerCertificate;         /**< Server certificate. Set using SOCKETS_SO_TRUSTED_SERVER_CERTIFICATE option in SOCKETS_SetSockOpt function. */
     uint32_t ulServerCertificateLength; /**< Length of the server certificate. */
+    #if ( configPLATFORM_SOCKET_UDP_SUPPORT == 1 )
+        uint16_t usLocalPort;
```
sarenameas: It is great that it is documented in the PR that ST does not support bind, but can we write that in the comments too?

leegeth: I also think that will be a good idea as it will be easier to find the comments in the code while reading the code rather than finding the PR documentation

vendors/st/boards/stm32l475_discovery/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -564,6 +704,9 @@ Socket_t SOCKETS_Socket( int32_t lDomain,
         xSockets[ ulSocketNumber ].pvTLSContext = NULL;
         xSockets[ ulSocketNumber ].pcServerCertificate = NULL;
         xSockets[ ulSocketNumber ].ulServerCertificateLength = 0;
+        #if ( configPLATFORM_SOCKET_UDP_SUPPORT == 1 )
+            xSockets[ ulSocketNumber ].usLocalPort = -1;
```
leegeth: usLocalPort is unsigned int,  this will be a large positive number and also could be a valid port address.

ericyeh1995: Good catch, this was set to 0 originally but test binds to port 0 and doing so will result in binding to the same port. But now I have a solution in mind that should disregard the default value.

vendors/st/boards/stm32l475_discovery/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -862,6 +1005,180 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+                                                       ( uint16_t ) xDataLength,
+                                                       &( usReceivedBytes ),
+                                                       pxSocket->ulReceiveTimeout,
+                                                       ( uint8_t * ) xWiFiConnection.RemoteIP,
+                                                       ( uint16_t ) xWiFiConnection.RemotePort );
+
+                pxAddress->usPort = SOCKETS_ntohs( xWiFiConnection.RemotePort ); /* WiFi Module expects the port number in host byte order. */
+                memcpy( &( pxAddress->ulAddress ),
+                        &( xWiFiConnection.RemoteIP ),
+                        4 );
```
ericyeh1995: Use sizeof() for consistency.

vendors/st/boards/stm32l475_discovery/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -862,6 +1005,180 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+                if( ( xWiFiResult == ES_WIFI_STATUS_OK ) && ( usReceivedBytes != 0 ) )
+                {
+                    /* Success, return the number of bytes received. */
+                    xRetVal = ( BaseType_t ) usReceivedBytes;
+                }
+            }
+            else
+            {
+                /* The socket has been closed for read. */
+                lSentBytes = SOCKETS_ECLOSED;
```
ericyeh1995: should be xRetVal = SOCKETS_ECLOSED;

ericyeh1995:
Observation after some experimentation on the ES WiFi API

1. Setting local port (P2=) in start client connection does not effect UDP connection.
2. Setting remote port (P4=) in start client connection will actually be the local port it sends data from
3. System will return error if address is not given (no P3) or if address is invalid (ex: P3=0.0.0.0)
4. System will automatically assign port if it is not given (no P4)
5. System will return error if trying to bind (P4=) to an already bound port or if trying to bind to port 0
6. SendTo and RecvFrom requires successful connection to be made

vendors/st/stm32l475_discovery/BSP/Components/es_wifi/es_wifi.c
```diff
@@ -1580,17 +1578,28 @@ ES_WIFI_Status_t ES_WIFI_StartClientConnection(ES_WIFIObject_t *Obj, ES_WIFI_Con
 
   if (ret == ES_WIFI_STATUS_OK)
   {
-    sprintf((char*)Obj->CmdData,"P2=%d\r", conn->LocalPort);
+    sprintf( ( char * ) Obj->CmdData, "P2=%d\r", /*LocalPort*/ 56830 ); /* WARN: Does not work! */
```
ericyeh1995: This may work for TCP, revert to previous version.

vendors/st/boards/stm32l475_discovery/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -148,6 +148,9 @@ typedef struct STSecureSocket
     void * pvTLSContext;                /**< The TLS Context. */
     char * pcServerCertificate;         /**< Server certificate. Set using SOCKETS_SO_TRUSTED_SERVER_CERTIFICATE option in SOCKETS_SetSockOpt function. */
     uint32_t ulServerCertificateLength; /**< Length of the server certificate. */
+    #if ( configPLATFORM_SOCKET_UDP_SUPPORT == 1 )
+        uint16_t usLocalPort;
```
ericyeh1995: no longer necessary

PR #1231, opened 2019-09-17
add detailed API document to i2c

qiutongs:
Description
-----------

- add detailed API document to i2c
- format

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/include/iot_i2c.h
+++ b/libraries/abstractions/common_io/include/iot_i2c.h
@@ -42,32 +42,32 @@
 /**
  * The speeds supported by I2C bus.
  */
-#define IOT_I2C_STANDARD_MODE_BPS         ( 100000 )
-#define IOT_I2C_FAST_MODE_BPS             ( 400000 )
-#define IOT_I2C_FAST_MODE_PLUS_BPS        ( 1000000 )
-#define IOT_I2C_HIGH_SPEED_BPS            ( 3400000 )
+#define IOT_I2C_STANDARD_MODE_BPS         ( 100000 )  /*!< Standard mode bits per second. */
+#define IOT_I2C_FAST_MODE_BPS             ( 400000 )  /*!< Fast mode bits per second. */
+#define IOT_I2C_FAST_MODE_PLUS_BPS        ( 1000000 ) /*!< Fast plus mode bits per second. */
+#define IOT_I2C_HIGH_SPEED_BPS            ( 3400000 ) /*!< High speed mode bits per second. */
 
 /**
  * The return codes for the functions in I2C.
  */
-#define IOT_I2C_SUCCESS                       ( 0 )
-#define IOT_I2C_INVALID_VALUE                 ( 1 )
-#define IOT_I2C_BUSY                          ( 2 )
-#define IOT_I2C_WRITE_FAILED                  ( 3 )
-#define IOT_I2C_READ_FAILED                   ( 4 )
-#define IOT_I2C_NACK                          ( 5 )
-#define IOT_I2C_BUS_TIMEOUT                   ( 6 )
-#define IOT_I2C_NOTHING_TO_CANCEL             ( 7 )
-#define IOT_I2C_FUNCTION_NOT_SUPPORTED        ( 8 )
-#define IOT_I2C_SLAVE_ADDRESS_NOT_SET         ( 9 )
+#define IOT_I2C_SUCCESS                   ( 0 )     /*!< I2C operation completed successfully. */
+#define IOT_I2C_INVALID_VALUE             ( 1 )     /*!< At least one parameter is invalid. */
+#define IOT_I2C_BUSY                      ( 2 )     /*!< I2C bus is busy at current time. */
+#define IOT_I2C_WRITE_FAILED              ( 3 )     /*!< I2C driver returns error when performing write operation. */
+#define IOT_I2C_READ_FAILED               ( 4 )     /*!< I2C driver returns error when performing read operation. */
+#define IOT_I2C_NACK                      ( 5 )     /*!< Unexpected NACK is caught. */
+#define IOT_I2C_BUS_TIMEOUT               ( 6 )     /*!< I2C operation not completed within specified timeout. */
+#define IOT_I2C_NOTHING_TO_CANCEL         ( 7 )     /*!< No ongoing transaction when cancel operation is performed. */
+#define IOT_I2C_FUNCTION_NOT_SUPPORTED    ( 8 )     /*!< I2C operation is not supported. */
+#define IOT_I2C_SLAVE_ADDRESS_NOT_SET     ( 9 )     /*!< Slave address is not set before calling I2C read or write operation. */
 
 /**
  * @brief I2C Bus status
  */
 typedef enum
 {
-    eI2CBusIdle = 0,
-    eI2cBusBusy = IOT_I2C_BUSY,
+    eI2CBusIdle = 0,            /*!< I2C bus is idle. */
+    eI2cBusBusy = IOT_I2C_BUSY, /*!< I2C bus is busy. */
 } IotI2CBusStatus_t;
 
 /**
@@ -75,10 +75,10 @@ typedef enum
  */
 typedef enum
 {
-    eI2CCompleted = IOT_I2C_SUCCESS,
-    eI2CDriverFailed,
-    eI2CNackFromSlave = IOT_I2C_NACK,
-    eI2CMasterTimeout = IOT_I2C_BUS_TIMEOUT,
+    eI2CCompleted = IOT_I2C_SUCCESS,         /*!< I2C operation completed successfully. */
+    eI2CDriverFailed,                        /*!< I2C driver returns error during last operation. */
+    eI2CNackFromSlave = IOT_I2C_NACK,        /*!< Unexpected NACK is caught. */
+    eI2CMasterTimeout = IOT_I2C_BUS_TIMEOUT, /*!< I2C operation not completed within specified timeout. */
 } IotI2COperationStatus_t;
 
 /**
@@ -86,8 +86,8 @@ typedef enum
  */
 typedef struct IotI2CConfig
 {
-    uint32_t ulMasterTimeout;   /**<! Master timeout value in msec, to relinquish the bus if slave does not respond. This is an extended feature of SMBus which is not part of I2C protocol. */
-    uint32_t ulBusFreq;         /**<! Bus frequency/baud rate */
+    uint32_t ulMasterTimeout; /**<! Master timeout value in msec, to relinquish the bus if slave does not respond. This is an extended feature of SMBus which is not part of I2C protocol. */
+    uint32_t ulBusFreq;       /**<! Bus frequency/baud rate */
 } IotI2CConfig_t;
 
 /**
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -42,70 +42,69 @@
...
  * @brief Ioctl request types.
  */
 typedef enum
 {
-    eI2CSendNoStopFlag,     /**<! Set flag to not send stop after transaction */
-                            /**<! Default is always stop for every transaction */
-                            /**<! Flag will auto reset to stop after one transaction if you set no stop */
-    eI2CSetSlaveAddr,       /**<! All the operations use this slave address after it is set. */
-    eI2CSetMasterConfig,    /**<! Sets the I2C bus frequency and timeout using the struct IotI2CConfig_t,
-                             *   default speed is Standard */
```
lundinc2: Why remove this? Isn't this statement true?

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -130,7 +129,8 @@ typedef struct IotI2CDescriptor   * IotI2CHandle_t;
  *                          passed back to the caller in the callback.
  *
  */
```
lundinc2: Same comment here

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: Not remove, just made it one line...

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: I didn't see anything is removed here...

libraries/abstractions/common_io/include/iot_i2c.h
lundinc2: Oh my bad

lundinc2: Oh I realize what happened here. My git plugin is not displaying this PR properly

libraries/abstractions/common_io/include/iot_i2c.h
lundinc2: sorry problem with my PR plugin

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -178,29 +179,30 @@ void iot_i2c_set_callback( IotI2CHandle_t const pxI2CPeripheral,
  * @param[in] xBytes The number of bytes to read.
  *
  * @return
- * - IOT_I2C_SUCCESS, on success and all the requested bytes have been read
```
lundinc2: This is what I meant to comment on, is it necessary to remove this line?

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -215,52 +217,102 @@ int32_t iot_i2c_read_sync( IotI2CHandle_t const pxI2CPeripheral,
  * @param[in] xBytes The number of bytes to write.
  *
  * @return
- * - IOT_I2C_SUCCESS, on success and all the requested bytes have been written
```
lundinc2: same comment

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: Yeah, I removed because I think partial read could happen.

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: same as above

libraries/abstractions/common_io/include/iot_i2c.h
lundinc2: hmm I assumed it would return Timeout if all the request bytes weren't sent

qiutongs: Oh, you are right. Let me add those comments back.

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -42,70 +42,69 @@
...
-                             *   default speed is Standard */
-    eI2CGetMasterConfig,    /**<! Gets the I2C bus frequency and timeout set for the I2C master. */
-    eI2CGetBusState,        /**<! Get the current I2C bus status. Returns eI2CBusIdle or eI2CBusy */
-    eI2CBusReset,           /**<! Master resets the bus. */
-    eI2CGetTxNoOfbytes,     /**<! Get the number of bytes sent in write operation. */
-    eI2CGetRxNoOfbytes,     /**<! Get the number of bytes received in read operation. */
+    eI2CSendNoStopFlag,  /**<! Set flag to not send stop after transaction */
+                         /**<! Default is always stop for every transaction */
+                         /**<! Flag will auto reset to stop after one transaction if you set no stop */
+    eI2CSetSlaveAddr,    /**<! All the operations use this slave address after it is set. */
```
brianfr-amazon: Do we need to add anything about the expectations of the address?  7-bit address vs. 8-bit address vs. 10-bit address?

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -42,70 +42,69 @@
...
+    eI2CSetMasterConfig, /**<! Sets the I2C bus frequency and timeout using the struct IotI2CConfig_t, default speed is Standard mode. */
+    eI2CGetMasterConfig, /**<! Gets the I2C bus frequency and timeout set for the I2C master. */
+    eI2CGetBusState,     /**<! Get the current I2C bus status. Returns eI2CBusIdle or eI2CBusy */
+    eI2CBusReset,        /**<! Master resets the bus. */
+    eI2CGetTxNoOfbytes,  /**<! Get the number of bytes sent in write operation. */
+    eI2CGetRxNoOfbytes,  /**<! Get the number of bytes received in read operation. */
```
brianfr-amazon: Q:  What is the expectation on these two Get operations?  If I do a write, and then a read, is the expectation that GetTxNoOfbytes will still return the # of bytes from the write, or is the expectation that it is reset to zero at the beginning of each transaction?

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -118,7 +117,7 @@ struct                    IotI2CDescriptor;
  *        This is initialized in open and returned to caller. The caller must pass
  *        this pointer to the rest of APIs.
  */
-typedef struct IotI2CDescriptor   * IotI2CHandle_t;
+typedef struct IotI2CDescriptor * IotI2CHandle_t;
```
brianfr-amazon: Where is IotI2CDescriptor defined?  Where do we tell implementors what values it must/should have?

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -150,6 +150,7 @@ IotI2CHandle_t iot_i2c_open( int32_t lI2CInstance );
 /**
  * @brief Sets the application callback to be called on completion of an operation.
  *
+ * @note This is only used in asynchronous call.
```
brianfr-amazon: I think we need to indicate that only 1 callback is supported.  So, if this is called twice, the previous callback function will be over-written.  Separate callback functions are read_async and write_async are not supported.

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -215,52 +217,102 @@ int32_t iot_i2c_read_sync( IotI2CHandle_t const pxI2CPeripheral,
...
+ * @brief Starts the I2C master read operation in asynchronous mode.
+ *
+ * This function attempts to read certain number of bytes from a pre-allocated buffer, in asynchronous way.
+ * It returns immediately when the operation is started and the status can be check by calling iot_i2c_ioctl.
+ * Once the operation completes, successful or not, the user callback will be invoked.
+ *
+ * Partial read might happen, e.g. slave device unable to receive more data.
+ * And the number of bytes that have been actually read can be obtained by calling iot_i2c_ioctl.
+ *
+ * @Note Usually, the address of register needs to be written before calling this function.
```
brianfr-amazon: Suggest adding comment about usually calling iot_i2c_set_callback() prior to this, so that implementor can get notified when the async call is complete.

libraries/abstractions/common_io/include/iot_i2c.h
```diff
@@ -130,7 +129,8 @@ typedef struct IotI2CDescriptor   * IotI2CHandle_t;
  *                          passed back to the caller in the callback.
  *
  */
-typedef void (* IotI2CCallback_t) ( IotI2COperationStatus_t xOpStatus, void * pvUserContext );
+typedef void (* IotI2CCallback_t) ( IotI2COperationStatus_t xOpStatus,
+                                    void * pvUserContext );
 
 /**
```
brianfr-amazon: I feel like we should add a comment here that the expectation is that there will be one instance for each I2C device.  It is technically possible that someone could have one instance to support multiple devices on the same bus, as long as they changed/set the slave address between each operation.  But, this would require a higher level to synchronize the operations.

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: Not sure what is 8-bit address. But I think both 7-bit and 10-bit should work as long as the driver supports.

Correspondingly, users need to give a uint8_t and uint16_t. Does that make sense to you?

I will clarify in next PR.

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: I think GetTxNoOfbytes or eI2CGetRxNoOfbytes will return the value in "last transaction".
For example,
- write -> no stop -> read: GetTx returns # of bytes in write and GetRx returns # of bytes in read
- write -> stop: GetTx returns # of bytes in write
- read -> stop: GetRx returns # of bytes in read

Other cases we can claim is undefined?

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: This anonymous struct is defined in iot_i2c.c, so board specific. Are you saying we should guide vendors how to implement iot_i2c.c?

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: Good point. I will clarify

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: Yes, will do.

libraries/abstractions/common_io/include/iot_i2c.h
brianfr-amazon: Agreed.  My point is that we should document, and in your 3rd case of read->stop:  GetTx should return 0.  Since there were no Tx bytes.

libraries/abstractions/common_io/include/iot_i2c.h
brianfr-amazon: No, but I think we should guide them on what should go in the IotI2CDescriptor.  We can clarify that this is vendor/driver/component specific, but that it should probably include ...

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: I see. So it means if users wants to know the # of bytes, they must call ioctl immediately after read/write, before next read/write happen. Is that what you think?

If there were a chance to re-design the interface, I would like read_sync and write_sync to return # of bytes (negative value is error).  Thus there will be fewer states in the library, making the behavior clearer.

libraries/abstractions/common_io/include/iot_i2c.h
qiutongs: ok, I see. This information is only useful to vendor, not users. So I will put it to porting guide.

Issue #1119, opened 2019-08-20
Unable to import JITP Certificate in Flash


bgintz-rsc:
**Describe the bug**
Calling xProvisionCertificate in was_dev_mode_key_provisioning fails

**System information**
- ESP32
- Eclipse
- AFR 2019-6.00

**Expected behavior**
This function should securely store the JITP certificate in FLASH for use by the TLS connect.  It fails with a panic in the nvs blob storage due to a null filename.

** Cause and Proposed Fix **
This is caused by failure to handle the PKCS #11 label for Just-In-Time-Provisioning (pkcs11configLABEL_JITP_CERTIFICATE) in the switch statement in  prvLabelToFilenameHandle() in iot_pkcs11_pal.c which returns eInvalidHandle.  

**Additional context**
Also, it appears that even if the JITP Certificate is imported properly the TLS connect code does not look for an imported file and still looks to access the certificate in prvInitializeClientCredential() by checking for a null or supplied value for keyJITR_DEVICE_CERTIFICATE_AUTHORITY_PEM on line 409

Thank you!

Best
Bobby

dcgaws:
@bgintz-rsc thank you, it sounds like this is indeed a problem. We're working on a repro. If you're inclined to submit a PR to our master branch, we'd be happy to take it.

By way of background, we did not carve out space for the JITR certificate when we did our initial ports of iot_pkcs11_pal.c. That was arguably a requirements miss (and in any case only incidental to this bug), but the tradeoff is that each additional object supported by our PKCS #11 PAL complicates the flash layout of each supported port. The question is, which is the more common use case: that the JITR certificate is common across a given deployment, and therefore can be compiled into the firmware image, or that the JITR certificate may be independent of the flash image, and therefore should be handled by the PKCS #11 PAL? Any comments you could share regarding your application will be welcome.

bgintz-rsc:
Thank you for the quick response.  

I understand the tradeoff.  Ultimately, I feel that importing the JITR into a PKCS object is more secure than hard coding it.  At a minimum, my initial suggestion is to remove the code that is intended to provision the JITR from dev_mode_key_provisioning so as not to cause another developer to attempt to use it.  

That said - I believe the changes required to support this it seems pretty straightforward.  

AFAIK the following changes are required:  

1. Modify the switch statement in prvLabelToFilenameHandle() (to return a defined file name for the label:  pkcs11configLABEL_JITP_CERTIFICATE which is already defined as "JITP Cert"

and

[...]

PR #1234, opened 2019-09-18
Anoxon/jitpfix

alexa-noxon:
Currently, PKCS #11 software implementation expects objects to be stored in "PAL" NVM.

As PKCS #11 is adapted for use with more and more crypto-objects, we need to introduce some flexibility in the retrieval location.  This change adds function pointers to the cached object storage in PKCS #11.

The first alternate set of function pointers is the "Code" function pointers, which provides storage/retrieval options for objects that are likely to be common across a fleet of devices- ie the JITR certificate, the code signer certificate, and the server root CA certificate.

While not technically allowing "Import" of JITR/JITP certificates, this change is meant to provide a work-around for #1119, and allow JITR/JITP certificates to be accessed via PKCS #11. 

Checklist:
----------

- [ ] My code is Linted.


alexa-noxon:
Still need to update comments, add some casts for ESP build to pass.

alexa-noxon:
\bot run checks

PR #1237, opened 2019-09-18
Feature/task notification messages

jsrmalvarez:
Asynchronous message passing
-----------

This feature implements an asynchronous message passing mechanism, for the tasks to be able
to notify and wait for messages.

It's similar to the classic notify and wait mechanism, but with the ablility to pass
more than 32 bits of information.

This is advantageous in scenarios where shared memory can overcomplicate the tasks synchronization
and raise the risk of deadlocks.

For example, if one task reads and formats data from a sensor and other task (GUI) shows the formatted data, and that data is shared between tasks, there must be some synchronization, for the GUI task not showing still unformatted data, and for the sensor reading task not writing data which is still being processed for show.

One option is to use lightweigth mutexes (classic wait and notify).

Another option would be the sensor reading and formatting task copies the info into a message (struct) and pass it to the showing task. When the GUI task receives the message, the information inside is not shared and can be used anytime without synchronization worries.

The drawback of not sharing memory is that the data must be duplicated, at least a period of time.

Checklist:
----------

- [ ] My code is Linted.


RichardBarry:
I have had a quick overview of the code.  Am I right in thinking the mechanics of the notification mechanism have not changed, but now you can send a buffer rather than just the 32-bit value?

One potential concern would be that the buffer allocated in the TCB would have to be the same for every task, whereas it might only get used by one or two tasks, and therefore be wasteful on MCUs where RAM is a premium.  So while I'm very grateful for you posting, in this case I think the light weight stream or message buffers could be used to achieve something similar without the above mentioned draw back.

We are experimenting with enhancing the task notification mechanism to remove the restriction of there only being one per task - which can lead to developers tripping over themselves, especially as the stream and message buffers make use of them themselves.  The experimental code (far from complete) gives each task an array of notifications so the application designer passes an index along with the notification to say which notification to update.

jsrmalvarez:
> I have had a quick overview of the code. Am I right in thinking the mechanics of the notification mechanism have not changed, but now you can send a buffer rather than just the 32-bit value?

Yes. Instead of copying into the task value, I do a memcpy on the task buffer.

> One potential concern would be that the size of the buffer allocated in the TCB would have to be the same for every task, whereas it might only get used by one or two tasks, and therefore be wasteful on MCUs where RAM is a premium. So while I'm very grateful for you posting, in this case I think the light weight stream or message buffers could be used to achieve something similar without the above mentioned draw back.

Yes, the buffer for every task must be set to the maximum size of all the transmitted messages.
That could be enhanced passing buffer size at xTaskCreate.

Wow, I didn't noticed about lightweight stream or message buffers. When I did this modification I was learning about Erlang and FreeRTOS at the same time, and I couldn't resist to implement something similar to Erlang's message passing system into FreeRTOS.

> We are experimenting with enhancing the task notification mechanism to remove the restriction of there only being one per task - which can lead to developers tripping over themselves, especially as the stream and message buffers make use of them themselves. The experimental code (far from complete) gives each task an array of notifications so the application designer passes an index along with the notification to say which notification to update.

Great! That sounds similar the other pull request I did:
[https://github.com/aws/amazon-freertos/pull/1236](https://github.com/aws/amazon-freertos/pull/1236)
At least I'm proud of having spotted a feature that worths implementing.
Do you know when it's going to be available?

Thank you for your reply.
It's great to have this code to play with.

PR #1242, opened 2019-09-19
Increase the "test runner" task delay to 10 secs (from 5 secs)

xuelix:
For test framework being able to catch the test runner task start marker ("----STARTING TESTS---") (for Microchip), the delay is change to 10 secs (from 5 secs).

### DIFF
--- a/tests/common/aws_test_framework.c
+++ b/tests/common/aws_test_framework.c
@@ -80,7 +80,7 @@ void TEST_NotifyTestStart()
 {
     /* Wait for test script to open serial port before starting tests on the
      * board.*/
-    vTaskDelay( pdMS_TO_TICKS( 5000UL ) );
+    vTaskDelay( pdMS_TO_TICKS( 10000UL ) );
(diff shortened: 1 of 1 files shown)


tests/common/aws_test_framework.c
```diff
@@ -80,7 +80,7 @@ void TEST_NotifyTestStart()
 {
     /* Wait for test script to open serial port before starting tests on the
      * board.*/
-    vTaskDelay( pdMS_TO_TICKS( 5000UL ) );
+    vTaskDelay( pdMS_TO_TICKS( 10000UL ) );
```
alexa-noxon: Would we consider something like this that takes a -D that is only included by the test infra build? https://github.com/aws/amazon-freertos/commit/7b43e18e667eef235fd991059b7d849403eff8b2

pavanmr94:
Implemented in AFRHostTools

tests/common/aws_test_framework.c
pavanmr94: Closed.

PR #1246, opened 2019-09-20
Added Test_protect to prevent cascading failures. Fixed callback func…

lundinc2:
…tion signature. Fixed test case for master not initialized API return. Added config for frequency

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_spi.c
+++ b/libraries/abstractions/common_io/test/test_iot_spi.c
@@ -31,45 +31,50 @@
 
 /* Test includes */
 #include <string.h>
+
+/* Test includes */
 #include "unity.h"
 #include "unity_fixture.h"
 
 /* Driver includes */
 #include "iot_spi.h"
+
+/* FreeRTOS includes */
 #include "FreeRTOS.h"
 #include "semphr.h"
 
 #define SPI_BUFFER_SIZE (32)
-#define testIotSPI_DEFAULT_SEMAPHORE_DELAY (100000U)
-#define SPI_TEST_FREQ (500000U)
+#define testIotSPI_DEFAULT_SEMAPHORE_DELAY (3000U)
 
 /*-----------------------------------------------------------*/
 /* Globals values which can be overwritten by the test
  * framework invoking these tests */
 /*-----------------------------------------------------------*/
-uint32_t ultestIotSpiInstance = 3;                            /* Use SPI instance 0 */
-IotSPIMode_t xtestDefaultConfigMode = eSPIMode0;              /* Use SPI Mode eSPIMode0 */
-IotSPIBitOrder_t xtestDefaultconfigBitOrder = eSPIMSBFirst;   /* Use SPI Bit Order eSPIMSBFirst */
-static SemaphoreHandle_t xtestIotSpiSemaphore = NULL;
-
+uint32_t ultestIotSpiInstance = 0;                            /* Test SPI Instance */
+uint32_t ultestIotSPIFrequency = 500000U;                     /* Test SPI Frequency */
+uint32_t ultestIotSPIDummyValue = 0;                          /* Test SPI Dummy Value */
+IotSPIMode_t xtestIotSPIDefaultConfigMode = eSPIMode0;              /* Default SPI eSPIMode0 */
+IotSPIBitOrder_t xtestIotSPIDefaultconfigBitOrder = eSPIMSBFirst;   /* Default SPI bit order eSPIMSBFirst */
+static SemaphoreHandle_t xtestIotSPISemaphore = NULL;
+static StaticSemaphore_t xtestIotSPICompleted;
 /*-----------------------------------------------------------*/
 
-static void prvSpiAsyncCallback(void *pvUserContext)
+/**
+ * @brief Application/POSIX defined callback for asynchronous operations
+ * This callback function releases a semaphore every time it is called.
+ */
+static void prvSpiAsyncCallback( IotSPITransactionStatus_t xStatus, void *pvUserContext )
 {
     BaseType_t xHigherPriorityTaskWoken;
 
-    xSemaphoreGiveFromISR(xtestIotSpiSemaphore, &xHigherPriorityTaskWoken);
+    xSemaphoreGiveFromISR(xtestIotSPISemaphore, &xHigherPriorityTaskWoken);
     portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
 }
-
-
-
-
+/*-----------------------------------------------------------*/
 
 /*-----------------------------------------------------------*/
 /* Define Test Group. */
 TEST_GROUP( TEST_IOT_SPI );
-
 /*-----------------------------------------------------------*/
 
 /**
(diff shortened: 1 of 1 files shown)


lundinc2:
TODO: Run uncrustify after this PR gets approved.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -102,12 +106,12 @@ TEST_GROUP_RUNNER( TEST_IOT_SPI )
...
     RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_CancelSuccess );
 
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_ReadSync );
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_ReadAsync );
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_WriteSync );
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_WriteAsync );
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_TransferSync );
-    //RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_TransferAsync );
+    RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_ReadSync );
+    RUN_TEST_CASE( TEST_IOT_SPI, AFQP_IotSPI_ReadAsync );
```
vjjunnut: These tests require a SPI slave, and are run with assisted test framework. By enabling these here, how do we expect them to run?

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -200,25 +208,29 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSync )
...
+        lRetVal = iot_spi_ioctl( xSPIHandle, eSPISetMasterConfig, &xTestConfig);
+        TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
 
-    //lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetRxNoOfbytes, &xBytesRx);
-    //TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
-    //TEST_ASSERT_EQUAL( xBytesRx, 4 );
+        lRetVal = iot_spi_read_sync( xSPIHandle, ucRxBuf, 4);
+        TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
+
+        lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetRxNoOfbytes, &xBytesRx);
```
vjjunnut: for sync APIs, RxNoOfBytes may not return the bytes. Sync API shall return success only if all the bytes are available.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -31,65 +31,69 @@
...
 /*-----------------------------------------------------------*/
 
 /**
  * @brief Setup function called before each test in this group is executed.
  */
 TEST_SETUP( TEST_IOT_SPI )
 {
-    xtestIotSpiSemaphore = xSemaphoreCreateBinary();
-    TEST_ASSERT_NOT_EQUAL(NULL, xtestIotSpiSemaphore);
+   xtestIotSPISemaphore = xSemaphoreCreateCountingStatic( 10, 0, &xtestIotSPICompleted );
```
brianfr-amazon: Why the change?

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -308,20 +327,24 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_WriteSync )
...
 
-    //lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetTxNoOfbytes, &xBytesTx);
-    //TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
-    //TEST_ASSERT_EQUAL( xBytesTx, 4 );
+        lRetVal = iot_spi_write_sync( xSPIHandle, ucTxBuf, 4);
+        TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
+
+        lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetTxNoOfbytes, &xBytesTx);
+        TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
+        TEST_ASSERT_EQUAL( xBytesTx, 4 );
```
vjjunnut: Same here. for sync APIs, TxNoOfBytes may not return number of bytes transmitted. Its used for async calls. Sync API is supposed to return SUCCESS only if all the bytes are sent.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -681,7 +755,7 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncFuzzing )
     TEST_ASSERT_NOT_EQUAL( NULL, xSPIHandle );
 
     lRetVal = iot_spi_read_sync(NULL, ucBuffer, SPI_BUFFER_SIZE);
-    TEST_ASSERT_EQUAL( IOT_SPI_INVALID_VALUE, lRetVal );
```
brianfr-amazon: Following the same behavior of all the other api's, if we pass in a NULL handle the return is INVALID_VALUE, not NOT_INITIALIZED.

libraries/abstractions/common_io/test/test_iot_spi.c
vjjunnut: Whats the advantage of using counting here?

libraries/abstractions/common_io/test/test_iot_spi.c
lundinc2: I will add the #define config so they can be toggled on or off.

libraries/abstractions/common_io/test/test_iot_spi.c
lundinc2: I was debugging a strange behavior, fixed that now. I'm okay with changing it back to binary, but I would like to keep it static if you agree

libraries/abstractions/common_io/test/test_iot_spi.c
lundinc2: Why can't it be used with sync? What if a partial amount of the operation was sent and we wanted to see how much of our data was able send?

libraries/abstractions/common_io/test/test_iot_spi.c
lundinc2: I agree with this, what is the purpose for IOT_SPI_MASTER_NOT_INITIALIZED in the API? Should we just remove it if we are not going to use it?

libraries/abstractions/common_io/test/test_iot_spi.c
lundinc2: I will revert this to binary

PR #1253, opened 2019-09-20
UDP support for Amazon FreeRTOS Sockets on TI

ericyeh1995:
Description
-----------

Porting UDP on top of TI's [SimpleLink CC3220 WiFi Driver](http://dev.ti.com/tirex/content/simplelink_cc32xx_sdk_1_40_00_03/docs/wifi_host_driver_api/html/index.html). 

Test issues:
RecvFrom peek is not supported, see [here](http://dev.ti.com/tirex/content/simplelink_cc32xx_sdk_1_40_00_03/docs/wifi_host_driver_api/html/group___socket.html#ga02f8b2eb34db39069a1385d4fce80eea).

2. Integration tests
SendTo and RecvFrom loop (to and from the echo server) has fail rate of around 2-3%. Since UDP does not have retries, tests that have multiple SendTo and RecvFrom loops are prone to fail. The cause of this high fail rate should be investigated. 

This PR included some UDP test modification that may help with passing failed tests on other boards.

Checklist:
----------

- [x] My code is Linted.


vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -902,6 +902,207 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+                xAddress.sin_port = pxAddress->usPort;
+                xAddress.sin_addr.s_addr = pxAddress->ulAddress;
+
+                /* Initiate connect. */
+                sTIRetCode = sl_Bind( pxSocketContext->sSocketDescriptor,
+                                      ( const SlSockAddr_t * ) &( xAddress ), /*lint !e9087 !e740 SlSockAddr_t behaves like a union of SlSockAddrIn_t.
+                                                                              * SlSockAddrIn_t is the appropriate interpretation of a IPv4 address. */
+                                      ( _i16 ) sizeof( SlSockAddrIn_t ) );
+
+                /* If connection is successful, mark the socket as connected. */
```
gkwicker: nit: this comment describes the success case but the line beneath tests for the failure case.  I'd either reword the comment as _"If the connection wasn't successful, return a socket error"_ or delete the comment entirely.

vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -902,6 +902,207 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+        {
+            /* Shortcut for easy access. */
+            pxSocketContext = &( xSockets[ ulSocketNumber ] );
+
+            /* Connect. */
+            if( lRetCode == SOCKETS_ERROR_NONE )
+            {
+                /* Set up the address to connect to. */
+                xAddress.sin_family = SL_AF_INET;
+                xAddress.sin_port = pxAddress->usPort;
```
gkwicker: `pxAddress` can be `NULL` here if `prvIsValidSocket( ulSocketNumber )` was true on line 920.  I suspect the condition on line 920 should be an `&&`.

vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -902,6 +902,207 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+        SSocketContextPtr_t pxSocketContext;
+        _i16 sTIRetCode;
+        int32_t lRetCode = SOCKETS_ERROR_NONE;
+        uint32_t ulSocketNumber = ( uint32_t ) xSocket;
+
+        /* Remove compiler warnings about unused parameters. */
+        ( void ) xAddressLength;
+
+        /* Ensure that the socket and address is valid. */
+        if( ( prvIsValidSocket( ulSocketNumber ) == pdTRUE ) || ( pxAddress != NULL ) )
```
gkwicker: If this is supposed to be _"socket __and__ address is valid"_, then the test should read:

```c
if( ( prvIsValidSocket( ulSocketNumber ) == pdTRUE ) && ( pxAddress != NULL ) ) 
```

vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -902,6 +902,207 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+        {
+            /* Shortcut for easy access. */
+            pxSocketContext = &( xSockets[ ulSocketNumber ] );
+
+            /* Connect. */
+            if( lRetCode == SOCKETS_ERROR_NONE )
```
gkwicker: `lRetCode` hasn't been modified since initialization on line 913.  Lines 926/927 & 946 can be removed.

libraries/abstractions/platform/freertos/include/platform/iot_network_udp.h
```diff
@@ -0,0 +1,158 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file iot_network_udp.h
+ * @brief Declares the network stack functions specified in platform/iot_network.h for
+ * Amazon FreeRTOS UDP Sockets.
```
dcgaws: Why are we creating a completely new interface rather than using the existing sockets one?

libraries/abstractions/platform/freertos/include/platform/iot_network_udp.h
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * Passed to #IotNetworkUdp_Create as `pConnectionInfo`.
+ *
+ * All instances of #IotNetworkServerInfoUdp_t should be initialized with
+ * #IOT_NETWORK_SERVER_INFO_UDP_INITIALIZER.
+ * @Note If noBind is set to true, underlying stack must support bind in send
+ */
+typedef struct IotNetworkServerInfoUdp
+{
+    const char * pHostName;     /**< @brief Server host name. Must be NULL-terminated. */
```
dcgaws: Per above, our UDP network interface should be based on Berkeley, including the standard socket options.

libraries/abstractions/platform/freertos/iot_network_udp.c
```diff
@@ -0,0 +1,543 @@
...
+
+    /* Set the receive callback and context. Each pConnection has its own receiveCallback and pReceiveContext. */
+    pNetworkConnection->receiveCallback = receiveCallback;
+    pNetworkConnection->pReceiveContext = pContext;
+
+    /* No flags should be set. */
+    configASSERT( xEventGroupGetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ) ) == 0 );
+
+    /* Create task that waits for incoming data. */
+    if( xTaskCreate( _networkReceiveTask,
```
dcgaws: What do we think is the use case in which an app will be using UDP and want us to incur the overhead of an additional task for handling async receives?

libraries/abstractions/platform/test/iot_test_platform_network_udp.c
```diff
@@ -0,0 +1,563 @@
...
+#include <string.h>
+#include "unity_fixture.h"
+#include "platform/iot_network_udp.h"
+#include "platform/iot_threads.h"
+
+/* UDP server */
+#include "aws_test_tcp.h"
+/*-----------------------------------------------------------*/
+
+#define _TEST_UDP_SERVER_DNS_NAME    "pool.ntp.org"
```
dcgaws: We've been running our own echo server for TCP and TLS. Why do we need the above?

libraries/abstractions/platform/test/iot_test_platform_network_udp.c
```diff
@@ -0,0 +1,563 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Test group runner for Platform Threads tests.
+ */
+TEST_GROUP_RUNNER( UTIL_Platform_Network_UDP )
+{
+    RUN_TEST_CASE( UTIL_Platform_Network_UDP, IotUpdAbstraction_CreateConnectionInvalidParams );
```
dcgaws: I think that the prefix was intended to be "IoTUdpAbstraction_".

vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -902,6 +902,203 @@ int32_t SOCKETS_Send( Socket_t xSocket,
 }
 /*-----------------------------------------------------------*/
 
+#if ( configPLATFORM_SOCKET_UDP_SUPPORT == 1 )
+    int32_t SOCKETS_Bind( Socket_t xSocket,
```
dcgaws: Calling sendto or recvfrom make bind redundant, at least on other major platforms. Is that the intent of our port as well?

libraries/abstractions/platform/freertos/include/platform/iot_network_udp.h
ericyeh1995: Sorry, the branch I was trying to merge this branch into was outdated. Platform related topic should be another PR. This PR is only focusing on UDP socket implementation for TI.

As for the question, the network interface for UDP to mirror the current abstraction layer (in iot_network_freertos.h). Future application may need this abstraction instead of the existing sockets.

vendors/ti/boards/cc3220_launchpad/ports/secure_sockets/iot_secure_sockets.c
ericyeh1995: I'm not sure if sendto and recvfrom make bind redundant. Bind allows API user to select the local port.

abhidixi11: To add  to above comment:  @dcgaws  is correct when UDP is limited to only client software. bind() is not explicitly needed, since a client does not have to listen to a specific local port. sendto will auto bind to a random port. If we ever want to support server like functionality , i.e., listen to a specific port, then bind is required.

gkwicker: I think the question here is if an application can call `SOCKETS_RecvFrom()` or `Sockets_SendTo()` in this port without calling `SOCKETS_Bind()` first.  I suspect that the answer is yes if TI's implementation follows the typical behavior, but it would be good to have tests to show that this is the case.

ericyeh1995: Yes there should be a test to address that. I'm only confident that TI does bind if not bound before a sendTo, since there were tests that were testing other aspect of sendTo without bound. This bring up a question of how do we test recvFrom without sendTo since we need to send data to echo sever before we can receive anything.

PR #1265, opened 2019-09-24
Make TLS test group aware of key import and type capabilities.

dcgaws:
This change improves TLS test group compatibility with secure element-equipped boards by respecting capability configuration settings for key import, ECDSA, and RSA.

### DIFF
--- a/libraries/freertos_plus/standard/tls/test/iot_test_tls.c
+++ b/libraries/freertos_plus/standard/tls/test/iot_test_tls.c
@@ -80,11 +80,15 @@ TEST_TEAR_DOWN( Full_TLS )
 
 TEST_GROUP_RUNNER( Full_TLS )
 {
-    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectEC );
-    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectRSA );
-    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectMalformedCert );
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/standard/tls/test/iot_test_tls.c
```diff
@@ -80,11 +80,15 @@ TEST_TEAR_DOWN( Full_TLS )
...
-    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectUntrustedCert );
-    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectBYOCCredentials );
+    RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectDefault );
+    #if ( pkcs11configIMPORT_PRIVATE_KEYS_SUPPORTED == 1 )
+        #ifdef pkcs11testEC_KEY_SUPPORT
+            RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectEC );
+        #endif
+        RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectMalformedCert );
+        RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectUntrustedCert );
+        RUN_TEST_CASE( Full_TLS, AFQP_TLS_ConnectBYOCCredentials );
```
alexa-noxon: Should we add BYOC credentials to the EC test support?

I'm on the fence about this, since technically it depends on what gets generated into the header files.  But device tester generates EC ones, correct?

dcgaws: Good catch - I updated.

PR #1267, opened 2019-09-24
Fix: CMake for Nordic

ravibhagavandas:
Make Cmake for Nordic working.

Description
-----------
Following changes are done:

* Include all Nordic SDK files
* Override the default compiler and assembler commands to have SES specific compiling and assembler steps.
* Fixes compiler/linker path and options in CMakefile
* Include compiler specific libc libraries under linker libraries.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/nordic/boards/nrf52840-dk/CMakeLists.txt
+++ b/vendors/nordic/boards/nrf52840-dk/CMakeLists.txt
@@ -86,6 +86,7 @@ target_compile_definitions(
 
 set(c_flags
     "-std=gnu99"
+    "-O0"
     "-g"
     "-fmessage-length=0"
     "-fno-diagnostics-show-caret"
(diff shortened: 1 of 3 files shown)


vendors/nordic/boards/nrf52840-dk/CMakeLists.txt
```diff
@@ -68,25 +80,79 @@ set(
...
-"-Map Output/Debug/Exe/aws_tests.map"
-"-u_vectors"
-"-o Output/Debug/Exe/aws_tests.elf"
-"--emit-relocs"
---start-group "@/home/ANT.AMAZON.COM/hbouvier/Desktop/afr-dev/amazon-freertos/vendors/nordic/boards/nrf52840-dk/aws_tests/ses/Output/aws_tests Debug/Obj/aws_tests.ind" --end-group
+    INTERFACE
+    ${defined_symbols}
 )
 
+set(c_flags
```
yuhui-zheng: I'm not sure whether in our CMake structure, there's a way to pass in optimization level flag. If no, I would explicitly call out "no optimization". This way user at least knows where to modify if they want to change optimization level.

ravibhagavandas: Good point. Added optimization level to 0.

PR #1297, opened 2019-09-27
enhance API doc for uart

qiutongs:
Description
-----------

enhance API doc for uart

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/include/iot_uart.h
+++ b/libraries/abstractions/common_io/include/iot_uart.h
@@ -155,7 +155,7 @@ void iot_uart_set_callback( IotUARTHandle_t const pxUartPeripheral,
 /**
  * @brief Starts receiving the data from UART synchronously.
  *
- * This function attempts to read certain number of bytes from slave device to a pre-allocated buffer, in synchronous way.
+ * This function attempts to read certain number of bytes from transmitter device to a pre-allocated buffer, in synchronous way.
  * Partial read might happen, e.g. no more data is available.
  * And the number of bytes that have been actually read can be obtained by calling iot_uart_ioctl.
  *
@@ -186,8 +186,8 @@ int32_t iot_uart_read_sync( IotUARTHandle_t const pxUartPeripheral,
 /**
  * @brief Starts the transmission of data from UART synchronously.
  *
- * This function attempts to write certain number of bytes from a pre-allocated buffer to a slave device, in synchronous way.
- * Partial write might happen, e.g. slave device unable to receive more data.
+ * This function attempts to write certain number of bytes from a pre-allocated buffer to a receiver device, in synchronous way.
+ * Partial write might happen, e.g. receiver device unable to receive more data.
  * And the number of bytes that have been actually written can be obtained by calling iot_uart_ioctl.
  *
  * @warning None of other read or write functions shall be called during this function.
@@ -244,7 +244,7 @@ int32_t iot_uart_read_async( IotUARTHandle_t const pxUartPeripheral,
 /**
  * @brief Starts the transmission of data from UART asynchronously.
  *
- * This function attempts to write certain number of bytes from a pre-allocated buffer to a slave device, in asynchronous way.
+ * This function attempts to write certain number of bytes from a pre-allocated buffer to a receiver device, in asynchronous way.
  * It returns immediately when the operation is started and the status can be check by calling iot_uart_ioctl.
  * Once the operation completes, successful or not, the user callback will be invoked.
  *
--- a/libraries/abstractions/common_io/include/iot_uart.h
+++ b/libraries/abstractions/common_io/include/iot_uart.h
@@ -24,7 +24,7 @@
  */
 
 /**
- * @file iot_uart.h
+ * @file IOT_UART.h
  * @brief File for the HAL APIs of UART called by application layer.
  */
 #ifndef _IOT_UART_H_
@@ -35,7 +35,7 @@
 #include <stdint.h>
 
 /**
- * @defgroup iot_uart UART Abstraction APIs.
+ * @defgroup IOT_UART UART Abstraction APIs.
  * @{
  */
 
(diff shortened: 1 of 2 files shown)


libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -47,34 +47,35 @@
...
-#define IOT_UART_READ_FAILED               ( 3 )
-#define IOT_UART_BUSY                      ( 4 )
-#define IOT_UART_NOTHING_TO_CANCEL         ( 5 )
-#define IOT_UART_FUNCTION_NOT_SUPPORTED    ( 6 )
+#define IOT_UART_SUCCESS                   ( 0 ) /*!< UART operation completed successfully. */
+#define IOT_UART_INVALID_VALUE             ( 1 ) /*!< At least one parameter is invalid. */
+#define IOT_UART_WRITE_FAILED              ( 2 ) /*!< UART driver returns error when performing write operation. */
+#define IOT_UART_READ_FAILED               ( 3 ) /*!< UART driver returns error when performing read operation. */
+#define IOT_UART_BUSY                      ( 4 ) /*!< UART bus is busy at current time. */
+#define IOT_UART_NOTHING_TO_CANCEL         ( 5 ) /*!< No ongoing receiving or transmitting when cancel operation is performed. */
```
lundinc2: nit: operation instead of receiving or transmitting

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -116,23 +117,32 @@ typedef struct
...
- * function callback with the state of the transaction. The callback function and the parameters
- * for the state of the callback are set using iot_uart_set_callback.
- * The application should call this function with the callback function and its parameters.
+ * The callback is guaranteed to be invoked when the current asynchronous operation completes, either successful or failed.
+ * This simply provides a notification mechanism to user's application. It has no impact if the callback is not set.
+ *
+ * @note This callback will not be invoked when synchronous operation completes.
+ * @note This callback is per handle. Each instance has its own callback.
+ * @note Single callback is used for both read_async and write_async. Newly set callback overrides the one previously set.
+ * @warning If the input handle is invalid, this function silently takes no action.
```
lundinc2: Something I believe we should do for all common-io callback functions: We should specify if the user can pass in a  null callback, and if they can pass in a null user context. IMO they should be able to do both. Thoughts?

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: ok

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: Yes, they can do both. That is why I say "If the input handle is invalid, this function silently takes no action."

Do you have better suggestion?

lundinc2: I think the other two inputs should be able to be set to null is what I mean

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -116,23 +117,32 @@ typedef struct
 } IotUARTConfig_t;
 
 /**
- * @brief Initializes the UART peripheral of the board with default configuration.
+ * @brief Initializes the UART peripheral of the board.
```
vjjunnut: For IotUARTConfig_t data structure above. I think we need to make the variables more optimized and update the documentation. 

1) ulParity can be 8 bit -> ucParity, and update the documentation to be "0-> no parity,  1-> even parity and 2->odd_parity"

2) ulWordLength -> It can be ucWordLength as the max word length cannot be more than 4095 bits anyway.

3) ulStopbits -> It can be uint8_t -> ucStopbits as the max bits cannot be more than 4095. Also clarify 0 means no stop bits, 1 means 1 stop bit etc..

4) ulFlowControl -> change to uint8_t -> ucFlowControl -> 0 means disabled, 1 means enabled.

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: I see. Will update.

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: Will update.

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -108,74 +109,102 @@ typedef enum
...
+ *
+ * @note This callback will not be invoked when synchronous operation completes.
+ * @note This callback is per handle. Each instance has its own callback.
+ * @note Single callback is used for both read_async and write_async. Newly set callback overrides the one previously set.
+ * @warning If the input handle is invalid, this function silently takes no action.
  *
  * @param[in] pxUartPeripheral The peripheral handle returned in the open() call.
- * @param[in] xCallback The callback function to be called on completion of transaction.
- * @param[in] pvUserContext The user context to be passed back when callback is called.
+ * @param[in] xCallback The callback function to be called on completion of transaction (This can be NULL).
```
vjjunnut: Why do we have to say it can be NULL? If its NULL, then the function is a no-op

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -108,74 +109,102 @@ typedef enum
...
 void iot_uart_set_callback( IotUARTHandle_t const pxUartPeripheral,
                             IotUARTCallback_t xCallback,
                             void * pvUserContext );
 
 /**
- * @brief Starts receiving the data from UART synchronously i.e. this is a
- * blocking call.
+ * @brief Starts receiving the data from UART synchronously.
+ *
+ * This function attempts to read certain number of bytes from slave device to a pre-allocated buffer, in synchronous way.
```
vjjunnut: There is a no slave device in UART. Can we please change this to "read certain number of bytes from receiver"?

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -108,74 +109,102 @@ typedef enum
...
 
 /**
- * @brief Starts the transmission of data from UART synchronously i.e. this is a
- * blocking call.
+ * @brief Starts the transmission of data from UART synchronously.
  *
- * The application should call this function to start synchronous write.
- * The application should also give the address of the buffer to send the data
- * from with number of bytes to write.
+ * This function attempts to write certain number of bytes from a pre-allocated buffer to a slave device, in synchronous way.
```
vjjunnut: Please remove slave device

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -108,74 +109,102 @@ typedef enum
...
+ * Partial write might happen, e.g. slave device unable to receive more data.
```
vjjunnut: Same here.

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -203,16 +244,29 @@ int32_t iot_uart_read_async( IotUARTHandle_t const pxUartPeripheral,
 /**
  * @brief Starts the transmission of data from UART asynchronously.
  *
- * The application should call this function to start asynchronous write.
- * The application should also give the address of the buffer to send the data
- * from; with number of bytes to write.
+ * This function attempts to write certain number of bytes from a pre-allocated buffer to a slave device, in asynchronous way.
```
vjjunnut: please remove slave device

libraries/abstractions/common_io/include/iot_uart.h
lundinc2: IMO in case an implementation assumes that the callback could never be NULL and a user tries to change it to NULL for a no-op. I think it is a valid case that an application wants to change the callback to NULL without closing the handle. What are your thoughts?

vjjunnut: Yes, understood that it can be NULL for no-op or not set the callback are the same. But do we need to specify that "This can be NULL"?

lundinc2: I think we should or add a test in order to enforce the behavior

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: Oh yeah. I copied some words from I2C doc. I will remove all the slave. Thanks for catching this.

qiutongs: And I believe you meant "read certain number of bytes from transmitter"

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: ok

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: ok

libraries/abstractions/common_io/include/iot_uart.h
qiutongs: ok

libraries/abstractions/common_io/include/iot_uart.h
vjjunnut: looks good.

PR #1301, opened 2019-09-30
Do not autogenerate configuration & kernel patches

karkhaz:
This commit deletes a patch that gets applied to freertos_kernel. It
also prevents prepare.py from generating patch files for configuration
directories, since those generated patches also apply to the kernel.

This commit fixes the breakage caused by freertos_kernel becoming its
own submodule. Our patches get applied with `git patch`, but the patches
applied to files under freertos_kernel now need to be generated relative
to that directory because git patch doesn't enter submodules. A more
permanent fix is forthcoming.


markrtuttle:
Thanks, Kareem!

PR #1303, opened 2019-09-30
UDP support for Amazon FreeRTOS Sockets on NXP

ericyeh1995:
Description
-----------

Porting UDP on Qualcomm QCA4004 for NXP board.

Test issues:
RecvFrom flag is not used in the source code API. However, peek functionality may be available through other means.

2. AFQP_SOCKETS_BindToTheSamePort
The current module does not return error if bound to the same port. Two sockets are able to send from the same port, but will fail on receive.

Checklist:
----------

- [x] My code is Linted.


vendors/nxp/boards/lpc54018iotmodule/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -474,6 +474,204 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+
+    int32_t SOCKETS_SendTo( Socket_t xSocket,
+                            const void * pvBuffer,
+                            size_t xDataLength,
+                            uint32_t ulFlags,
+                            const SocketsSockaddr_t * pxAddress,
+                            Socklen_t xAddressLength )
+    {
+        /* The WiFi module refuses to send data if it exceeds this threshold defined in
+         * atheros_stack_offload.h */
```
abhidixi11: Where this comment is addressed ?  Do we check anywhere if it is increasing the threshold?

vendors/nxp/boards/lpc54018iotmodule/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -474,6 +474,204 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+            {
+                pxContext->xRecvFlags = ( BaseType_t ) ulFlags;
+
+                for( ; ; )
+                {
+                    /* Check if there is anything to be received on this socket. */
+                    xSelectStatus = ( A_STATUS ) t_select( enetCtx,
+                                                           ( uint32_t ) pxContext->xSocket,
+                                                           nxpsecuresocketsONE_MILLISECOND );
+
```
abhidixi11: Do we have to use select here ? Does NXP support non blocking sockets ? Why can't  we use qcom_recvfrom() directly ?

ericyeh1995: Referring to the doc, 3.2.2 UDP receive "recvfrom call is blocking, so select can be used to periodically check for incoming packets."

vendors/nxp/boards/lpc54018iotmodule/ports/secure_sockets/iot_secure_sockets.c
ericyeh1995: Not sure how this comment gets here, will remove if not relevant.

vendors/nxp/boards/lpc54018iotmodule/ports/secure_sockets/iot_secure_sockets.c
ericyeh1995: I did threadsafe test and confirmed that

abhidixi11: OK, so basically their implementation does not support non-blocking sockets.

vendors/nxp/boards/lpc54018iotmodule/ports/secure_sockets/iot_secure_sockets.c
```diff
@@ -474,6 +474,204 @@ int32_t SOCKETS_Send( Socket_t xSocket,
...
+                        }
+                        else
+                        {
+                            lStatus = SOCKETS_SOCKET_ERROR;
+                            break;
+                        }
+                    }
+                    else
+                    {
+                        lStatus = SOCKETS_SOCKET_ERROR;
```
abhidixi11: What happens when select does not detect any data on socket?  Do we return error ? I should return equivalent of EWOULDBLOCK to indicate application that it is an error. 
In network abstraction, we can use this feature to implement PEEK option , if not supported by the socket.

ericyeh1995: When select return A_ERROR it means timeout or no activity. Does

PR #623, opened 2019-05-10
Add CBMC memory-safety proof for SendEventToIPTask

Baha:
Description
-----------

This PR includes the CBMC proof that demonstrates the memory safety of SendEventToIPTask, a function used for sending different events to IP-Task.

Checklist:
----------

- [x] Changes have been tested.
- [x] Configuration files are included.
- [x] Documentation for proof and parts of the test harness.

PR #1317, opened 2019-10-02
Add IP memory safety proofs

markrtuttle:
Add CBMC memory safety proof for SendEventToIPTask originally written by Adrian Palacios and now rebased against the current master branch.

I have verified that the proofs work against the current master.


markrtuttle:
An earlier version https://github.com/aws/amazon-freertos/pull/623 of this pull request was approved and ready for merge.

markrtuttle:
The proofs in this pull request are works in progress.

We are committing them now because they are correct and have value, and because we want them to keep current with code changes.

But they make assumptions, and in the next pass we will do a better job of extracting the assumptions, stating the assumptions, and making the assumptions as weak as possible (to make the proofs as strong as possible).

PR #1321, opened 2019-10-02
Fix buffer overflow in vARPGenerateRequestPacket.

markrtuttle:
Check that Ethernet buffer has enough space for an ARP packet in vARPGenerateRequestPacket.

The check added by this pull request simply returns if there is not enough space, but some
logging or error handling should be done.


markrtuttle:
Closing, already merged as part of https://github.com/aws/amazon-freertos/pull/1322

PR #1327, opened 2019-10-03
Nuvoton - reduce heap size

ericyeh1995:
Description
-----------

Reduce heap size to allow HTTPS demo build on Nutovon. 

Checklist:
----------

- [ ] My code is Linted.


leegeth:
I think we would need more description in the PR and in the commit message for such a change.  The analysis that we have done and how this is verified etc

leegeth:
> I think we would need more description in the PR and in the commit message for such a change. The analysis that we have done and how this is verified etc

Discussed with Gary and agreed that the existing comments are good to go forward with

PR #1332, opened 2019-10-03
[BLE] Merge bonded CB and pairing CB to one API

ravibhagavandas:
Merge bonded CB and pairing CB to one API

Description
-----------

* Remove Bonded CB from tMiddleware, HAL and testsuites
* Merge bonded state change in Pairing Callback APIS
* Modify Nordic and ESP HAL layer port
* Bump up major version to 3.0.0 as its a breaking change.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_common_gap.c
+++ b/vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_common_gap.c
@@ -664,10 +664,10 @@ void prvGAPeventHandler( ble_evt_t const * p_ble_evt,
                    if( xBTCallbacks.pxPairingStateChangedCb && ( status == 0x85 ) ) /* NOTE: Decode other errors */
                    {
                        xBTCallbacks.pxPairingStateChangedCb( eBTStatusFail,
-                                               &xConnectionRemoteAddress,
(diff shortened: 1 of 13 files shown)


vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_common_gap.c
```diff
@@ -1425,17 +1423,6 @@ void prvPmEventHandler( const pm_evt_t * event )
 
     switch( event->evt_id )
     {
-        case PM_EVT_BONDED_PEER_CONNECTED:
```
lundinc2: What's the context for removing this event instead of using pxPairingStateChangedCb?

ravibhagavandas: Bonding is part of the pairing process itself, persisting the security parameters. Hence bond state is included in pairing callback  instead of a separate callback. Also this event handling is not needed as anyways pairingStateChangeCallback will be called after bonded peer connects.

lundinc2: Okay thank you

PR #1340, opened 2019-10-04
Fix provisioning and key import policy on OPTIGA.

dcgaws:
Rationalize key import policy settings. Fix public key slot location. Improve key provisioning flow.

### DIFF
--- a/demos/dev_mode_key_provisioning/README.md
+++ b/demos/dev_mode_key_provisioning/README.md
@@ -12,10 +12,10 @@ If your device does not allow the import of private keys, or if your solution ca
 First, perform the steps in [Configuring the Amazon FreeRTOS Demos](https://docs.aws.amazon.com/freertos/latest/userguide/freertos-configure.html), but skip the last step (that is, don't do *To format your AWS IoT credentials*). The net result should be that the *demos/include/aws_clientcredential.h* file has been updated with your settings, but the *demos/include/aws_clientcredential_keys.h* file has not. 
 
 ### Demo Project Configuration ###
-Open the Hello World MQTT demo as described in the [Getting Started guide](https://docs.aws.amazon.com/freertos/latest/userguide/getting-started-guides.html) for your board. In the project, open the file *aws_dev_mode_key_provisioning.c* and find the definition keyprovisioningFORCE_GENERATE_NEW_KEYSET, which is set to zero by default. Set it to one:
+Open the Hello World MQTT demo as described in the [Getting Started guide](https://docs.aws.amazon.com/freertos/latest/userguide/getting-started-guides.html) for your board. In the project, open the file *aws_dev_mode_key_provisioning.c* and find the definition keyprovisioningFORCE_GENERATE_NEW_KEY_PAIR, which is set to zero by default. Set it to one:
 
 ```
-#define keyprovisioningFORCE_GENERATE_NEW_KEYSET 1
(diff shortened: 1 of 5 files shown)


demos/dev_mode_key_provisioning/README.md
```diff
@@ -36,6 +45,12 @@ Use openssl to format the binary encoded (DER) device public key as PEM:
 openssl ec -inform der -in DevicePublicKeyDer.bin -pubin -pubout -outform pem -out DevicePublicKey.pem
 ```
 
+Don't forget to disable the temporary key generation setting you enabled above. Otherwise, the device will create yet another keyset, and you will have to repeat the previous steps:
```
alexa-noxon: Is the use of the word keyset vs keypair meaningful?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -1027,24 +1031,34 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
...
+        }
 
-                prvExportPublicKey( xSession,
-                                    xProvisionedState.xPublicKey,
-                                    &xProvisionedState.pucDerPublicKey,
-                                    &xProvisionedState.ulDerPublicKeyLength );
-            }
+        /* Ignore errors. If any of the above objects couldn't be read, try to
+         * generate new ones below. */
+        xResult = CKR_OK;
```
alexa-noxon: Why are we ignoring this/do we use the return of prvGetProvsionedState for anything?  

Even if object are not found, the result would still be CKR_OK, right?  What is the error that is returned that we want to ignore?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -1027,24 +1031,34 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
...
+        xResult = xProvisionGenerateKeyPairEC( xSession,
+                                               ( uint8_t * ) pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS,
+                                               ( uint8_t * ) pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS,
+                                               &xProvisionedState.xPrivateKey,
+                                               &xProvisionedState.xPublicKey );
+
+        /* Get the bytes of the new public key. */
+        if( CKR_OK == xResult )
+        {
+            prvExportPublicKey( xSession,
```
alexa-noxon: If we have a keypair already generated, and then we also use the FORCE GENERATE option, is there any chance we double allocate the public key bytes here (first in prvGetProvisionedState above, then again here)?

demos/dev_mode_key_provisioning/README.md
dcgaws: Fixed, thank you.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
dcgaws: I updated the inline comment.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
dcgaws: Fixed, thank you.

PR #1345, opened 2019-10-05
fix: Re-architect Defender Library.

abhidixi11:
Use shared MQTT connection and remove sync call from taskpool.

Before this change, Defender library created and maintained its own
MQTT connection. Use of shared MQTT connection will save memory and
processing time.
Defender library used taskpool task which made blocking calls to
MQTT library. Because MQTT library also uses taskpool, blocking calls
may cause deadlock if limited number of taskpool tasks are available.
This change removes blocking calls from taskpool context.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/defender/aws_iot_demo_defender.c
+++ b/demos/defender/aws_iot_demo_defender.c
@@ -1,6 +1,5 @@
 /*
- * Amazon FreeRTOS V201908.00
- * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ * Copyright (C) 2018 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
  * this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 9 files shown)


demos/defender/aws_iot_demo_defender.c
```diff
@@ -1,6 +1,5 @@
 /*
- * Amazon FreeRTOS V201908.00
- * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
leegeth: We would need the License header with Amazon FreeRTOS V201908.00 info. This can be handled before release as well.

demos/defender/aws_iot_demo_defender.c
```diff
@@ -18,124 +17,63 @@
  * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- * http://aws.amazon.com/freertos
- * http://www.FreeRTOS.org
```
leegeth: This would need to be added back

demos/defender/aws_iot_demo_defender.c
```diff
@@ -18,124 +17,63 @@
...
-    ( void ) awsIotMqttMode;
-    ( void ) pIdentifier;
-    ( void ) pNetworkServerInfo;
-    ( void ) pNetworkCredentialInfo;
-    ( void ) pNetworkInterface;
+    ( void ) pCallbackContext;
 
-    /* Initialize the MQTT library. */
-    mqttInitStatus = IotMqtt_Init();
+    IotLogInfo( "User's callback is invoked on event: %s.", AwsIotDefender_EventType( pCallbackInfo->eventType ) );
```
leegeth: A null check would be required before pCallbackInfo is used

demos/defender/aws_iot_demo_defender.c
abhidixi11: OK

demos/defender/aws_iot_demo_defender.c
```diff
@@ -154,138 +92,216 @@ void _defenderCallback( void * param1,
...
+        {
+            status = EXIT_FAILURE;
+        }
+    }
 
-#if _DEMO_WITH_SOCKET_CONNECTED_TO_ECHO_SERVER == 1
+    if( status == EXIT_SUCCESS )
+    {
+        /* Set metrics report period to 5 minutes(300 seconds) */
+        defenderResult = AwsIotDefender_SetPeriod( 300 );
```
leegeth: wouldn't it be better if you use a constant here?

demos/defender/aws_iot_demo_defender.c
abhidixi11: pCallbackInfo is populated by the library :
https://github.com/aws/amazon-freertos/pull/1345/files#diff-07a09f90dafd00ac812be087a5f9e378R569
So it will not be NULL.
But it is good practice.

libraries/c_sdk/aws/defender/include/aws_iot_defender.h
```diff
@@ -18,9 +18,6 @@
  * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- * http://aws.amazon.com/freertos
- * http://www.FreeRTOS.org
```
leegeth: I think this license header portion would be required in AFR

demos/defender/aws_iot_demo_defender.c
abhidixi11: Yes.

libraries/c_sdk/aws/defender/src/aws_iot_defender_api.c
```diff
@@ -18,11 +18,13 @@
  * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- * http://aws.amazon.com/freertos
- * http://www.FreeRTOS.org
```
leegeth: similar to other comments, this would be required

libraries/c_sdk/aws/defender/src/aws_iot_defender_mqtt.c
```diff
@@ -18,11 +18,9 @@
  * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- * http://aws.amazon.com/freertos
- * http://www.FreeRTOS.org
```
leegeth: This would be required

libraries/c_sdk/aws/defender/src/private/aws_iot_defender_internal.h
```diff
@@ -236,20 +249,11 @@
  * Define encoder/decoder based on configuration AWS_IOT_DEFENDER_FORMAT.
  */
 #if AWS_IOT_DEFENDER_FORMAT == AWS_IOT_DEFENDER_FORMAT_CBOR
-
-    #define DEFENDER_FORMAT     "cbor"
-    #define _defenderEncoder    _IotSerializerCborEncoder       /**< Global defined in iot_serializer.h . */
-    #define _defenderDecoder    _IotSerializerCborDecoder       /**< Global defined in iot_serializer.h . */
-
+    #define DEFENDER_FORMAT    "cbor"
 #elif AWS_IOT_DEFENDER_FORMAT == AWS_IOT_DEFENDER_FORMAT_JSON
```
yuhui-zheng: Minor.. I remember there are SIM issues lingering around for JSON support in Device Defender report. Suggest closing the topic on JSON, and let's keep CBOR only in Defender report. We are very memory constrained...

(Keeping "JSON" in code is hinting subtly that we are going to support it. But really, no... )

libraries/c_sdk/aws/defender/src/aws_iot_defender_mqtt.c
```diff
@@ -37,30 +35,35 @@
...
-    uint16_t thingNameLength = _startInfo.mqttConnectionInfo.clientIdentifierLength;
+    const char * pThingName = _startInfo.pClientIdentifier;
+    uint16_t thingNameLength = _startInfo.clientIdentifierLength;
+    size_t topicPrefixLength = strlen( TOPIC_PREFIX );
 
     /* Calculate topics lengths. Plus one for string terminator. */
-    size_t publishTopicLength = strlen( TOPIC_PREFIX ) + thingNameLength + strlen( TOPIC_SUFFIX_PUBLISH ) + 1;
-    size_t acceptTopicLength = strlen( TOPIC_PREFIX ) + thingNameLength + strlen( TOPIC_SUFFIX_ACCEPTED ) + 1;
-    size_t rejectTopicLength = strlen( TOPIC_PREFIX ) + thingNameLength + strlen( TOPIC_SUFFIX_REJECTED ) + 1;
+    size_t publishTopicLength = topicPrefixLength + thingNameLength + strlen( TOPIC_SUFFIX_PUBLISH ) + 1;
```
yuhui-zheng: Minor. For ```strlen``` usage in general. The length of the macro defined string is known at compile time, and these strings are likely not going to be changed (since that'll require server side change as well). Given that, I think ```strlen``` is unnecessary, and we could further remove dependency on ```string.h```. No harm is done, just commenting as a minimalist.

libraries/c_sdk/aws/defender/src/aws_iot_defender_v1.c
```diff
@@ -36,8 +36,15 @@
 /* Defender agent's status, initialized with eDefenderRepInit. */
 static DefenderReportStatus_t _status = eDefenderRepInit;
 
-static IotNetworkServerInfo_t _AWS_IOT_SERVER_INFO = AWS_IOT_NETWORK_SERVER_INFO_AFR_INITIALIZER;
-static IotNetworkCredentials_t _AWS_IOT_CREDENTIALS = AWS_IOT_NETWORK_CREDENTIALS_AFR_INITIALIZER;
+static IotNetworkServerInfo_t _serverInfo = AWS_IOT_NETWORK_SERVER_INFO_AFR_INITIALIZER;
```
yuhui-zheng: I believe ```aws_iot_defender_v1.c``` is not being used. Do we really need these changes?

libraries/c_sdk/aws/defender/src/private/aws_iot_defender_internal.h
```diff
@@ -236,20 +249,11 @@
...
-
-    #define DEFENDER_FORMAT     "json"
-    #define _defenderEncoder    _IotSerializerJsonEncoder       /**< Global defined in iot_serializer.h . */
-    #define _defenderDecoder    _IotSerializerJsonDecoder       /**< Global defined in iot_serializer.h . */
-
+    #define DEFENDER_FORMAT    "json"
```
yuhui-zheng: I'd propose removing "json"

yuhui-zheng:
Let's put this PR in test infra and collect some data on each platform. 

To be investigated:
- QoS0 vs QoS1 for defender MQTT publish. 
(When WiFi/Ethernet driver is not able to transmit a report probably due to size, we need to confirm defender is not informing user "published". e.g. I'm seeing this on STM32L475, where WiFi driver can only transmit 1200 characters. With QoS0 and previous version of defender, we are showing user report transmitted, but in reality not. )

Approving since no regression.

libraries/c_sdk/aws/defender/src/private/aws_iot_defender_internal.h
abhidixi11: OK will remove.

libraries/c_sdk/aws/defender/src/aws_iot_defender_v1.c
abhidixi11: Yes, we need to investigate that. It was failing compilation and right now I don't want to modify cmake , project files , so just made changes so that it compiles.

libraries/c_sdk/aws/defender/src/private/aws_iot_defender_internal.h
abhidixi11: removed.

PR #1348, opened 2019-10-06
Fix buffer overflow in vendor queue code

markrtuttle:
This patch fixes a potential buffer overflow in Nordic vendor queue code.

This vendor just happens to be the one picked up in the CBMC memory safety proofs of the kernel Queue code.  I'm not sure if I have any business patching vendor code, and I'm not sure if we should be checking all vendor code fo this issue.

### DIFF
--- a/vendors/nordic/nRF5_SDK_15.2.0/external/freertos/source/queue.c
+++ b/vendors/nordic/nRF5_SDK_15.2.0/external/freertos/source/queue.c
@@ -501,6 +501,8 @@ static void prvInitialiseNewQueue( const UBaseType_t uxQueueLength, const UBaseT
 	Queue_t *pxNewQueue;
 	const UBaseType_t uxMutexLength = ( UBaseType_t ) 1, uxMutexSize = ( UBaseType_t ) 0;
 
+		configASSERT( pxStaticQueue );
+
 		/* Prevent compiler warnings about unused parameters if
 		configUSE_TRACE_FACILITY does not equal 1. */
 		( void ) ucQueueType;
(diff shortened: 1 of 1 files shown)


vendors/nordic/nRF5_SDK_15.2.0/external/freertos/source/queue.c
```diff
@@ -1919,7 +1922,11 @@ Queue_t *pxQueue;
 
 	taskENTER_CRITICAL();
 	{
-		uxReturn = pxQueue->uxLength - pxQueue->uxMessagesWaiting;
+		if(pxQueue->uxMessagesWaiting <= pxQueue->uxLength){
+			uxReturn = pxQueue->uxLength - pxQueue->uxMessagesWaiting;
+		}else{
+			uxReturn = 0;
```
karkhaz: I'm not sure about this. It seems to me that `pxQueue->uxMessagesWaiting <= pxQueue->uxLength` is an invariant, so this on an invariant violation branch. Silently returning doesn't seem like the best idea.

Deferring to @dcgaws @dan4thewin for logging or error handling ideas, but alternatively, I suggest `configASSERT( pxQueue->uxMessagesWaiting <= pxQueue->uxLength )` and then just do the subtraction.

markrtuttle: I think it is reasonable to return 0 for space available when the inequality fails.  I don't think that is failing silently.

markrtuttle:
Closing this pull request to issue issue a pull request against the freertos kernel itself that asserts there is room in the buffer.

PR #1372, opened 2019-10-09
Modify PKCS #11 GenerateRandom to return entropy & CRYPTO layer to ma…

alexa-noxon:
…intain DRBG

In order to maintain consistency with Secure Elements and reduce the number of bus
crosses when randomness is required, this commit modifies PKCS #11 C_GenerateRandom
to return entropy, rather than to return random numbers from a DRBG.

The DRBG is now maintained in the CRYPTO layer, and all applications requiring randomness
should now call CRYPTO_GetRandomBytes, or if suppling the f_rng to mbedTLS,
CRYPTO_GenerateRandomBytesMbedTls with a NULL context.

Test Run:

https://amazon-freertos-ci.corp.amazon.com/job/master/job/custom_ide/233/

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -3789,10 +3789,10 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateKeyPair )( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
-extern int renamed_mbedtls_hardware_poll( void * data,
-                                          unsigned char * output,
-                                          size_t len,
-                                          size_t * olen );
+extern int ulPortGetEntropyFromHardware( void * data,
+                                         unsigned char * output,
+                                         size_t len,
+                                         size_t * olen );
 
 /**
  * @brief Generate cryptographically random bytes.
@@ -3825,7 +3825,7 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateRandom )( CK_SESSION_HANDLE xSession,
 
     if( xResult == CKR_OK )
     {
-        lMbedResult = renamed_mbedtls_hardware_poll( NULL, pucRandomData, ulRandomLen, &olen );
+        lMbedResult = ulPortGetEntropyFromHardware( NULL, pucRandomData, ulRandomLen, &olen );
     }
 
     if( ( lMbedResult != 0 ) || ( olen != ulRandomLen ) )
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -3856,6 +3856,12 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateRandom )( CK_SESSION_HANDLE xSession,
                                                    &pucRandomData[ totalBytes ],
                                                    ulRandomLen - totalBytes,
                                                    &olen );
+
+            if( lResult != 0 )
+            {
+                break;
+            }
+
             totalBytes += olen;
             lLoops++;
         }
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -3869,7 +3869,7 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateRandom )( CK_SESSION_HANDLE xSession,
 
     if( ( lResult != 0 ) || ( totalBytes != ulRandomLen ) )
     {
-        PKCS11_PRINT( ( "ERROR: Failed to get entropy from hardware. Error %d, Bytes %d\r\n ", lResult, olen ) );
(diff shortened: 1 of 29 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3817,6 +3789,11 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateKeyPair )( CK_SESSION_HANDLE xSession,
     return xResult;
 }
 
+extern int ulPortGetEntropyFromHardware( void * data,
```
dcgaws: lPortGetEntropyFromHardware

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3836,6 +3813,7 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateRandom )( CK_SESSION_HANDLE xSession,
 {
     CK_RV xResult = CKR_OK;
     int lMbedResult = 0;
```
dcgaws: We're not calling mbed routines here anymore so this name becomes confusing. However, inconveniently, CK_RV is unsigned while the Port call below returns signed. Still, since we're not capturing a specific status code from the Port call, it seems that we don't need a separate local variable.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
alexa-noxon: whoops fixing now thanks.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
alexa-noxon: Added a PKCS print on error, so I will rename to be lResult so I can still print that out.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3847,14 +3850,28 @@ CK_DECLARE_FUNCTION( CK_RV, C_GenerateRandom )( CK_SESSION_HANDLE xSession,
 
     if( xResult == CKR_OK )
     {
-        lMbedResult = mbedtls_ctr_drbg_random( &xP11Context.xMbedDrbgCtx, pucRandomData, ulRandomLen );
-
-        if( lMbedResult != 0 )
+        while( ( lLoops < ENTROPY_MAX_LOOPS ) && ( totalBytes < ulRandomLen ) )
```
dan4thewin: Is there any blocking behavior here?  Is there another interface that permits a process to block until enough entropy is available?

libraries/freertos_plus/standard/crypto/src/iot_crypto.c
```diff
@@ -263,6 +278,101 @@ void CRYPTO_ConfigureThreading( void )
                                aws_mbedtls_mutex_unlock );
 }
 
+
+void CRYPTO_ConfigureDRBG( void )
```
dan4thewin: Is there always exactly one context?

libraries/freertos_plus/standard/crypto/src/iot_crypto.c
```diff
@@ -263,6 +278,101 @@ void CRYPTO_ConfigureThreading( void )
...
+    if( ( NULL == pRandomBytes ) ||
+        ( length == 0 ) )
+    {
+        CRYPTO_PRINT( ( "ERROR: Invalid inputs to CRYPTO_GetRandomBytes \r\n" ) );
+        lResult = CRYPTO_ERROR_RNG;
+    }
+
+    if( lResult == 0 )
+    {
+        lResult = mbedtls_ctr_drbg_random( &xDrbg.xMbedDrbgCtx, pRandomBytes, length );
```
dan4thewin: Does there need to be a test of xIsInitialized here?

libraries/freertos_plus/standard/crypto/src/iot_crypto.c
alexa-noxon: Yes, in the CRYPTO layer I am going to maintain only 1 DRBG that all other applications/libraries and request data from.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
alexa-noxon: since entropy calls are port specific, I don't think generally there is a entropy available callback that I can rely on, and I would be dependent on the port implementation to decide when it wants to yield.

libraries/freertos_plus/standard/crypto/src/iot_crypto.c
alexa-noxon: yes thank you fixed.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
alexa-noxon: After offline conversation, have added a  delay to give hardware time to gather more entropy

PR #1402, opened 2019-10-15
AFR Sockets UDP Integration Test Revisit

ericyeh1995:
Description
-----------

Cherry pick UDP test fix into this commit. 

Added UDP echo server for test.

The majority of this PR is allow UDP test to pass even if some of the packets are lost. As an example, the test SendRecv_VaryLength expects the micro-controller to send UDP data to and from the echo server perfectly for 80 times. Since UDP is an unreliable connectionless protocol, a fail is inevitable and one error in SendTo or RecvFrom will lead to a fail in the test. Some of the boards are using WiFi which causes data transmission to be less dependable.

Assume a setup that has UDP SendTo and RecvFrom loop to an echo server failing 2% of the time. A test that requires UDP SendTo and RecvFrom to execute perfectly 80 times has the fail rate of 80%. The adjustment in this PR added retries on integration tests, if a data is not send or if data is not received from the echo server, the test will rerun the SendTo RecvFrom loop again with the same parameter. The test will fail if SendTo RecvFrom loop is retried udptestMAX_RETRY times. Adding retry drastically reduces the fail rate, continuing with the above example if we set udptestMAX_RETRY to 2 the test fail rate is reduced to 0.064%. 

This PR modified the following UDP test:
- SendRecv_VaryLength
- Threadsafe_SameSocketDifferentTasks

Checklist:
----------

- [x] My code is Linted.


libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -479,32 +478,107 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+    xResult = pdFAIL;
+    ulRetry = 0;
+
+    memset( pucTxBuffer, udptestTX_BUFFER_FILLER, udptestBUFFER_SIZE );
+
+    prvCreateTxData( ( char * ) pucTxBuffer,
+                     xLength,
+                     ulTxCount );
+
+    for( ; ulRetry < udptestMAX_RETRY; ulRetry++ )
```
gkwicker: Why do we need retries in the test?  Is this test reliable?  How do we know we have enough/too many retries?

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -1004,7 +1045,12 @@ static void prvEchoClientTxTask( void * pvParameters )
 
         if( xReturned <= 0 )
         {
-            xStatus = pdFAIL;
+            /* xStatus = pdFAIL; */
```
gkwicker: Why is this commented out?  If it's not needed it should be deleted.

tests/include/aws_test_tcp.h
```diff
@@ -30,10 +30,10 @@
  * Update udptestECHO_SERVER_ADDR# and
  * udptestECHO_PORT with IP address
  * and port of unencrypted TCP echo server. */
-#define udptestECHO_SERVER_ADDR0         10
-#define udptestECHO_SERVER_ADDR1         184
-#define udptestECHO_SERVER_ADDR2         166
-#define udptestECHO_SERVER_ADDR3         154
+#define udptestECHO_SERVER_ADDR0         34
```
gkwicker: Why is an IP address hard-coded in here?

abhidixi11: We do have hard address for TCP echo server  as well. (see below tcptestECHO_SERVER_ADDR0 ...) this file is for the test infra and hence has a hard coded address for testing with our test server (although not the best design), Ideally this server should be the same  as TCP echo server, but running UDP tests internally with external EC2 machine does not work.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -479,32 +478,107 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+            if( pucTxBuffer[ xIndex ] != udptestTX_BUFFER_FILLER )
+            {
+                xReturn = pdFAIL;
+                udptestFAILUREPRINTF( ( "Tx buffer padding modified. Message Length %d, Byte Index %d \r\n", xMessageLength, xIndex ) );
+            }
         }
     }
 
-    udptestASSERT( xReturn == pdPASS );
+    /* udptestASSERT( xReturn == pdPASS ); */
```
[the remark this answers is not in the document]
ericyeh1995: I commented it out so test won't fail when data mismatch happen. Understanding UDP, there should not be any memory mismatch, send and receive should either 1) fail or 2) data integrity is preserved. However, I observe there is a probability that memory mismatch is going to happen due to reasons not related to UDP. The observation is below:

During some stress test on the retries, there were cases where data send is different from data received.
If you take a look at a [recent test](https://paste.amazon.com/show/ericyeh/1575681793) testing with 8000 send and recv loops (for each data length of {1, 2, 7, 8, 9, 100, 1000, udptest_MAX_FRAME_SIZE} we do 1000 loops continuously, after 1000 loops are done we move to the next data length). The observation is there are busts of data mismatch: on loop index 3938 to 3999, on 7733 to 7999. Those data mismatch may be caused by previous loops failures that is caused by congestion, that's why we see it fail until loop index 3999 and 7999, between 3999 and 4000 and between 7999 and 8000 there is a short delay during switch of data length to relief network congestion. 

The above stress test is still under experimentation: [I added delay each time the send recv loop failed and the memory mismatch error stop happening.](https://paste.amazon.com/show/ericyeh/1575683857) I'll uncomment it if I adjust the test such that the probability of memory mismatch is 0.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: The reason behind retries should be in the PR description. Adding retries to 2 on TI board in lab setup should be enough that test will only fail due to network reasons once every 4 years of nightly test (using the 2% fail rate per loop). But the fail rate is different on each board, and adding wifi to the equation may increase the fail rate. 

Currently the retry is set to 8. This amount of retries should given us assurance that the if UDP fail it may be the setup or implementation and will not have to worry about network/wifi issue.

tests/include/aws_test_tcp.h
gkwicker: This is common library code and its tests should have no dependencies on any particular HW lab environment or setup.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
gkwicker: I have a number of questions on this PR:

1. What is the purpose of this test?  To determine if UDP has been implemented correctly or to determine that it can work some percentage of the time?
1. How do we know that the test is failing due to connectivity or infrastructure problems rather than an intermittent bug in the code?
1. This library should be platform-independent, why are we basing the number of retries in a test on historical results from a particular board (TI) in a particular HW lab?  This test will be run in many different environments and on many different platforms.  Would this number of retries be appropriate for a platform which uses wired Ethernet or PPP over serial for example?

ericyeh1995: 1. We are trying to make sure implementation of the socket is correct and works on hardware. If the test is passed, we are confident the device is able to send and receive UDP data, vary in data length.
2. With the same socket code on a device the fail rate varies randomly (from 0.1% to 2%+) when testing at different time. We can try to find the root cause of the random factor, otherwise we are suspecting instability of connection to the server.
3. The library is platform-independent, this test will run on all platforms the same configuration. Tests on software socket libraries (FreeRTOS+TCP and LWIP) has the lowest fail rate passing all 80 tries, but occasionally one or two loops will fail causing the whole test to fail. Hardware implementation of UDP are observed to fail more often. We can adjust the number of retries base on different platform, but may require more formal testing to determine those numbers.

tests/include/aws_test_tcp.h
ericyeh1995: This IP is public.

ericyeh1995: Squashed commit so internal test UDP echo server address is not in record

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -498,6 +506,91 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+                                       xLength,
+                                       pxEchoServerAddress );
+
+        if( xIntermResult != pdPASS )
+        {
+            udptestPRINTF( ( "Tx %d data failed to send on try %d\r\n", ulTxCount, ulRetry ) );
+            continue;
+        }
+
+        memset( pucRxBuffer, udptestRX_BUFFER_FILLER, udptestBUFFER_SIZE );
```
dan4thewin: Given that later the tx and rx buffers are compared, I think the contents should be varied - either random or oscillating.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -498,6 +506,91 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+    {
+        xIntermResult = prvSendHelper( xSocket,
+                                       pucTxBuffer,
+                                       xLength,
+                                       pxEchoServerAddress );
+
+        if( xIntermResult != pdPASS )
+        {
+            udptestPRINTF( ( "Tx %d data failed to send on try %d\r\n", ulTxCount, ulRetry ) );
+            continue;
```
dan4thewin: I notice you have vTaskDelay() for other cases below, but not this one.  Consider having one vTaskDelay() at the top of the loop when ulRetry > 0.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -511,7 +604,7 @@ TEST_GROUP_RUNNER( Full_UDP )
     RUN_TEST_CASE( Full_UDP, AFQP_SOCKETS_SendToInvalidParams );
     RUN_TEST_CASE( Full_UDP, AFQP_SOCKETS_RecvFromInvalidParams );
     RUN_TEST_CASE( Full_UDP, AFQP_SOCKETS_SendRecv_VaryLength );
-    RUN_TEST_CASE( Full_UDP, AFQP_SOCKETS_RecvFromWithPeek );
+    /* RUN_TEST_CASE( Full_UDP, AFQP_SOCKETS_RecvFromWithPeek ); */
```
dan4thewin: Any reason this is commented out?

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -1123,6 +1199,22 @@ static void prvSOCKETS_Threadsafe_SameSocketDifferentTasks()
...
+            }
+            else
+            {
+                ulRetry++;
+
+                if( ulRetry > udptestMAX_RETRY )
+                {
+                    TEST_FAIL();
+                }
+            }
```
dan4thewin: This strikes me as convoluted.  I think it would be simpler to track total successes and total failures, where passing is having enough successes without passing a failure threshold.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: should be uncommented

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -498,6 +506,91 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+{
+    BaseType_t xResult;
+    BaseType_t xIntermResult;
+    uint8_t * pucTxBuffer = ( uint8_t * ) pcTxBuffer;
+    uint8_t * pucRxBuffer = ( uint8_t * ) pcRxBuffer;
+    SocketsSockaddr_t xReceiveAddress = { 0 };
+    uint32_t ulRetry;
+
+    xIntermResult = pdPASS;
+    xResult = pdPASS;
```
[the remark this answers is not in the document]
ericyeh1995: Good point, the total number of tries should be udptestMAX_RETRY+1 since retry implies it had done at least one loop already. Will be consistent with prvSOCKETS_Threadsafe_SameSocketDifferentTasks().

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: Tx buffers is varied in prvCreateTxData()

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: Adding a delay on every loop may delay the test. Since we are not sure if adding delay (and how much) helps with higher chance of successful transmission, I'm inclined to add it only when loop failed.
Without delay at each loop, the platform with the highest fail rate is able to achieve 99.90% successful transmission.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
dan4thewin: Excellent!

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: This was one of the design choice. The reason retry is used rather than finding the success rate is retry is faster and feels simpler. Taking 10% loop fail rate and 100 loop as an example.

With the retry method where number of retry (on top of the first try) is greater than 2, on average the test will perform 111.1 loops. (Out of 100 initial loops 10 loops will retry the first time, out of the 10 retries 1 loop will do another retry, and so on.) Calculating the test fail rate given 2 max retries in addition to the first try, we get 9.5%. Each incremental of max retry will reduce the test fail rate by factor of approximately 10, also each incremental of max retry does not increase the number of loop the test perform thus time is saved.

Using the success rate method, and again working with 100 loops. If we set the threshold to 90, the test will fail 54% of the time. We can reduce the test fail rate by 1) increase number of loops or 2) reduce the threshold. Increasing the number of loop will increase the number of time it take and will only converge to the theoretical number (which is 50%). If we were to reduce the threshold we can set any number between 1 and number of loops such that test will only fail say once every 1000 tests. In this case, we are reducing the number of success loop in exchange for higher chance of test passing, and if we want to mitigate and increase the number of success loop we will have to increase the number of loop it will run which will increase the amount of time the test will take. This method will require multiple variable tuning and does not perform as well as the retry method.

ericyeh1995: There are also integration tests that only perform a few loops. May need to increasing the number of loops by 10 fold to get something statistically significant.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
```diff
@@ -498,6 +506,92 @@ static BaseType_t prvCheckRxTxBuffers( uint8_t * pucTxBuffer,
...
+
+        xIntermResult = prvCheckRxTxBuffers( pucTxBuffer,
+                                             pucRxBuffer,
+                                             xLength );
+
+        if( xIntermResult != pdPASS )
+        {
+            udptestPRINTF( ( "Rx %d data mismatch on try %d\r\n", ulTxCount, ulRetry ) );
+
+            /* Pause, failure may cause by network congestion */
```
dcgaws: Referring to earlier comment from gwicker, we need to distinguish between failure cases. On one hand, dropped and out of order packets are expected in UDP. Fragmentation is a potential issue as well. In all cases, the application code (e.g. this test) must mitigate those risks via appropriate logic. 

On the other hand, packet corruption due to the network is unlikely. If we've handled the cases above, and we're seeing corruption, that's a device bug and it should be flagged immediately rather than retried.

The net result should be that either (a) the test is able to successfully receive the transmitted buffer every time by requesting re-transmission of specific dropped data segments or (b) the test hard fails because a corrupted packet was received.

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: Modified to threshold method

libraries/abstractions/secure_sockets/test/iot_test_udp.c
ericyeh1995: The test now fails if send data is different from received data

PR #1406, opened 2019-10-15
add doxygen config and fix header syntax

qiutongs:
Description
-----------

add doxygen config and fix header syntax

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/doc/config/common_io
+++ b/doc/config/common_io
@@ -1,0 +1,24 @@
+# Include common configuration options.
+@INCLUDE_PATH = doc/config
+@INCLUDE = common
+
+# Basic project information.
+PROJECT_NAME = "Common I/O"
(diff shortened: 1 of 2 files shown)


doc/config/common_io
```diff
@@ -0,0 +1,24 @@
...
+# Generate Doxygen tag file for this library.
+GENERATE_TAGFILE = doc/output/common_io.tag
+
+# Directories containing library source code.
+INPUT = doc/lib/ \
+        libraries/abstractions/common_io/include
+
+
+# Library file names.
+FILE_PATTERNS = iot_i2c.h iot_uart.h
```
lundinc2: Are we only including these two in the docs?

qiutongs: No. There will be more. This is just initial commit.

PR #1409, opened 2019-10-15
Rangpava/changelog update

pavanmr94:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


CHANGELOG.md
```diff
@@ -26,12 +26,16 @@
...
 - Update the PKCS #11 third-party headers to be the latest from OASIS (version 2.40 Plus Errata 01).
+- Add PKCS #11 as a submodule.
 
 #### Bluetooth Low Energy
 - ESP32 board supports NIMBLE as the default underlying stack for Bluetooth Low Energy.
 - Contains bug fixes and enhancements in ESP NIMBLE stack.
 - Contains bug fixes in ESP Bluedroid stack.
 
+#### Marvell SDK Update
+- Marvell SDK now updated to R8 P2 (v1.2.r8.p2)
```
yuhui-zheng: We never used a version number for Marvell SDK. (If Marvell has an official versioning "system", through which users could easily get what each version is about, I think using version number may make sense.) Or how about:

- Provided static libraries for ```psm, rfget, partition, wifidriver, and wlcmgr```.
- Added ```mdev``` driver.

yuhui-zheng: From Pavan -- https://github.com/aws/amazon-freertos/blob/4abfab81c1d511c97cd35834af64082ab0bac58f/vendors/marvell/WMSDK/mw320/build/amzsdk.mk#L31 -- "v1.2.r8"

PR #1411, opened 2019-10-15
add uart code example in header

qiutongs:
Description
-----------

add uart code example in header

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/include/iot_uart.h
+++ b/libraries/abstractions/common_io/include/iot_uart.h
@@ -149,6 +149,40 @@ typedef struct
  * - 'NULL', if
  *     - invalid instance number
  *     - open same instance more than once before closing it
+ *
+ * <b>Example</b>
+ * @code{c}
+ *  // These two buffers can contain 32 bytes for reading and writing.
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -149,6 +149,40 @@ typedef struct
...
+ *
+ *  IotUARTHandle_t xOpen;
+ *  int32_t lRead, lWrite, lClose;
+ *  BaseType_t xCallbackReturn;
+ *  uint8_t ucPort = 1; /* Each UART peripheral will be assigned an integer.
+ *
+ *  xOpen = iot_uart_open( ucPort );
+ *  if( xOpen != NULL )
+ *  {
+ *      iot_uart_set_callback( xOpen, prvReadWriteCallback, NULL );
```
lundinc2: I think we should add the prvReadWriteCallback function to this code block to give the user an example of what a callback could look like.

qiutongs: Sure, we can improve it later.

Issue #1357, opened 2019-10-08
[GreenGrass] Can't create new MQTT connection if the previous one was not closed


ahmedwahdan:
**Describe the bug**
I'm using greengrass demo to connect to greengrass core over EC2 instance.
I'm using mobile phone as AP.
If the hardware was restarted without closing the connection, I have to reboot the mobile phone!

Edit:
Sometimes even with closing the connection, new connection can't be created.

**System information**
- ESP32
- VSCode
- Operating System [Windows]
- Version of Amazon FreeRTOS (201906.00_Major-539-g3a519702)
- Project [Greengrass]

**Expected behavior**
This issue is not visible when using other demos, I think the issue may be related to using the MQTT_Agent instead of directly using MQTT v2 APIs directly

Thank you!

sarenameas:
Hello ahmedwahdan,

How do you restart the hardware? Do you power cycle? 

I am trying this out.

ahmedwahdan:
Hello Sarenameas,
Sometimes after successful demo complete, and some times if I restart the esp32 (power reset) before closing the connection, the connection can't established again with the greengrass core.

sarenameas:
Could you please turn on logging in the TCP/IP stack in 
**vendors\espressif\boards\esp32\aws_demos\config_files\FreeRTOSIPConfig.h** by setting:
```
#define ipconfigHAS_DEBUG_PRINTF    1
```
Then please post the logs here.

ahmedwahdan:
` make flash monitor
Flashing binaries to serial port COM6 (app at offset 0x20000 )...
esptool.py v2.6
Serial port COM6
Connecting.....
Chip is ESP32D0WDQ6 (revision 1)
Features: WiFi, BT, Dual Core, Coding Scheme None
MAC: 30:ae:a4:22:ce:80
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 921600
Changed.
Configuring flash size...
Auto-detected Flash size: 4MB
Compressed 23712 bytes to 14071...
Wrote 23712 bytes (14071 compressed) at 0x00001000 in 0.2 seconds (effective 1100.8 kbit/s)...
Hash of data verified.
Compressed 1099136 bytes to 669493...
Wrote 1099136 bytes (669493 compressed) at 0x00020000 in 10.6 seconds (effective 827.2 kbit/s)...
Hash of data verified.
Compressed 3072 bytes to 141...
[...]

PR #1413, opened 2019-10-15
Fix FreeRTOS+TCP rand32 mapping in ESP32.

sarenameas:
- This was mapping to c std rand() and therefore the same numbers were being output each board reset.
- ulRand() defined in the secure sockets port for FreeRTOS+TCP maps to the ESP32 hardware entropy logic.

This fixes issue #1357 

Checklist:
----------

- [ ] My code is Linted.


sarenameas:
\bot run checks

Issue #1380, opened 2019-10-10
iot_demo_freertos.c _cleanup missing AwsIotNetworkManager_DisableNetwork( configENABLED_NETWORKS );


teuteuguy:
**Describe the bug**
When wanting to run the demo code multiple times, the _cleanup function does not cleanup the wifi network. Looks like it's simply missing a call to AwsIotNetworkManager_DisableNetwork( configENABLED_NETWORKS );

**System information**
- Which hardware board or part numbers? ESP32
- IDE used: vscode
- Operating System [Windows|Linux|MacOS] Mac
- Version of Amazon FreeRTOS (run ``git describe --tags`` to find it) latest
- Project [Custom Application|MQTT|Shadow|Greengrass|OTA|Device Defender|TCP Echo|Sub pub Demo|Simple TCP Server] MQTT
- If your project is a Custom Application, please add the relevant code snippet in the section `Code to reproduce the bug`. 

**Expected behavior**
I expect the demo code to be able to be run multiple times

sarenameas:
How exactly are you trying to run the demo multiple times? Could you post your code snippet or changes?
I am able to rerun the same demo code multiple times by adding the following in **demos\demo_runner\iot_demo_freertos.c** :
```
#include "platform/iot_clock.h"
...
        while ( true )
        {
            /* Run the demo. */
            status = pContext->demoFunction( true,
                                            clientcredentialIOT_THING_NAME,
                                            pConnectionParams,
                                            pCredentials,
                                            pNetworkInterface );

            /* Log the demo status. */
            if( status == EXIT_SUCCESS )
            {
                /* DO NOT EDIT - This message is used in the test framework to
                * determine whether or not the demo was successful. */
                IotLogInfo( "Demo completed successfully." );
            }
            else
            {
                IotLogError( "Error running demo." );
            }
            IotClock_SleepMs( 1000 );
        }
```

teuteuguy:
Hi @sarenameas,

yes, your code will work, however this assumes that the _initialize has been executed before once, before a re-run.

My objective was to have a full init, and cleanup at each run of the demo code.
In the context of the networking stack / wifi etc ... if you want to re-run the _initialize function, you need the AwsIotNetworkManager_DisableNetwork( configENABLED_NETWORKS ); to be executed in the _cleanup.

My simple change is in the _cleanup function with the following code:
```
static void _cleanup( void )
{
    /* Remove network manager subscription */
[...]

PR #1416, opened 2019-10-16
disable network in demo cleanup code so that demo can be re-run

qiutongs:
Description
-----------

disable network in demo cleanup code so that demo can be re-run

Related issue: https://github.com/aws/amazon-freertos/issues/1380

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/demo_runner/iot_demo_freertos.c
+++ b/demos/demo_runner/iot_demo_freertos.c
@@ -283,6 +283,7 @@ static int _initialize( demoContext_t * pContext )
  */
 static void _cleanup( void )
 {
+    AwsIotNetworkManager_DisableNetwork( configENABLED_NETWORKS );
     /* Remove network manager subscription */
     AwsIotNetworkManager_RemoveSubscription( subscription );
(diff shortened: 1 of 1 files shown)


demos/demo_runner/iot_demo_freertos.c
```diff
@@ -283,6 +283,7 @@ static int _initialize( demoContext_t * pContext )
  */
 static void _cleanup( void )
 {
+    AwsIotNetworkManager_DisableNetwork( configENABLED_NETWORKS );
```
yuhui-zheng: Should MQTT unsubscribe happen before disabling network? 

In setup -- we start network first, then MQTT connect, then subscribe. In tear down, I think we should do things in the exact opposite order.

qiutongs: I thought about this and Ravi told me the order doesn't matter here. 
@ravibhagavandas do you have any comment?

ravibhagavandas: Subscriptions are stored separately from network enable/disable context. So it shouldn't matter if you remove subscriptions  later

PR #1419, opened 2019-10-16
Remove client credential from greengrass discovery

muneebahmed10:
Description
-----------

Removed demo includes from aws_clientcredential.h in greengrass library code.
Updated greengrass demo to use v2 MQTT APIs instead of v1.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/greengrass/include/aws_greengrass_discovery.h
+++ b/libraries/freertos_plus/aws/greengrass/include/aws_greengrass_discovery.h
@@ -31,7 +31,6 @@
 #ifndef _AWS_GREENGRASS_DISCOVERY_H_
 #define _AWS_GREENGRASS_DISCOVERY_H_
 #include "FreeRTOS.h"
-#include "aws_clientcredential.h"
 #include "iot_secure_sockets.h"
 
 /**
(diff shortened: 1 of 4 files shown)


libraries/freertos_plus/aws/greengrass/include/aws_greengrass_discovery.h
```diff
@@ -85,7 +90,10 @@ typedef struct
  * @return If connection was successful then pdPASS is
  * returned.  Otherwise pdFAIL is returned.
  */
-BaseType_t GGD_GetGGCIPandCertificate( char * pcBuffer,
+BaseType_t GGD_GetGGCIPandCertificate( const char * pcHostAddress,
```
qiutongs: For security reason, people usually suggest to add string length aside string pointer. Consider about it.

muneebahmed10: This commit was meant mainly to update the greengrass library to remove macros included from demos headers, so no changes were made to the library besides moving these to parameters. Since including the string length along with `pcHostAddress` would require adding a field to `GGD_HostAddressData_t`, I think that change would be more appropriate when rewriting the library.

qiutongs: Got it. I agree.

PR #1422, opened 2019-10-17
Don't check for ECDSA support if it's disabled in static configuration.

dcgaws:
If ECDSA support has been disabled via static configuration, the PKCS #11 test module should completely skip test cases that involve that cryptosystem.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -825,8 +825,10 @@ CK_DECLARE_FUNCTION( CK_RV, C_GetMechanismInfo )( CK_SLOT_ID slotID,
     {
         { CKM_RSA_PKCS,        { 2048, 2048, CKF_SIGN              } },
         { CKM_RSA_X_509,       { 2048, 2048, CKF_VERIFY            } },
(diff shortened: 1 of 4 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -825,8 +825,10 @@ CK_DECLARE_FUNCTION( CK_RV, C_GetMechanismInfo )( CK_SLOT_ID slotID,
     {
         { CKM_RSA_PKCS,        { 2048, 2048, CKF_SIGN              } },
         { CKM_RSA_X_509,       { 2048, 2048, CKF_VERIFY            } },
-        { CKM_ECDSA,           { 256,  256,  CKF_SIGN | CKF_VERIFY } },
-        { CKM_EC_KEY_PAIR_GEN, { 256,  256,  CKF_GENERATE_KEY_PAIR } },
+        #if ( pkcs11configSUPPRESS_ECDSA_MECHANISM != 1 )
```
alexa-noxon: In a world where we had a lot of time, I feel like it would better to make a pkcs11configEC_KEY_SUPPORT, then to get rid of the pkcs11testEC_KEY_SUPPORT and use one config to rule this information.

PR #1433, opened 2019-10-18
Prevent overflow of Renesas RNG buffer

alexa-noxon:
When data requested is not a multiple of 4 bytes,
mbedtls_hardware_poll could have overflowed the output
buffer.

Tests
https://amazon-freertos-ci.corp.amazon.com/job/master/job/custom_ide/262/console
Rebuild after PR feedback:
https://amazon-freertos-ci.corp.amazon.com/job/master/job/custom_ide/274/console (all tests pass)

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/renesas/amazon_freertos_common/entropy_hardware_poll.c
+++ b/vendors/renesas/amazon_freertos_common/entropy_hardware_poll.c
@@ -21,7 +21,7 @@ int mbedtls_hardware_poll( void *data,
     R_INTERNAL_NOT_USED(data);
 
     uint32_t random_number = 0;
-    size_t num_bytes = ( len < 4 ) ? len : 4;
+    size_t num_bytes = ( len < sizeof( uint32_t ) ) ? len : sizeof( uint32_t );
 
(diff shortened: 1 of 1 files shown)


vendors/renesas/amazon_freertos_common/entropy_hardware_poll.c
```diff
@@ -19,15 +19,15 @@ int mbedtls_hardware_poll( void *data,
                            unsigned char *output, size_t len, size_t *olen )
 {
     R_INTERNAL_NOT_USED(data);
-    R_INTERNAL_NOT_USED(len);
 
     uint32_t random_number = 0;
+    size_t num_bytes = ( len < 4 ) ? len : 4;
```
dan4thewin: Would it make more sense to use sizeof( uint32_t ) above?  Or to use 4 below?  Seems like maybe they ought to be the same.

alexa-noxon: yes good thinking.

Issue #1436, opened 2019-10-19
crash when using IotNetworkInterface->setReceiveCallback with lwip porting


eddid:
hi there,

I meet crash when using `IotNetworkInterface->setReceiveCallback`, you could recurring the issue with following code piece

```
#include "platform/iot_network.h"
#include "platform/iot_network_afr.h"
void Echo_ReceiveCallback(void *pNetworkConnection, void *pReceiveContext)
{
    char buffer[32];
    const IotNetworkInterface_t *pNetworkInterface = (const IotNetworkInterface_t *)pReceiveContext;
    size_t recvLength;

    recvLength = pNetworkInterface->receive(pNetworkConnection, (uint8_t *)buffer, sizeof(buffer));
    vLoggingPrintf("%s %d %.*s\n", __FUNCTION__, recvLength, recvLength, buffer);
}

void handle_echo_task(void *pvParameters)
{
    const IotNetworkInterface_t *pNetworkInterface = IOT_NETWORK_INTERFACE_AFR;
    IotNetworkServerInfo_t *pServerInfo = (IotNetworkServerInfo_t *)pvParameters;
    void *pNetworkConnection = NULL;
    size_t msgLength, sendLength;
    IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;

    IotSdk_Init();

    networkStatus = pNetworkInterface->create(pServerInfo, NULL, &pNetworkConnection);
    if (networkStatus != IOT_NETWORK_SUCCESS)
    {
        vLoggingPrintf("%s:%d networkStatus=%d\n", __FUNCTION__, __LINE__, networkStatus);
        goto exit;
    }

    networkStatus = pNetworkInterface->setReceiveCallback(pNetworkConnection, Echo_ReceiveCallback, (void *)pNetworkInterface);
    if (networkStatus != IOT_NETWORK_SUCCESS)
    {
        vLoggingPrintf("%s:%d networkStatus=%d\n", __FUNCTION__, __LINE__, networkStatus);
        goto exit;
    }

    vTaskDelay(1);

    msgLength = strlen("IOT_NETWORK_INTERFACE_AFR");
    sendLength = pNetworkInterface->send(pNetworkConnection, "IOT_NETWORK_INTERFACE_AFR", msgLength);
    if (sendLength != msgLength)
    {
        vLoggingPrintf("%s:%d %d != %d\n", __FUNCTION__, __LINE__, sendLength, msgLength);
        goto exit;
    }

exit:
[...]

PR #1437, opened 2019-10-19
modify IotNetworkAfr_Close in iot_network_freertos.c to fixed crash issue

eddid:
mqtt create a thread for receive data in block mode, so when mqtt close the socket, the receiving thread might be blocked, we should not shutdown RX in this situation, notify receiving thread shutdown RX would be more safe

Description
-----------

- the socket of mqtt works in full duplex mode with two threads, one for TX and another for RX(passive), the TX thread should change the socket to half duplex mode and wait, other than close the socket directly

- it is a critical issue with lwip porting, while freertos_plus_tcp porting works well, since there is a eSOCKET_CLOSED event triger in FreeRTOS_recv

```
BaseType_t FreeRTOS_recv( Socket_t xSocket, void *pvBuffer, size_t xBufferLength, BaseType_t xFlags )
{
   ...
                /* Block until there is a down-stream event. */
                xEventBits = xEventGroupWaitBits( pxSocket->xEventGroup,
                    eSOCKET_RECEIVE | eSOCKET_CLOSED | eSOCKET_INTR,
                    pdTRUE /*xClearOnExit*/, pdFALSE /*xWaitAllBits*/, xRemainingTime );
   ...
}
```

- the PR referred to #1436 

Checklist:
----------

- [+] I have tested my changes. No regression in existing tests.
- [+] My code is Linted.

### DIFF
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -634,7 +634,7 @@ IotNetworkError_t IotNetworkAfr_Close( void * pConnection )
 
     /* Call Secure Sockets shutdown function to close connection. */
     socketStatus = SOCKETS_Shutdown( pNetworkConnection->socket,
-                                     SOCKETS_SHUT_RDWR );
+                                     SOCKETS_SHUT_WR );
 
     if( socketStatus != SOCKETS_ERROR_NONE )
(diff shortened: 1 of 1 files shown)


ravibhagavandas:
Closing the PR as the workaround is not required with latest fixes in lwip `2.1.2`

PR #1441, opened 2019-10-21
Implement LWIP_NETCONN_SEM_PER_THREAD and LWIP_NETCONN_FULLDUPLEX

dan4thewin:
These changes implement the minimum functionality necessary to enable lwIP's "full duplex" mode.  This mode "enables code that allows reading from one thread, writing from a 2nd thread and closing from a 3rd thread at the same time."

To keep the implementation self-contained, the per-task semaphore lookup function does the allocation on first use by a given task, and the semaphore is never freed.  This is compatible with the standard persistent task model.

This work is the companion of the code committed in https://github.com/aws/amazon-freertos/pull/1017/commits/e6dee5a266a2bbf01c1aef47ea40b55935c607b3 .

The TCP regression suite passes, including:
TEST(Full_TCP, AFQP_SOCKETS_Threadsafe_SameSocketDifferentTasks) PASS
TEST(Full_TCP, AFQP_SOCKETS_Threadsafe_DifferentSocketsDifferentTasks) PASS

### DIFF
--- a/libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
+++ b/libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
@@ -101,15 +101,16 @@ err_t sys_mbox_new( sys_mbox_t *pxMailBox, int iSize )
  * Outputs:
  *      sys_mbox_t              -- Handle to new mailbox
  *---------------------------------------------------------------------------*/
-void sys_mbox_free( volatile sys_mbox_t *pxMailBox )
+void sys_mbox_free( sys_mbox_t *pxMailBox )
 {
     unsigned long ulMessagesWaiting;
     QueueHandle_t xMbox;
     TaskHandle_t xTask;
+    sys_mbox_t volatile * pvxMailBox = pxMailBox;
 
-    if( pxMailBox != NULL )
+    if( pvxMailBox != NULL )
     {
-        ulMessagesWaiting = uxQueueMessagesWaiting( pxMailBox->xMbox );
+        ulMessagesWaiting = uxQueueMessagesWaiting( pvxMailBox->xMbox );
         configASSERT( ( ulMessagesWaiting == 0 ) );
 
         #if SYS_STATS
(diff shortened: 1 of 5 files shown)


libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
```diff
@@ -578,6 +578,46 @@ sys_thread_t xReturn;
...
+        sys_sem_t *sem;
+        err_t err;
+        /* allocate memory for this semaphore */
+        sem = mem_malloc( sizeof( sys_sem_t ) );
+        configASSERT( sem != NULL );
+        err = sys_sem_new( sem, 0 );
+        configASSERT( err == ERR_OK );
+        configASSERT( sys_sem_valid( sem ) );
+        vTaskSetThreadLocalStoragePointer( task, 0, sem );
+        ret = sem;
```
alexa-noxon: Is there any chance that 2 tasks both try to grab the semaphore before it's been created and cause this semaphore to be double allocated?

libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
```diff
@@ -578,6 +578,46 @@ sys_thread_t xReturn;
...
+    TaskHandle_t task = xTaskGetCurrentTaskHandle();
+    configASSERT( task != NULL );
+
+    ret = pvTaskGetThreadLocalStoragePointer( task, 0 );
+    if( ret == NULL )
+    {
+        sys_sem_t *sem;
+        err_t err;
+        /* allocate memory for this semaphore */
+        sem = mem_malloc( sizeof( sys_sem_t ) );
```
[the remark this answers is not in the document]
dan4thewin: lwIP provides more than one memory allocation scheme, which are configured via include files and may differ by platform.  Using mem_malloc respects those configurations.

libraries/3rdparty/lwip/src/portable/arch/sys_arch.c
dan4thewin: The semaphore is per-task, so if two tasks grab semaphores, they get two different semaphores.  The lwIP approach to concurrency is non-intuitive.

PR #1461, opened 2019-10-25
Add exponential retry logic to all MQTT System tests.

sarenameas:
Add exponential retry logic to all MQTT System tests.

Description
-----------
To reduce the probability of failures in MQTT connect because of transient network issues, we add exponential retries to all IotMqtt_Connect() calls in all MQTT_System tests.

This change once approved will be repeated in the CSDK. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
+++ b/libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
@@ -60,12 +60,19 @@
  * Provide default values of test configuration constants.
  */
 #ifndef IOT_TEST_MQTT_TIMEOUT_MS
-    #define IOT_TEST_MQTT_TIMEOUT_MS      ( 5000 )
+    #define IOT_TEST_MQTT_TIMEOUT_MS             ( 5000 )
 #endif
 #ifndef IOT_TEST_MQTT_TOPIC_PREFIX
-    #define IOT_TEST_MQTT_TOPIC_PREFIX    "iotmqtttest"
+    #define IOT_TEST_MQTT_TOPIC_PREFIX           "iotmqtttest"
+#endif
+#ifndef IOT_TEST_MQTT_CONNECT_RETRY_COUNT
+    #define IOT_TEST_MQTT_CONNECT_RETRY_COUNT    ( 1 )
 #endif
 /** @endcond */
+
+#if IOT_TEST_MQTT_CONNECT_RETRY_COUNT < 1
+    #error "IOT_TEST_MQTT_CONNECT_RETRY_COUNT must be at least 1."
+#endif
 
 /**
  * @brief Determine which MQTT server mode to test (AWS IoT or Mosquitto).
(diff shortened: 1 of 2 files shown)


tests/include/iot_config_common.h
```diff
@@ -252,6 +252,10 @@ struct IotMqttSerializer;
 extern const struct IotMqttSerializer * IotTestNetwork_GetSerializer( void );
 #define IOT_TEST_MQTT_SERIALIZER    IotTestNetwork_GetSerializer()
 
+/* Retry the MQTT Connections in the MQTT System unit tests for all hardware
+   platforms supported in Amazon FreeRTOS. */
+#define IOT_TEST_MQTT_CONNECT_RETRY_ENABLED  1
```
sarenameas: This macro will not be defined in the CSDK so retires will not be enabled.

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
```diff
@@ -211,6 +211,40 @@ static bool _disconnectSerializerOverride = false;  /**< @brief Tracks if #_disc
...
+static IotMqttError_t _mqttConnect( const IotMqttNetworkInfo_t * pNetworkInfo,
+                          const IotMqttConnectInfo_t * pConnectInfo,
+                          uint32_t timeoutMs,
+                          IotMqttConnection_t * const pMqttConnection )
+{
+    IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
+#if ( IOT_TEST_MQTT_CONNECT_RETRY_ENABLED == 1)
+    int32_t retried = 0;
+    uint32_t periodMs = 200;
+    int32_t retries = 3;
```
gkwicker: Let's make sure the retry times are consistent with the default values used in other SDKs, e.g. the [JavaScript SDK.](https://github.com/aws/aws-iot-device-sdk-js/blob/master/device/index.js#L341)

sarenameas: Thanks, updating to 1000ms + a bit (100ms). https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html

gkwicker: Why is 3 the right number of retries?

sarenameas: There is no particular reason for this number other than it feeling good

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
```diff
@@ -211,6 +211,46 @@ static bool _disconnectSerializerOverride = false;  /**< @brief Tracks if #_disc
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Establish an MQTT connection. Retry if retries are enabled.
```
gordonwang0: Nit: Change second sentence to "Retry if enabled."

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
```diff
@@ -211,6 +211,46 @@ static bool _disconnectSerializerOverride = false;  /**< @brief Tracks if #_disc
...
+ */
+static IotMqttError_t _mqttConnect( const IotMqttNetworkInfo_t * pNetworkInfo,
+                                    const IotMqttConnectInfo_t * pConnectInfo,
+                                    uint32_t timeoutMs,
+                                    IotMqttConnection_t * const pMqttConnection )
+{
+    IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
+
+    #if ( IOT_TEST_MQTT_CONNECT_RETRY_ENABLED == 1 )
```
gordonwang0: Let's make this configurable as the maximum number of retries (instead of just 0/1).

```c
// Set this to the number of connection attempts for the MQTT tests.
// For example, 3 retries.  To disable retries, set to 1.
#define IOT_TEST_MQTT_RETRY_COUNT   3
```

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
sarenameas: ok

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
sarenameas: ok

libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
```diff
@@ -211,6 +211,46 @@ static bool _disconnectSerializerOverride = false;  /**< @brief Tracks if #_disc
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Establish an MQTT connection. Retry if enabled.
+ */
+static IotMqttError_t _mqttConnect( const IotMqttNetworkInfo_t * pNetworkInfo,
```
gordonwang0: At the top of this file (before line 62) could you add:

```c
#ifndef IOT_TEST_MQTT_CONNECT_RETRY_COUNT
#define IOT_TEST_MQTT_CONNECT_RETRY_COUNT 1
#endif
```

And also before line 70:
```c
#if IOT_TEST_MQTT_CONNECT_RETRY_COUNT < 1
#error "IOT_TEST_MQTT_CONNECT_RETRY_COUNT must be at least 1."
#endif
```

You can then remove all the preprocessor directives from this function and just have a for loop.

PR #1465, opened 2019-10-28
BUGFIX: Marvell MAC address was not properly initialized in TCP stack

yuhui-zheng:
Description
-----------
```IP-task``` is created before ```wifi_driver``` and ```wlcmgr``` task. Right after entering IP task, network down event is sent to the queue in the same task to initialize local MAC address. Due to the condition ```if ( xInterfaceState == INTERFACE_UP )```, the code skips reading MAC address from NVM and fails updating IP stack local MAC address properly. 

The fix is proposed given reading MAC address does not need WiFi layer to be ready prior. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/mw300_rd/NetworkInterface.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/mw300_rd/NetworkInterface.c
@@ -148,22 +148,16 @@ void handle_data_packet(const t_u8 interface, const t_u8 *rcvdata,
 
 BaseType_t xNetworkInterfaceInitialise( void )
 {
-    static BaseType_t xMACAdrInitialized = pdFALSE;
     uint8_t ret;
-
-    if (xInterfaceState == INTERFACE_UP) {
-	if (xMACAdrInitialized == pdFALSE) {
-	    mac_addr_t mac_addr;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/mw300_rd/NetworkInterface.c [hunk 1]
```diff
@@ -148,22 +148,16 @@ void handle_data_packet(const t_u8 interface, const t_u8 *rcvdata,
...
-	    }
-	    FreeRTOS_UpdateMACAddress(mac_addr.mac);
-	    xMACAdrInitialized = pdTRUE;
-	 }
-	 return pdTRUE;
-    }
+	}
+	FreeRTOS_UpdateMACAddress(mac_addr.mac);
+
     return pdFALSE;
```
ravibhagavandas: Seems the function always returns pdFALSE now. Should the return type checked by TCP IP stack to know if the interface is initialized.

yuhui-zheng: Good catch.. was too deep into the fix, left the return out... 

(The return logic followed the same logic we previously have. I think, the next step is reaching out to Marvell to see if there are good this this fix. )

PR #1471, opened 2019-10-29
Update library versions for 201910.00

pavanmr94:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Amazon FreeRTOS Secure Sockets V1.1.6
+ * Amazon FreeRTOS Secure Sockets V1.1.7
```
alexa-noxon: Do we keep the lwip and freertos+tcp sockets versions in unison always?  Not 100% sure if any changes were made to lwip

sarenameas:
Looks like we are missing files under *libraries\abstractions\posix\include\FreeRTOS_POSIX*?

PR #1477, opened 2019-10-29
Add read long value test

zshijing:
Description
-----------

 - Add read long value test
 - rewrite BLE_Write_Notification_Size_Greater_Than_MTU_3 test case

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/ble_hal/test/src/iot_test_ble_hal_afqp.c
+++ b/libraries/abstractions/ble_hal/test/src/iot_test_ble_hal_afqp.c
@@ -113,6 +113,9 @@ TEST_GROUP_RUNNER( Full_BLE )
     #if ENABLE_TC_AFQP_WRITE_LONG
         RUN_TEST_CASE( Full_BLE, BLE_Property_WriteLongCharacteristic );
     #endif
+    #if ENABLE_TC_AFQP_READ_LONG
+        RUN_TEST_CASE( Full_BLE, BLE_Property_ReadLongCharacteristic );
+    #endif
 
     RUN_TEST_CASE( Full_BLE, BLE_Property_WriteCharacteristic );
     RUN_TEST_CASE( Full_BLE, BLE_Property_WriteDescriptor );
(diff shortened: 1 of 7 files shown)


libraries/abstractions/ble_hal/test/src/iot_test_ble_hal_afqp.c
```diff
@@ -651,6 +654,44 @@ TEST( Full_BLE, BLE_Property_WriteLongCharacteristic )
...
 
+TEST( Full_BLE, BLE_Property_ReadLongCharacteristic )
+{
+    BLETESTreadAttrCallback_t xReadEvent;
+    BTGattResponse_t xGattResponse;
+    BLETESTconfirmCallback_t xConfirmEvent;
+    BTStatus_t xStatus;
+    uint8_t LongReadBuffer[ bletests_LONG_WRITE_LEN ];
+
+    memset( LongReadBuffer, 49, bletests_LONG_WRITE_LEN * sizeof( uint8_t ) );
```
ravibhagavandas: Instead of memset you can directly assign to {0} in the declaration above.

libraries/abstractions/ble_hal/test/src/iot_test_ble_hal_afqp.c
```diff
@@ -651,6 +654,44 @@ TEST( Full_BLE, BLE_Property_WriteLongCharacteristic )
     }
 }
 
+TEST( Full_BLE, BLE_Property_ReadLongCharacteristic )
+{
+    BLETESTreadAttrCallback_t xReadEvent;
+    BTGattResponse_t xGattResponse;
+    BLETESTconfirmCallback_t xConfirmEvent;
+    BTStatus_t xStatus;
+    uint8_t LongReadBuffer[ bletests_LONG_WRITE_LEN ];
```
ravibhagavandas: bletests_LONG_WRITE_LEN - Can we have a separate define for the long reads, like bletests_LONG_READ_LEN 

Also its better to not allocate long read buffer on stack, as it might over flow the stack buffer. Could be a static global variable instead.

zshijing: Sure

libraries/abstractions/ble_hal/test/src/iot_test_ble_hal_afqp.c
zshijing: Right now we just use the same value { '1', '1', '1', ...} that the long write test sent. It's easier to check the value on the RPi side.

PR #1500, opened 2019-11-05
AFR NimBLE: Misc fixes in prvBTSendResponse, prvValidGattRequest and application code (aws_demos & aws_tests)

prasad-alatkar:
Helps in fixing the integration test:TEST( Full_BLE_Integration_Test_Connection, BLE_Send_Data_After_Disconected )​ getting stuck in middle.

Description
-----------
- Fix return code in `prvBTSendResponse`.
- Changes `prvValidGattRequest` to handle valid/invalid gatt request properly.
- Introduce `prvGattGiveSemaphore` API inline with `prvGattGetSemaphore`.
- Remove redundant `esp_bt_controller_mem_release( ESP_BT_MODE_BTDM )` from **application_code** of **aws_tests** and **aws_demos** as `ESP_BT_MODE_CLASSIC_BT` memory is already released at the start. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/aws_demos/application_code/main.c
+++ b/vendors/espressif/boards/esp32/aws_demos/application_code/main.c
@@ -185,11 +185,6 @@ static void prvMiscInitialization( void )
 
             xRet = esp_bt_controller_mem_release( ESP_BT_MODE_BLE );
 
-            if( xRet == ESP_OK )
-            {
-                xRet = esp_bt_controller_mem_release( ESP_BT_MODE_BTDM );
-            }
-
             return xRet;
         }
 
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -466,6 +466,7 @@ BTStatus_t prvBTGattServerInit( const BTGattServerCallbacks_t * pxCallbacks )
     BTStatus_t xStatus = eBTStatusSuccess;
 
     ble_hs_cfg.gatts_register_cb = prvGATTRegisterCb;
+    serviceCnt = 0;
 
     memset( espServices, 0, sizeof( struct ble_gatt_svc_def ) * ( MAX_SERVICES + 1 ) );
 
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -47,11 +47,19 @@ static struct ble_gatt_svc_def espServices[ MAX_SERVICES + 1 ];
 static BTService_t * afrServices[ MAX_SERVICES ];
 static uint16_t serviceCnt = 0;
 static SemaphoreHandle_t xSem;
+static bool xSemLock;
 uint16_t gattOffset = 0;
 
 void prvGattGetSemaphore()
 {
+    xSemLock = 1;
     xSemaphoreTake( xSem, portMAX_DELAY );
+}
+
+void prvGattGiveSemaphore()
+{
+    xSemaphoreGive( xSem );
+    xSemLock = 0;
 }
 
 void * pvPortCalloc( size_t xNum,
@@ -624,7 +632,7 @@ BTStatus_t prvBTSendIndication( uint8_t ucServerIf,
 
 static bool prvValidGattRequest()
 {
-    if( uxSemaphoreGetCount( xSem ) == 0 )
+    if( xSemLock )
     {
         return true;
     }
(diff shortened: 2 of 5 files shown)


vendors/espressif/boards/esp32/aws_demos/application_code/main.c
```diff
@@ -185,11 +185,6 @@ static void prvMiscInitialization( void )
 
             xRet = esp_bt_controller_mem_release( ESP_BT_MODE_BLE );
 
-            if( xRet == ESP_OK )
-            {
-                xRet = esp_bt_controller_mem_release( ESP_BT_MODE_BTDM );
-            }
```
dhrishi: This is removed because, controller memory for classic BT is already freed in case of NimBLE. So, we just need to free controller memory for BLE in case of cleanup.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -47,13 +47,21 @@ static struct ble_gatt_svc_def espServices[ MAX_SERVICES + 1 ];
...
+    xSemLock = 1;
     xSemaphoreTake( xSem, portMAX_DELAY );
 }
 
+void prvGattGiveSemaphore()
+{
+    xSemaphoreGive( xSem );
+    xSemLock = 0;
+}
+
```
ravibhagavandas: There could be a race condition here, if a request is already being processed, and a new request comes, it sets xSemLock = 1 and waits for the semaphore. Old request will post the semaphore, and resets xSemLock to 0. So it can cause prvValidRequest() to return false for the second request.

prasad-alatkar: Hi @ravibhagavandas 

> if a request is already being processed, and a new request comes, it sets xSemLock = 1 and waits for the semaphore. Old request will post the semaphore, and resets xSemLock to 0

This is not possible here because nimble host is running in a single thread. The same thread takes the semaphore (gets blocked till required content is received from AFR thread), and gives the semaphore when it is done using it (it has received required content). So there is no possibility of "_new request coming, when old request is being processed_ ". 
CC: @dhrishi Please correct me if you find anything wrong in the explanation.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -373,10 +381,13 @@ static int prvGATTCharAccessCb( uint16_t conn_handle,
             trans_id = ( uint32_t ) ctxt;
             if( xGattServerCb.pxRequestWriteCb != NULL )
             {
+                xSemLock = 1;
```
dhrishi: Hold on. @prasad-alatkar Why is this done here? Shouldn't this be set in the `else` part of the below `if` block?

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -373,10 +381,13 @@ static int prvGATTCharAccessCb( uint16_t conn_handle,
...
+
                 if( ( ctxt->op == BLE_GATT_ACCESS_OP_WRITE_CHR ) && ( ctxt->chr->flags & BLE_GATT_CHR_F_WRITE_NO_RSP ) )
                 {
                     need_rsp = 0;
                     trans_id = 0;
+                    xSemLock = 0;
```
dhrishi: Why is this set to 0 here?

dhrishi: Ohh, so basically you set `xSemLock` for all cases and reset it if the request is write without response, so this will fine. But, I think this can be done the other way for better clarity.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -373,10 +381,13 @@ static int prvGATTCharAccessCb( uint16_t conn_handle,
...
                 }
 
                 if( xGattServerCb.pxRequestWriteCb != NULL )
```
dhrishi: Redundant `if` check. Can be removed.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -456,6 +466,7 @@ BTStatus_t prvBTGattServerInit( const BTGattServerCallbacks_t * pxCallbacks )
     BTStatus_t xStatus = eBTStatusSuccess;
 
     ble_hs_cfg.gatts_register_cb = prvGATTRegisterCb;
+    serviceCnt = 0;
```
prasad-alatkar: This small change addresses the issue observed when creating services twice, e.g. enabling "BLE_Write_Notification_Size_Greater_Than_MTU_3​" and "BLE_Send_Data_After_Disconnected" at the same time
while running `iot_test_ble_hal_integration`. Previous to this change the valid handle was not getting initialized.
CC: @ravibhagavandas @dhrishi

PR #1518, opened 2019-11-12
Convert lwIP to a submodule

dan4thewin:
In this commit, the lwip directory is replaced with a submodule pointing at the head of the lwIP distribution.  I chose the head because of the numerous bug fixes added since the last stable release of lwIP one year ago.  At least one of those bugs disastrously impacted the full duplex feature.

The FreeRTOS and board specific files that previously occupied the portable directory under lwip have been relocated.

https://amazon-freertos-ci.corp.amazon.com/job/master/job/custom/job/cypress/job/cypress54-ide/112/console


.gitmodules
```diff
@@ -14,3 +14,6 @@
 	path = libraries/3rdparty/pkcs11
 	url = https://github.com/amazon-freertos/pkcs11.git
 	branch = v2.40_errata01
+[submodule "libraries/3rdparty/lwip"]
```
pavanmr94: --name should be "lwip" when doing a git submodule add

dan4thewin: fixed

.gitmodules
```diff
@@ -14,6 +14,6 @@
 	path = libraries/3rdparty/pkcs11
 	url = https://github.com/amazon-freertos/pkcs11.git
 	branch = v2.40_errata01
-[submodule "libraries/3rdparty/lwip"]
+[submodule "lwip"]
 	path = libraries/3rdparty/lwip
 	url = git@github.com:lwip-tcpip/lwip.git
```
pavanmr94: this should be the https URL, not ssh

dan4thewin: fixed

libraries/abstractions/lwip_osal/src/sys_arch.c
```diff
@@ -0,0 +1,712 @@
+/*
```
gkwicker: Verified that this file is a duplicate of [the original](https://github.com/aws/amazon-freertos/blob/master/libraries/3rdparty/lwip/src/portable/arch/sys_arch.c) with minor formatting changes.

PR #1530, opened 2019-11-18
Fix for prvConnectHelperWithRetry function in iot_test_tcp.c

shubhamkulkarni97:
Description
-----------

The function ```prvTcpSocketHelper``` called by ```prvConnectHelperWithRetry``` returns error code, if Socket cannot be allocated and the error code is stored in variable ```xResult```.

Previously this variable was never addressed after an error is received. Thus, the code runs in an infinite loop(which is not expected). This is the reason for failure of ```AFQP_SOCKETS_Threadsafe_DifferentSocketsDifferentTasks``` test-case.

The expected behaviour is; the function should retry obtaining a socket repetitively after a specific delay.

In prvConnectHelperWithRetry, check is added to retry until valid socket is received.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -448,6 +448,8 @@ static BaseType_t prvConnectHelperWithRetry( volatile Socket_t * pxSocket,
         if( SOCKETS_INVALID_SOCKET == *pxSocket )
         {
             xResult = SOCKETS_SOCKET_ERROR;
+            tcptestFAILUREPRINTF( ( "Failed to allocate Socket Descriptor \r\n" ) );
+            break;
         }
 
         /* Set the appropriate socket options for the destination. */


yuhui-zheng:
Hi shubhamkulkarni97,

Thank you for raising this PR. I agree with you that if unable to create socket in ```prvTcpSocketHelper()```, ```prvConnectHelperWithRetry()``` will stuck in do-while loop. 

Though, I think there may be a better fix than exponential backoff and retry -- break out of the loop when failing to create socket without retry. Reasoning as below.

The original code looks like
```
        /* Create the socket. */
        *pxSocket = prvTcpSocketHelper( pxSocketIsAllocated );

        if( SOCKETS_INVALID_SOCKET == *pxSocket )
        {
            xResult = SOCKETS_SOCKET_ERROR;
        }
```
And in ```prvTcpSocketHelper()``` we create a socket by calling ```SOCKETS_Socket()```. The reason the system may fail to create socket can be:
- fail to allocate memory to hold internal structure.
- board does not support more than a number of sockets.

In the former case, though memory can be freed in between retries (assuming dynamic allocation is supported), system is already struggling with insufficient memory. In the latter case, retry would not make too much difference if system ran out of available socket, given when testing we don't have other things running. 

Given above, I think adding a ```break;``` statement is sufficient. Test will fail as expected. Let me know your thoughts.

dcgaws:
What is the root cause of the socket allocation failure? Would be helpful to know if it’s RAM, max connections, or etc.

shubhamkulkarni97:
> What is the root cause of the socket allocation failure? Would be helpful to know if it’s RAM, max connections, or etc.

If a closed socket is in time-wait state, PCB for that socket is not cleared. Hence, if the total count of active PCBs exceed the maximum count, socket allocation fails.

shubhamkulkarni97:
> Hi shubhamkulkarni97,
> 
> Thank you for raising this PR. I agree with you that if unable to create socket in `prvTcpSocketHelper()`, `prvConnectHelperWithRetry()` will stuck in do-while loop.
> 
> Though, I think there may be a better fix than exponential backoff and retry -- break out of the loop when failing to create socket without retry. Reasoning as below.
> 
> The original code looks like
> 
> ```
>         /* Create the socket. */
>         *pxSocket = prvTcpSocketHelper( pxSocketIsAllocated );
> 
>         if( SOCKETS_INVALID_SOCKET == *pxSocket )
>         {
>             xResult = SOCKETS_SOCKET_ERROR;
>         }
> ```
> 
> And in `prvTcpSocketHelper()` we create a socket by calling `SOCKETS_Socket()`. The reason the system may fail to create socket can be:
> 
> * fail to allocate memory to hold internal structure.
> * board does not support more than a number of sockets.
> 
> In the former case, though memory can be freed in between retries (assuming dynamic allocation is supported), system is already struggling with insufficient memory. In the latter case, retry would not make too much difference if system ran out of available socket, given when testing we don't have other things running.
> 
> Given above, I think adding a `break;` statement is sufficient. Test will fail as expected. Let me know your thoughts.

If a closed socket is in time-wait state, it will be cleared in finite amount of time. Hence, I think adding a retry with some delay may help in allocating a socket.

Let me know your views.

If you still think adding a `break;` is a better alternative, I can update the PR.

yuhui-zheng:
I understand the motivation for the retry. If this part of the code is in application, I would've agreed with the fix. However, ```prvConnectHelperWithRetry()``` is part of the test. And with every function call to ```prvConnectHelperWithRetry()``` in the same file, there is either an assert followed or check followed. Thus, breaking out the ```do-while``` loop early in the test, will make test more efficient. Another fact for our tests is that, while test is running, there is nothing else running. Thus, if we have trouble creating more than one socket already, then the test is solidly failed. 

So, yes, I'd vote for adding a ```break;``` statement after socket creation failure in the original code.

-------- more details --------
- Without any fix, this particular test case will stay in do-while loop upon failure to create more than one socket. The test group will timeout with other tests behind this test case unfinished. 
- With shubhamkulkarni97's original proposal for adding retries-timeout after socket creation failure, the test group will break out of do-while loop after retries. The test group will fail faster. 
- With ```break;``` after socket creation failure, the test fails immediately. This is given during test, nothing else is running.

shubhamkulkarni97:
@yuhui-zheng I have pushed required changes. Please take a look.

PR #1533, opened 2019-11-19
FreeRTOS+TCP: Revisiting the network interface for Renesas RX65N

htibosch:
Description
-----------
The NetworkInterface driver for Renesas' RX65N works fine to do a demo and some tests.
In order to make it ready for a real-life application, I would like to propose some changes:

- The detection of the PHY Link Status should be repaired, the code contains a few typo's.
~~~c
     if (ETHER_SUCCESS == ret)
     {
         if ((NULL != cb_func.pcb_func) && (FIT_NO_FUNC != cb_func.pcb_func))
         {
             cb_arg.channel = channel;
             cb_arg.event_id = ETHER_CB_EVENT_ID_LINK_ON;
             (*cb_func.pcb_func)((void *) &cb_arg);
         }
+        lchng_flag[channel] = ETHER_FLAG_ON_LINK_ON;
     }
     else
     {
         /* When PHY auto-negotiation is not completed */
         transfer_enable_flag[channel] = ETHER_FLAG_OFF;
-        lchng_flag[channel] = ETHER_FLAG_ON_LINK_ON;
+        lchng_flag[channel] = ETHER_FLAG_ON_LINK_OFF;
     }
~~~

This code occurs two times.

- The PHY Link Status can be used for two purposes:
  1. No packets should be sent as long as the status is low 
     `xNetworkInterfaceOutput()` will now check `xPHYLinkStatus`.
  2. Inform the user that the apparatus is unplugged from the LAN

- The PHY is now polled from a separate task called `check_ether_link()`.
  I would propose to make this a sub-task of `prvEMACDeferredInterruptHandlerTask()`.

- The PHY was checked every second, which is disturbing for the communication. I found a decrease in performance of almost 50%.
In other FreeRTOS+TCP drivers, the following approach is used:
  1. As long as packets are received, **assume** that the Link Status is high.
  2. Only when no packets come in for N seconds, poll the PHY Link Status.
  3. Once the Link Status is low, polling can be done frequently.
The above method is controlled by two macros:
~~~c
    /* Check if the LinkSStatus in the PHY is still high after 2 seconds of not
    receiving packets. */
    #define PHY_LS_HIGH_CHECK_TIME_MS   2000

    /* Check if the LinkSStatus in the PHY is still low every second. */
    #define PHY_LS_LOW_CHECK_TIME_MS    1000
~~~

- `xNetworkInterfaceInitialise()` is supposed to return `pdPASS` when both the peripheral has been initialised successfully, and when the PHY Link Status is up.
The current code blocks for ever until the Link Status is up. Therefore, the IP-task will also block for ever, and any +TCP API will block as well.
Let `prvEMACDeferredInterruptHandlerTask()` poll the PHY, and let `xNetworkInterfaceInitialise()` return `pdPASS` as soon as a link is detected (just like has been done in other NetworkInterfaces ).

- In `RX\ether_callback.c`, there is a callback mechanism for PHY changes. This has been prepared but not finished. I added a call to a new function `prvLinkStatusChange()` to let the NetworkInterface know that the PHY Link Status has changed.

- When developing and testing a +TCP application, it is important to keep an eye on a few resources:
  1. The number of available Network Buffers
  2. The free space in the working queue of the IP-task.
  3. The free HEAP space.
Many existing drivers check this and issue logging about it.

- Two tasks were started with a priority of `configMAX_PRIORITIES`. That should have been `configMAX_PRIORITIES-1` of course.  The priority of the IP-task was correctly set at:
~~~c
    #define ipconfigIP_TASK_PRIORITY                   ( configMAX_PRIORITIES - 2 )
~~~
Beside that, the tasks were given 100 words of stack space, which would soon lead to a stack overflow after a few additions of code.

- The maximum number of Network Buffers is important. It is defined as:
~~~c
    #define ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS    ETHER_CFG_EMAC_TX_DESCRIPTORS
~~~
  Which evaluates to 1. Any application with this setting will drop incoming packets. Chances are also big that tasks will wait for an available Network buffer, thus losing time.
I would rather propose a fair assesment, like e.g. :
~~~c
#define ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS    8
~~~

- Also the number of DMA RX-descriptors was very low :
~~~c
#define ETHER_CFG_EMAC_RX_DESCRIPTORS               (1)
~~~
I tested with a value of 8 and sliding TCP windows ( `ipconfigUSE_TCP_WIN = 1` ).

Before and after all changes, I tested both the demo and the test application and also added a IPERF3 server to measure performance. The IPERF3 code is not part of this PR.

Also tested: unplug and plug the LAN during a TCP connection. The connection resumed without a problem.

Uncrustify: I will apply it in a separate commit.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
@@ -16,18 +16,20 @@
 *
 * Copyright (C) 2018 Renesas Electronics Corporation. All rights reserved.
 ***********************************************************************************************************************/
+
 /***********************************************************************************************************************
 * File Name    : NetworkInterface.c
 * Device(s)    : RX
 * Description  : Interfaces FreeRTOS TCP/IP stack to RX Ethernet driver.
 ***********************************************************************************************************************/
+
 /***********************************************************************************************************************
 * History : DD.MM.YYYY Version  Description
 *         : 07.03.2018 0.1     Development
 ***********************************************************************************************************************/
 
 /***********************************************************************************************************************
-Includes   <System Includes> , "Project Includes"
+*  Includes   <System Includes> , "Project Includes"
 ***********************************************************************************************************************/
 #include <stdint.h>
 #include <stdio.h>
@@ -39,7 +41,7 @@ Includes   <System Includes> , "Project Includes"
 #include "task.h"
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_IP_Private.h"
-//#include "FreeRTOS_DNS.h"
+/*#include "FreeRTOS_DNS.h" */
 #include "NetworkBufferManagement.h"
 #include "NetworkInterface.h"
 
@@ -47,28 +49,29 @@ Includes   <System Includes> , "Project Includes"
 #include "r_pinset.h"
 
 /***********************************************************************************************************************
- Macro definitions
+ * Macro definitions
  **********************************************************************************************************************/
-#define ETHER_BUFSIZE_MIN 60
+#define ETHER_BUFSIZE_MIN    60
 
-#if defined(BSP_MCU_RX65N) || defined(BSP_MCU_RX64M) || defined(BSP_MCU_RX71M)
-#if ETHER_CFG_MODE_SEL == 0
-#define R_ETHER_PinSet_CHANNEL_0()  R_ETHER_PinSet_ETHERC0_MII()
-#elif ETHER_CFG_MODE_SEL == 1
-#define R_ETHER_PinSet_CHANNEL_0()  R_ETHER_PinSet_ETHERC0_RMII()
-#endif
-#elif defined(BSP_MCU_RX63N)
-#if ETHER_CFG_MODE_SEL == 0
-#define R_ETHER_PinSet_CHANNEL_0()  R_ETHER_PinSet_ETHERC_MII()
-#elif ETHER_CFG_MODE_SEL == 1
-#define R_ETHER_PinSet_CHANNEL_0()  R_ETHER_PinSet_ETHERC_RMII()
-#endif
-#endif
+#if defined( BSP_MCU_RX65N ) || defined( BSP_MCU_RX64M ) || defined( BSP_MCU_RX71M )
+    #if ETHER_CFG_MODE_SEL == 0
+        #define R_ETHER_PinSet_CHANNEL_0()    R_ETHER_PinSet_ETHERC0_MII()
+    #elif ETHER_CFG_MODE_SEL == 1
+        #define R_ETHER_PinSet_CHANNEL_0()    R_ETHER_PinSet_ETHERC0_RMII()
+    #endif
+#elif defined( BSP_MCU_RX63N )
+    #if ETHER_CFG_MODE_SEL == 0
+        #define R_ETHER_PinSet_CHANNEL_0()    R_ETHER_PinSet_ETHERC_MII()
+    #elif ETHER_CFG_MODE_SEL == 1
+        #define R_ETHER_PinSet_CHANNEL_0()    R_ETHER_PinSet_ETHERC_RMII()
+    #endif
+#endif /* if defined( BSP_MCU_RX65N ) || defined( BSP_MCU_RX64M ) || defined( BSP_MCU_RX71M ) */
 
 #ifndef PHY_LS_HIGH_CHECK_TIME_MS
-    /* Check if the LinkSStatus in the PHY is still high after 2 seconds of not
-    receiving packets. */
-    #define PHY_LS_HIGH_CHECK_TIME_MS   2000
+
+/* Check if the LinkSStatus in the PHY is still high after 2 seconds of not
+ * receiving packets. */
+    #define PHY_LS_HIGH_CHECK_TIME_MS    2000
 #endif
 
 #ifndef PHY_LS_LOW_CHECK_TIME_MS
@@ -77,13 +80,13 @@ Includes   <System Includes> , "Project Includes"
 #endif
 
 /***********************************************************************************************************************
- Private global variables and functions
+ * Private global variables and functions
  **********************************************************************************************************************/
 typedef enum
 {
-    eMACInit,                               /* Must initialise MAC. */
-    eMACPass,                               /* Initialisation was successful. */
-    eMACFailed,                             /* Initialisation failed. */
+    eMACInit,   /* Must initialise MAC. */
+    eMACPass,   /* Initialisation was successful. */
+    eMACFailed, /* Initialisation failed. */
 } eMAC_INIT_STATUS_TYPE;
 
 static TaskHandle_t ether_receive_check_task_handle = 0;
@@ -93,17 +96,19 @@ static BaseType_t xPHYLinkStatus;
 static BaseType_t xReportedStatus;
 static eMAC_INIT_STATUS_TYPE xMacInitStatus = eMACInit;
 
-static int16_t SendData( uint8_t *pucBuffer, size_t length );
-static int InitializeNetwork(void);
-static void prvEMACDeferredInterruptHandlerTask( void *pvParameters );
-static void clear_all_ether_rx_discriptors(uint32_t event);
+static int16_t SendData( uint8_t * pucBuffer,
+                         size_t length );
+static int InitializeNetwork( void );
+static void prvEMACDeferredInterruptHandlerTask( void * pvParameters );
+static void clear_all_ether_rx_discriptors( uint32_t event );
 
-int32_t callback_ether_regist(void);
-void EINT_Trig_isr(void *);
-void get_random_number(uint8_t *data, uint32_t len);
+int32_t callback_ether_regist( void );
+void EINT_Trig_isr( void * );
+void get_random_number( uint8_t * data,
+                        uint32_t len );
 
 void prvLinkStatusChange( BaseType_t xStatus );
-static void prvMonitorResources(void);
+static void prvMonitorResources( void );
 
 /***********************************************************************************************************************
  * Function Name: xNetworkInterfaceInitialise ()
@@ -132,6 +137,7 @@ BaseType_t xNetworkInterfaceInitialise( void )
             /* Indicate that the MAC initialisation succeeded. */
             xMacInitStatus = eMACPass;
         }
+
         configPRINTF( ( "InitializeNetwork returns %s\n", ( xMacInitStatus == eMACPass ) ? "OK" : " Fail" ) );
     }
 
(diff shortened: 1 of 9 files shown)


htibosch:
Uncrustified the most important file: `freertos_plus_tcp\source\portable\NetworkInterface\RX\NetworkInterface.c`

dan4thewin:
@htibosch , I see that one of the files you modified is within the Renesas FIT tree.  That tree appears to have v1.15, while it appears that v1.22 is available from Renesas.  It seems possible we'll have to upgrade at some point, and I'm concerned that your changes will be lost.  Can you explore what would be required to have these changes made in the upstream?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
```diff
@@ -127,14 +168,19 @@ BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxDescript
...
+    if( xPHYLinkStatus != 0 )
     {
-        vReleaseNetworkBufferAndDescriptor( pxDescriptor );
-        return pdFALSE;
+        if(SendData( pxDescriptor->pucEthernetBuffer, pxDescriptor->xDataLength ) >= 0)
+        {
+            xReturn = pdTRUE;
+            /* Call the standard trace macro to log the send event. */
+            iptraceNETWORK_INTERFACE_TRANSMIT();
+        }
```
dan4thewin: The old code always releases in the error case.  Should this behavior be maintained for backwards compatibility?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
```diff
@@ -168,20 +261,35 @@ static void prvEMACDeferredInterruptHandlerTask( void *pvParameters )
...
+    vTaskSetTimeOutState( &xPhyTime );
+    xPhyRemTime = pdMS_TO_TICKS( PHY_LS_LOW_CHECK_TIME_MS );
+
+    configPRINTF( ( "Deferred Interrupt Handler Task started\n" ) );
+    xTaskToNotify = ether_receive_check_task_handle;
+
     for( ;; )
     {
-        xTaskToNotify = ether_receive_check_task_handle;
+        prvMonitorResources();
```
dan4thewin: Seems like the monitoring should be conditionally compiled, so that it may be omitted from production code.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
```diff
@@ -252,6 +360,42 @@ static void prvEMACDeferredInterruptHandlerTask( void *pvParameters )
...
+            xPhyRemTime = pdMS_TO_TICKS( PHY_LS_HIGH_CHECK_TIME_MS );
+            /* Indicate that the Link Status is high, so that
+            xNetworkInterfaceOutput() can send packets. */
+            if( xPHYLinkStatus == 0 )
+            {
+                xPHYLinkStatus = 1;
+                FreeRTOS_printf( ( "prvEMACHandlerTask: PHY LS assume %d\n", xPHYLinkStatus ) );
+            }
+        }
+        else if( ( xTaskCheckForTimeOut( &xPhyTime, &xPhyRemTime ) != pdFALSE ) || ( FreeRTOS_IsNetworkUp() == pdFALSE ) )
```
dan4thewin: I don't understand the relationship between FreeRTOS_IsNetworkUp(), xPHYLinkStatus, and xReportedStatus.  Does the R_ETHER_LinkProcess(0) change what FreeRTOS_IsNetworkUp() will return?  Could you please add a comment to this section to give clarity?

vendors/renesas/boards/rx65n-rsk/aws_tests/config_files/FreeRTOSIPConfig.h
```diff
@@ -324,4 +317,7 @@ void vApplicationMQTTGetKeys( const char ** ppcRootCA,
                               const char ** ppcClientCert,
                               const char ** ppcClientPrivateKey );
 
+/* Let NetworkInterface check the free space in the working queue for the IP-task. */
+#define ipconfigCHECK_IP_QUEUE_SPACE			1
```
dan4thewin: I'd like to see the config for aws_tests and for aws_demos be the same unless there is a compelling reason to have differences.  Are the changes I see to the demo config above already in this file?

htibosch:
> @htibosch , I see that one of the files you modified is within the Renesas FIT tree.
> That tree appears to have v1.15, while it appears that v1.22 is available from Renesas.

Hello @dan4thewin , I was aware of that. I changed as little as possible to "their" source files, but I do believe this is repairing 2 errors.
I just downloaded their v1.22 : `r_ether_rx.c` now has v1.17, and it still contains the typo's.
It would be nice if we can communicate with Renesas' developers.

Or, a different way: use their function `phy_get_link_status()`, which is reliable. It is reading the PHY registers to retrieve the Link Status. I will investigate and test that.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
htibosch: No it should not be maintained, it was incorrect. A network buffer may only be released if `xReleaseAfterSend` is true.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
htibosch: You are right about that. In a release version, you don't want logging that nobody looks at.

I propose to make it conditional on `ipconfigHAS_PRINTF`. In a release version, that macro will be undefined or `0`.
More than 50% of the issues reported on the forum can be resolved by just monitoring these three resources. So I think it's important to have it by default, just like `ipconfigHAS_PRINTF` is `1` by default.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
htibosch: > Does the R_ETHER_LinkProcess(0) change what FreeRTOS_IsNetworkUp() will return?

Yes indirectly. But once the network is up, it will remain up for ever, even when disconnected from the LAN.
It is possible to bring the network down (`eNetworkDownEvent`), but that is rarely used.

It is not your shortcoming if you get lost here. I will try show it with some pseudo C code:
This is how IP-ask gets the network up:

~~~c
    for( ;; )
    {
        if( xNetworkInterfaceInitialise() == pdPASS )
        {
            /* The peripheral has been initialised, and the Link Status is up. */
            break;
        }
        /* 3 seconds sleep. */
        vTaskDelay( ipINITIALISATION_RETRY_DELAY );
    }
    /* From now on, FreeRTOS_IsNetworkUp() will return true. */
    xNetworkUp = true;
~~~

Within `NetworkInterface.c`, I introduced:
~~~c
BaseType_t xPHYLinkStatus;    /* The Link Status. */
BaseType_t xReportedStatus;   /* The Link Status as reported by the callback. */
~~~

Now here is the PHY polling:

~~~c
    /* Renesas driver: */
    void R_ETHER_LinkProcess (uint32_t channel)
    {
        /* Check the Link Status. */
        /* Call a user function */
    }
~~~

~~~c
    /* In NetworkInterface.c : the PHY call-back function: */
    void prvLinkStatusChange( BaseType_t xStatus )
    {
        if( xReportedStatus != xStatus )
        {
            configPRINTF( ( "prvLinkStatusChange( %d )\n", xStatus ) );
            xReportedStatus = xStatus;
        }
    }
~~~

and in the driver task:

~~~c
    for( ;; )
    {
        ulTaskNotifyTake();

        /* Read RX stream. */
        xBytesReceived = R_ETHER_Read_ZC2( 0, &buffer );

        /* PHY polling. */
        R_ETHER_LinkProcess( 0 );

        if( xPHYLinkStatus != xReportedStatus )
        {
            xPHYLinkStatus = xReportedStatus;
            FreeRTOS_printf( ( "prvEMACHandlerTask: PHY LS now %d\n", xPHYLinkStatus ) );
        }
    }
~~~

PS. I will replace all occurrences of `configPRINTF()` with `FreeRTOS_printf()`.

htibosch: > I will replace all occurrences of configPRINTF() with FreeRTOS_printf().

Just done so in this commit: "Use FreeRTOS_printf for debug logging in stead of configPRINTF"

`FreeRTOS_printf()` will normally become an empty statement in release versions.

vendors/renesas/boards/rx65n-rsk/aws_tests/config_files/FreeRTOSIPConfig.h
htibosch: In the existing repo, the are several differences between the two IP-config files.
After PR #1533, there will be 2 differences left:

~~~c
 /* Each TCP socket has a circular buffers for Rx and Tx, which have a fixed
  * maximum size.  Define the size of Rx buffer for TCP sockets. */
-#define ipconfigTCP_RX_BUFFER_LENGTH                   ( 1460 * 8 )
+#define ipconfigTCP_RX_BUFFER_LENGTH                   ( 1460 * 3 )
 
 /* Define the size of Tx buffer for TCP sockets. */
-#define ipconfigTCP_TX_BUFFER_LENGTH                   ( 1460 * 8 )
+#define ipconfigTCP_TX_BUFFER_LENGTH                   ( 1460 * 3 )
~~~

I don't know why the testing version needs larger TCP stream buffers. I have no problem if aws_demo gets the same configuration as aws_test.
Although I do not understand a difference in clock speeds:

~~~c
-#define configCPU_CLOCK_HZ				( 120000000UL )
+#define configCPU_CLOCK_HZ				( 96000000UL )
-#define configPERIPHERAL_CLOCK_HZ		( 60000000UL )
+#define configPERIPHERAL_CLOCK_HZ		( 48000000UL )
~~~
Found in:
  vendors\renesas\boards\rx65n-rsk\aws_tests\config_files\FreeRTOSConfig.h
  vendors\renesas\boards\rx65n-rsk\aws_demos\config_files\FreeRTOSConfig.h

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/RX/NetworkInterface.c
htibosch: Just committed: "Monitor resources only if FreeRTOS_printf is defined"
The function `prvMonitorResources()` will only be declared and called if `ipconfigHAS_PRINTF` is defined as `1`.

htibosch:
Hi , Mike Whalen, the Renesas demo and test project do not use lwip libraries (but FreeRTOS+TCP).
I don't mind rebasing it, but I'm afraid that it won't solve the problem that you encounter.
Thanks, Hein

mww-aws:
@htibosch , 

I think it is a matter of the project structure changing; even though you aren't modifying the files in question, they cause problems when we combine your branch with the baseline. 

Here is what we are doing:
  git clone git@github.com:aws/amazon-freertos.git --recurse-submodules
  git config --add remote.origin.fetch +refs/pull/*/head:refs/remotes/origin/pr/*
  git fetch origin
  #  the commit below was the current head of the PR #1533 yesterday
  git checkout 9d3c7d5ff147d3da154f1c76357723b6ef31e5c0       
 
This is the approach chosen by the Padstone team (e.g., Mark Tuttle) to bring in the PR.  When I do the checkout, I see errors on lwip:

error: The following untracked working tree files would be overwritten by checkout:", "\tlibraries/3rdparty/lwip/.gitattributes", "\tlibraries/3rdparty/lwip/.gitignore", "\tlibraries/3rdparty/lwip/CHANGELOG", "\tlibraries/3rdparty/lwip/CMakeLists.txt", "\tlibraries/3rdparty/lwip/COPYING", "\tlibraries/3rdparty/lwip/FEATURES", "\tlibraries/3rdparty/lwip/FILES", "\tlibraries/3rdparty/lwip/README", "\tlibraries/3rdparty/lwip/UPGRADING", "\tlibraries/3rdparty/lwip/doc/FILES", "\tlibraries/3rdparty/lwip/doc/NO_SYS_SampleCode.c", "\tlibraries/3rdparty/lwip/doc/ZeroCopyRx.c", "\tlibraries/3rdparty/lwip/doc/contrib.txt", "\tlibraries/3rdparty/lwip/doc/doxygen/generate.bat", ...

HirokiIshiguro:
Hello, I am RX MCUs device drivers developer in Renesas.
Thank you for reviewing RX65N Ethernet driver code. I have informed this issue to my team.

PR #1540, opened 2019-11-21
Add UART assisted tests

qiutongs:
Description
-----------

- Add UART assisted tests

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
+++ b/libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3
+# Amazon FreeRTOS V1.2.3
+# Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# http://aws.amazon.com/freertos
+# http://www.FreeRTOS.org
+
+
+import serial
+import csv
+import argparse
+import os, sys
+import re
+import socket
+from time import sleep
+import threading
+
+parentdir = os.path.dirname(os.getcwd())
+sys.path.insert(0, parentdir)
+from test_iot_test_template import test_template
+
+
+class TestUartAssisted(test_template):
+    """
+    Test class for uart tests.
+    """
+
+    rpi_output_file = "./uart_res.txt"
+
+    def __init__(self, serial, ip, login, pwd, csv_handler):
+        self._func_list = [self.test_IotUartWriteReadAsync,
+                           self.test_IotUartBaudChange,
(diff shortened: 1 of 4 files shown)


qiutongs:
> we are not consistent on license, I thought we are using "Amazon FreeRTOS Common IO V1.0.0" for all files related to common-io, not "Amazon FreeRTOS V1.2.3"

I think you are right. Let me update them.

libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def test_IotUartWriteAsync(self):
+        """
+        Test body of uart write async test. Return pass if expected message is received
+        :return: 'Pass' or 'Fail'
+        """
+        pi_result = None
+        dut_result = None
+        dut_res_pattern = re.compile(r"AFQP_AssistedIotUARTWriteAsync\W\sPASS")
+        pi_res_pattern = re.compile(r"\b(?:\\xaa){198}\b")
```
cmcgcc: why it is 198 here? Look like dut is sending out 200 bytes.

libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
```diff
@@ -0,0 +1,259 @@
...
+        self.run_shell_script(
+            " ".join([self.shell_script, self._ip, self._login, self._pwd, "-c"])
+        )
+
+
+# unit test
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('-i', '--ip', nargs=1, default=['10.53.30.75'], help='ip address of rpi')
```
cmcgcc: Could you take away the content in the default?

qiutongs: Oh, sorry. Yes, I will remove

qiutongs:
> I think the test coverage is still limited. The write read test could be separated as write sync and read sync, in case both are flipped. The read async, duplex sync/async should be covered as well. It is OK for this pr, but we should improve later.

Yeah, I agree. What in this PR is just a start for UART. Because assist tests is not the current priority. We will improve it later.

libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
lundinc2: https://github.com/aws/amazon-freertos/blob/2049152156e5ffc7d2d775da29b527487cc857b1/libraries/abstractions/common_io/test/test_iot_uart.c#L418 We lose one byte as the UART assisted tests use a '\n' byte to signal the write is done. In my PR I added a '\n' to the end of the buffer, but this change was lost when reviving that PR. We need to set cpBufferLarge[ testIotUART_BUFFER_LENGTH_LARGE  ] = '\n' and change this  to 199 bytes. I believe this was 198 as I was using '\r\n' but '\n' is sufficient and easier.

qiutongs: I will update it to 200

libraries/abstractions/common_io/test/test_scripts/uart/test_iot_uart_test.py
```diff
@@ -0,0 +1,260 @@
...
+    def test_IotUartWriteAsync(self):
+        """
+        Test body of uart write async test. Return pass if expected message is received
+        :return: 'Pass' or 'Fail'
+        """
+        pi_result = None
+        dut_result = None
+        dut_res_pattern = re.compile(r"AFQP_AssistedIotUARTWriteAsync\W\sPASS")
+        # 200 corresponds to the config testIotUART_BUFFER_LENGTH_LARGE on the DUT
+        pi_res_pattern = re.compile(r"\b(?:\\xaa){200}\b")
```
lundinc2: This should be 199, as the last byte in the buffer is '\n', not 0xAA

qiutongs: The size of buffer is 201: testIotUART_BUFFER_LENGTH_LARGE

lundinc2: The write is only 200 bytes. https://github.com/aws/amazon-freertos/blob/2049152156e5ffc7d2d775da29b527487cc857b1/libraries/abstractions/common_io/test/test_iot_uart.c#L428

qiutongs: Oh, You are right. Let me update the test to be:

```
 lWrite = iot_uart_write_async( xUartHandle, cpBufferLarge, testIotUART_BUFFER_LENGTH_LARGE + 1 );
```

cmcgcc: I do not see too much sense to modify c code just to keep 200 here. Let us just keep consistent with c code. It does not matter 199 or 200 here.

lundinc2: I agree with @cmcgcc

qiutongs: ok

PR #1548, opened 2019-11-25
fix UART non-assisted tests

qiutongs:
Description
-----------

1. Remove test case: AFQP_IotUARTWriteSyncReadAsyncLoopbackTest
This test works with a loopback wiring setup. When it does a sync write, all the bytes will go to Rx register and cause an overrun error. This test only works if there is a hardware FIFO buffer on the UART. 
2. AFQP_IotUARTWriteReadAsyncWithCallback is fixed for similar reason above.
3. Removed some local unused variable.

Checklist:
----------

- [ ] My code is Linted.


qiutongs:
> If we are removing AFQP_IotUARTWriteSyncReadAsyncLoopbackTest() how are we getting coverage for the loopback test.

Because of hardware limitation, I think the loopback test cannot cover all cases in nature. This write_sync + read_async is a good  example. That behavior has to be test with assisted tests.

PR #1550, opened 2019-11-26
Add missing file type for esp32 binary.

lundinc2:
Description
-----------
Cmake for esp is currently configured to use the IDF_PROJECT_EXECUTABLE, for building, flashing and more. If the suffix ".elf" is added, the logic for these steps create failures. A binary is created for example, the demo project, as "aws_demo". This binary is actually a .elf file, but lacks the file type ending. This makes it harder to find the elf file that is flashed. My proposed solution is to create a copy of the flashed binary and append ".elf" to it. This makes it easier for the user to find the elf file.

Checklist:
----------

- [ ] My code is Linted.


tgsong:
producing 2 identical files one with extension and one without is also kinda confusing. Should we instead ask Espressif to just get the output name of the executable target and convert it to bin? A more proper fix is to just call `set(CMAKE_EXECUTABLE_SUFFIX ".elf")` like we did in other boards. Then Espressif can use `get_target_property(exe_name ${IDF_PROJECT_EXECUTABLE} OUTPUT_NAME)` to get whatever executable file CMake would generate.

lundinc2:
Okay that fix looks good to me. I tried set(CMAKE_EXECUTABLE_SUFFIX ".elf") some weeks ago, and the error messages I saw made me think it would be take some work to change. Thanks!

tgsong:
@lundinc2 @cobusve Since this changes goes into esp-idf, we should check with espressif first before we merge. @mahavirj Can you review if this is right way?

lundinc2:
@tgsong I have reached out to @mahavirj and they are investigating this.

shubhamkulkarni97:
@lundinc2 I tried applying the changes. Application builds successfully and is flashed without errors. But, `idf.py monitor` doesn't work, as the value of app_elf is not modified to `.elf` in `project_description.json`. We are trying to figure out the reason, will keep you posted on this.

lundinc2:
@shubhamkulkarni97 thanks for taking a look.

lundinc2:
@shubhamkulkarni97 any update on this?

shubhamkulkarni97:
@lundinc2 Sorry for late reply.
We have not yet resolved the issue. Will notify you as soon as we find the solution.

shubhamkulkarni97:
@lundinc2 I'm attaching a patch which would help fix the missing file type issue. This fix will also be included in IDF.

[cmake-use-elf-as-suffix-thru-CMAKE_EXECUTABLE_SUFFIX.patch.zip](https://github.com/aws/amazon-freertos/files/4096050/cmake-use-elf-as-suffix-thru-CMAKE_EXECUTABLE_SUFFIX.patch.zip)

lundinc2:
@shubhamkulkarni97 thank you kindly for this patch!

PR #1553, opened 2019-11-26
BLE AFR: Make 16 bit Service UUID as default while advertising with Service Data (Bluedroid & NimBLE port)

prasad-alatkar:
Modify the default 128 bit Service UUID to 16 bit Service UUID in Service Data.

Description
-----------
In `iot_ble_hal_gap.c`, use 16 bit Service UUID inside the Service Data as a default while setting advertisement data with `prvSetAdvData`. 

Checklist:
----------

- [ ] My code is Linted.


prasad-alatkar:
Yes, it should be. In fact I will add another commit to this very same PR for the same.

PR #1556, opened 2019-11-26
PKCS#11 Multithread stack fix

lundinc2:
Increased mulithreaded task stack sizes for the pkcs#11 tests as the current config led to stack overflows on ESP.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
+++ b/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
@@ -344,7 +344,7 @@ static MultithreadTaskParams_t xGlobalTaskParams[ pkcs11testMULTI_THREAD_TASK_CO
 /*-----------------------------------------------------------*/
 /* Stack size of each task. This can be configured in iot_test_pkcs11_config.h. */
 #ifndef pkcs11testMULTI_TASK_STACK_SIZE
-    #define pkcs11testMULTI_TASK_STACK_SIZE    ( configMINIMAL_STACK_SIZE * 4 )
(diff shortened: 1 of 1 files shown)


libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -344,7 +344,7 @@ static MultithreadTaskParams_t xGlobalTaskParams[ pkcs11testMULTI_THREAD_TASK_CO
 /*-----------------------------------------------------------*/
 /* Stack size of each task. This can be configured in iot_test_pkcs11_config.h. */
 #ifndef pkcs11testMULTI_TASK_STACK_SIZE
-    #define pkcs11testMULTI_TASK_STACK_SIZE    ( configMINIMAL_STACK_SIZE * 4 )
+    #define pkcs11testMULTI_TASK_STACK_SIZE    ( configMINIMAL_STACK_SIZE * 8 )
```
dcgaws: This is a huge delta. Was this failing only on ESP?

lundinc2: Yes, I believe this was passing on Windows.

alexa-noxon: I feel like we should change this to be a number rather than a multiple of configMINIMAL_STACK_SIZE, since that varies so much platform to platform.   Could we use what you learned about the stack requirements on esp and pick an appropriate value here?

lundinc2: @dcgaws suggested I move this to a config file. I agree I can move this to iot_test_pkcs11_config.h.

PR #1557, opened 2019-11-26
Fix bad pointer de-reference in HTTPS Client demo

sarenameas:
Description
-----------
* Fix `_requestPool.pRequestDatas->scheduled = true;` to `_requestPool.pRequestContexts[ reqIndex ].scheduled = true;` as found by third-party static analysis tool.
* Changed all "requestDatas" to "requestContexts" this is because the "datas" is weird as data is already plural. "Context" postfixed to a variable name is vague to use, but better than postfixing "Data". These variables have a collection of information that get updated during the callbacks, so they are essentially "callbackContexts". 
* Fix spellings of words in all HTTPS Client library and demo files.
* Add more debug logging to the HTTPS Client library code and the async download demo.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/https/iot_demo_https_s3_download_async.c
+++ b/demos/https/iot_demo_https_s3_download_async.c
@@ -200,7 +200,7 @@
 /**
  * @brief This structure defines data consumed and kept track of in the async callbacks per request.
  */
-typedef struct _requestData
+typedef struct _requestContext
 {
     char pRangeValueStr[ RANGE_VALUE_MAX_LENGTH ]; /**< @brief This string is generated outside of the callback context and used to set the request Range header within the callback context. */
     int reqNum;                                    /**< @brief This is the current request number in the static pool of request memory. */
@@ -383,11 +383,11 @@ static int _getFreeRequestIndex( void )
 /*-----------------------------------------------------------*/
 
 /**
- * @brief Clear data for the request in preparation for the next request to use.
+ * @brief Clear the request context in preparation for the next request to use.
  *
  * User buffers and body buffers are zeroed out by the HTTPS Client library.
  */
-static void _clearRequestData( int i )
+static void _clearRequestContext( int i )
 {
     memset( &_requestPool.pRequestContexts[ i ], 0, sizeof( _requestContext_t ) );
 }
@@ -405,7 +405,7 @@ static void _freeRequestIndex( int i )
 {
     IotMutex_Lock( &( _inUseRequestsMutex ) );
     _pInUseRequests[ i ] = false;
-    _clearRequestData( i );
+    _clearRequestContext( i );
     IotMutex_Unlock( &( _inUseRequestsMutex ) );
 }
 
@@ -415,9 +415,16 @@ static void _freeRequestIndex( int i )
  * @brief Free all requests in the pool that have been marked as scheduled, except
  * for the exception index.
  *
+ * This function is used during reconnection to free any possible requests that
+ * might still have been scheduled to the HTTPS Client library to run but have
+ * not been sent yet. Reconnection happens during the _readReadyCallback().
+ * The response's memory cannot be used until the _responseCompleteCallback() return,
+ * so during reconnection, we need to indicate that all memory except the current
+ * response is to be marked as free.
+ *
  * @param[in] exception Request index to not free. This is -1 for no exceptions.
  */
-static void _freeAllScheduledRequests( int exception )
+static void _freeScheduledRequests( int exception )
 {
     /* Index into the request pool. */
     int i = 0;
@@ -430,7 +437,7 @@ static void _freeAllScheduledRequests( int exception )
             ( _requestPool.pRequestContexts[ i ].scheduled == true ) )
         {
             _pInUseRequests[ i ] = false;
-            _clearRequestData( i );
+            _clearRequestContext( i );
         }
     }
 
@@ -523,7 +530,9 @@ static bool _getNextIncompleteRange( uint32_t * currentRange )
     bool found = false;
     /* An intermediate bitMask calculation for checking if the currentRange is already downloaded or not. */
     uint32_t bitMask = 0;
-    /* An intermediate byteOffset into the downloadedBitmap to check if the currentRange is already downloaded or not. */
+
+    /* An intermediate byteOffset into the downloadedBitmap and scheduledBitmap to check if the currentRange is
+     * already downloaded or already scheduled. */
     uint32_t byteOffset = 0;
 
     /* A count from 0 to check if the total ranges is reached because, in the logic below, currentRange will wrap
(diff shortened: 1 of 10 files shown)


demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -412,9 +412,12 @@ static void _freeRequestIndex( int i )
 /*-----------------------------------------------------------*/
 
 /**
- * @brief Free all requests in the pool that have been marked as scheduled.
+ * @brief Free all requests in the pool that have been marked as scheduled, except
+ * for the exception index.
```
dcgaws: "Free all requests, with the possible exception of one of them," seems like unusual logic. Could this comment further expand on what the exceptional cases are?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -519,12 +523,12 @@ static bool _getNextIncompleteRange( uint32_t * currentRange )
     bool found = false;
     /* An intermediate bitMask calculation for checking if the currentRange is already downloaded or not. */
     uint32_t bitMask = 0;
-    /* An intermediate byteOffset into the downlaodedBitmap to check if the currentRange is already downloaded or not. */
+    /* An intermediate byteOffset into the downloadedBitmap to check if the currentRange is already downloaded or not. */
```
dcgaws: And scheduledBitmap.

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -519,12 +523,12 @@ static bool _getNextIncompleteRange( uint32_t * currentRange )
...
     uint32_t byteOffset = 0;
 
     /* A count from 0 to check if the total ranges is reached because, in the logic below, currentRange will wrap
      * around to check the start of the bitmap. */
-    int32_t rangeCheckCount = 0;
+    uint32_t rangeCheckCount = 0;
 
     do
     {
```
dcgaws: Can you please write a comment on what the first several lines of the below block are doing?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -774,12 +780,17 @@ static void _readReadyCallback( void * pPrivData,
             }
             else
             {
+                IotLogInfo( "Successfully reconnected to the S3 server." );
```
dcgaws: The name of the demo notwithstanding, couldn't this code be used to download from any HTTP server that supports range requests?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -774,12 +780,17 @@ static void _readReadyCallback( void * pPrivData,
...
+
                 /* For all requests from the pool that have been scheduled. We want to free them so they will get
-                 * rescheduled by the main application. */
-                _freeAllScheduledRequests();
+                 * rescheduled by the main application. When a request from the pool is set to unused its associated
```
dcgaws: The requests are being rescheduled because there was a reconnect?

demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -888,8 +899,10 @@ static void _errorCallback( void * pPrivData,
                             IotHttpsReturnCode_t rc )
 {
     ( void ) reqHandle;
-    char * pRangeValueStr = ( ( _requestData_t * ) ( pPrivData ) )->pRangeValueStr;
-    IotLogError( "An error occurred during asynchronous operation with code: %d", pRangeValueStr, rc );
+    ( void ) respHandle;
+
+    char * pRangeValueStr = ( ( _requestContext_t * ) ( pPrivData ) )->pRangeValueStr;
```
dcgaws: The API requires a non-NULL pPrivData?

demos/https/iot_demo_https_s3_upload_async.c
```diff
@@ -236,6 +236,8 @@ static void _responseCompleteCallback( void * pPrivData,
                                        IotHttpsReturnCode_t rc,
                                        uint16_t status )
 {
+    ( void ) rc;
```
dcgaws: Could rc indicate an error even when status indicates OK?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -232,10 +232,10 @@ static int _httpParserOnBodyCallback( http_parser * pHttpParser,
                                       size_t length );
 
 /**
- * @brief Callback from http-parser to indicate it reached the end of the HTTP response messsage.
+ * @brief Callback from http-parser to indicate it reached the end of the HTTP response message.
  *
  * The end of the message is signalled in a HTTP response message by another "\r\n" after the final header line, with no
- * entity body; or it is singalled by "\r\n" at the end of the entity body.
+ * entity body; or it is singaled by "\r\n" at the end of the entity body.
```
dcgaws: signalled

sarenameas: signaled since 1776

demos/https/iot_demo_https_s3_upload_async.c
sarenameas: status is HTTP response status. rc is library error. So yes, but this return code is a repeat of the rc from _errorCallback(). Error callback is called first: https://docs.aws.amazon.com/freertos/latest/lib-ref/https/https_design.html#Asynchronous_Callback_Order

demos/https/iot_demo_https_s3_download_async.c
sarenameas: API does not require a non-null pPrivData, but I utilize the private data to give a useful log.

demos/https/iot_demo_https_s3_download_async.c
sarenameas: Any requests that are still in the queue to be sent, but haven't yet sent. These need to be rescheduled. Reconnecting clears the internal request and response queue.

demos/https/iot_demo_https_s3_download_async.c
sarenameas: I suppose so.

demos/https/iot_demo_https_s3_download_async.c
sarenameas:
```
 * This function is used during reconnection to free any possible requests that 
 * might still have been scheduled to the HTTPS Client library to run but have 
 * not been sent yet. Reconnection happens during the _readReadyCallback(). 
 * The response's memory cannot be used until the _responseCompleteCallback() return, 
 * so during reconnection, we need to indicate that all memory except the current 
 * response is to be marked as free. 
```

demos/https/iot_demo_https_s3_download_async.c
sarenameas:
```
        /* Check if the current range in the file is already downloaded or is has already been scheduled. The 
         * bitmaps contains as many bits are there are blocks of byte ranges. The lines below help to index 
         * into the bitmaps and check if the bit representing that byte range is set. */
        bitMask = BITMASK( *currentRange );
        byteOffset = BYTE_OFFSET( *currentRange );

        /* If this range has not been downloaded yet and has not been schedule, then return the current range. */
        if( ( ( _fileDownloadInfo.downloadedBitmap[ byteOffset ] & bitMask ) == ( ( uint8_t ) 0 ) ) &&
            ( ( _fileDownloadInfo.scheduledBitmap[ byteOffset ] & bitMask ) == ( ( uint8_t ) 0 ) ) )
        {
            break;
        }
```

demos/https/iot_demo_https_s3_download_async.c
sarenameas: Updated everywhere to just say "server" instead of "s3 server"

libraries/c_sdk/standard/https/src/iot_https_client.c
dcgaws: That's cool. But I don't think that's what it says.

demos/https/iot_demo_https_s3_download_async.c
dcgaws: "bitmaps contains"

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: oh oops thanks

demos/https/iot_demo_https_s3_download_async.c
sarenameas: thanks

PR #1558, opened 2019-11-26
Add ESP32-WROOM-32SE to AFR

lundinc2:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


mahavirj:
@lundinc2 Recently we had refactored `common` components that can be shared across `applications` and/or `boards` for ESP32. Components from `vendors/espressif/boards/esp32/aws_{demos/tests}/application_code/espressif_code/` were moved to `vendors/espressif/boards/esp32/components`. This PR essentially duplicates those components again. This creates lot of trouble (in the form of code duplication) when user wants to create their own application, also maintenance issue when fixes needs to propagated at multiple places.

My recommendation is as follows:
ESP32-WROOM-32SE should not be treated as separate board, rather its pretty much compatible with ESP32-WROOM-32 module. You may create new `component` as `vendors/espressif/boards/esp32/components/pkcs11` and drop in `Kconfig` where user has configurable option to select hardware PKCS, thus allowing to pick appropriate PKCS layer in build system. PR that way will be extremely small and clean.

vendors/espressif/esp-idf/tools/cmake/idf_functions.cmake
```diff
@@ -113,6 +113,8 @@ function(idf_set_global_compile_options)
 
     list(APPEND compile_options "-Wall"
                                 "-Werror=all"
+                                "-Wno-error=char-subscripts"
```
mahavirj: Can you please add this to relevant component? Modifying `vendors/espressif/esp-idf` is not good idea, it means divergence from our internal release, it will be difficult for us to keep track of such changes.

vendors/espressif/manifest.cmake
```diff
@@ -1,5 +1,7 @@
 set(
     AFR_MANIFEST_SUPPORTED_BOARDS
+    esp32_plus_ecc608a_devkitc
+    esp32_plus_ecc608a_wrover_kit
```
mahavirj: Up until now we do not have `wrover_kit` based on `ecc608a`, so you might want to remove this entry

vendors/espressif/esp-idf/tools/cmake/idf_functions.cmake
lundinc2: okay, I'll see what I can do

vendors/espressif/manifest.cmake
lundinc2: Okay, I will remove these entries.

lundinc2: @mahavirj Removed these entries.

vendors/espressif/esp-idf/tools/cmake/idf_functions.cmake
lundinc2: @mahavirj I have moved these to our CMakelist.txt and only added these compiler flags to the relevant files.

vendors/espressif/boards/esp32/CMakeLists.txt
```diff
@@ -84,6 +90,40 @@ endif()
...
+            "${ecc608a_dir}/lib/hal/hal_freertos.c"
+            "${ecc608a_dir}/lib/hal/hal_esp32_i2c.c"
+            "${ecc608a_dir}/lib/hal/hal_esp32_timer.c"
+            "${ecc608a_dir}/lib/hal/atca_hal.h"
+    )
+
+    afr_glob_src(ecc608a_src DIRECTORY ${ecc608a_dir}/lib)
+    list(REMOVE_ITEM ecc608a_src "${ecc608a_dir}/lib/atca_cfgs.c")
+
+    set(ecc608_include "${ecc608a_dir}/lib")
```
ravibhagavandas: Never used ?

lundinc2: This config is set in a separate file, iot_pkcs11_secure_element.c

rajneeshsinha:
@mahavirj ATECC608A chip support is available with amazon freeRTOS SDK where we can use for ESP32 platform?

I read the below thread on amazon forums and they are requesting to wait for this port to merge with amazon freeRTOS master branch. Is this port now merged?

https://forums.aws.amazon.com/thread.jspa?messageID=925190

Also, do we have any guidelines available how to use this ATECC608A chip with ESP32 platform?

PR #1564, opened 2019-11-27
Fix warnings in latest version of Amazon FreeRTOS

shubhamkulkarni97:
Warnings were observed while building latest version of Amazon FreeRTOS. Fix includes CFLAGS changes in CMakeLists.txt, component.mk, Makefile and ESP-IDF specific changes.

Description
-----------

Checklist:
----------

- [x] My code is Linted.


vendors/espressif/esp-idf/components/driver/rtc_module.c
```diff
@@ -995,7 +995,7 @@ esp_err_t touch_pad_filter_start(uint32_t filter_period_ms)
     }
     if (s_touch_pad_filter->timer == NULL) {
         s_touch_pad_filter->timer = xTimerCreate("filter_tmr", filter_period_ms / portTICK_PERIOD_MS, pdFALSE,
-        NULL, touch_pad_filter_cb);
+        NULL, (void(*)(TimerHandle_t))touch_pad_filter_cb);
```
cobusve: It is generally better to let the compiler check the prototype, if we need a cast here there is probably something wrong with the prototype. What was the warning here?

I would really like to avoid casting a function pointer at all cost!

vendors/espressif/esp-idf/components/esp32/esp_adapter.c
```diff
@@ -523,24 +522,24 @@ wifi_osi_funcs_t g_wifi_osi_funcs = {
...
     ._queue_send_from_isr = queue_send_from_isr_wrapper,
     ._queue_send_to_back = queue_send_to_back_wrapper,
     ._queue_send_to_front = queue_send_to_front_wrapper,
     ._queue_recv = queue_recv_wrapper,
-    ._queue_msg_waiting = uxQueueMessagesWaiting,
-    ._event_group_create = xEventGroupCreate,
-    ._event_group_delete = vEventGroupDelete,
-    ._event_group_set_bits = xEventGroupSetBits,
-    ._event_group_clear_bits = xEventGroupClearBits,
+    ._queue_msg_waiting = (uint32_t(*)(void *))uxQueueMessagesWaiting,
```
cobusve: Again I would really like to avoid casting function pointers at all cost, why do we get a warning here?

vendors/espressif/esp-idf/components/esp32/hwcrypto/sha.c
```diff
@@ -195,7 +195,7 @@ static bool esp_sha_lock_engine_common(esp_sha_type sha_type, TickType_t ticks_t
 
 void esp_sha_unlock_engine(esp_sha_type sha_type)
 {
-    SemaphoreHandle_t *engine_state = sha_get_engine_state(sha_type);
```
cobusve: Should we be changing the ESP-IDF code like this?

vendors/espressif/esp-idf/components/ethernet/emac_main.c
```diff
@@ -770,7 +770,7 @@ static bool emac_link_check_timer_init(void)
                               (CONFIG_EMAC_CHECK_LINK_PERIOD_MS / portTICK_PERIOD_MS),
                               pdTRUE,
                               NULL,
-                              emac_link_check_func);
+                              (void(*) (TimerHandle_t)) emac_link_check_func);
```
cobusve: Again I do not think we should change the esp-idf code

vendors/espressif/esp-idf/components/driver/rtc_module.c
shubhamkulkarni97: The warning here was `warning: passing argument 5 of 'xTimerCreate' from incompatible pointer type [-Wincompatible-pointer-types]`. This is because `TimerHandle_t` is declared as `typedef struct tmrTimerControl * TimerHandle_t;` in AmazonFreeRTOS, while it is declared as `typedef void * TimerHandle_t;` in esp-idf which results in a warning.

vendors/espressif/esp-idf/components/esp32/esp_adapter.c
shubhamkulkarni97: Similar is the case with `._queue_msg_waiting`. Parameter to `_queue_msg_waiting` is `void *`, while the parameter to the function `uxQueueMessagesWaiting` is `const QueueHandle_t`.

vendors/espressif/esp-idf/components/esp32/hwcrypto/sha.c
shubhamkulkarni97: Return type of the function `sha_get_engine_state` is `SemaphoreHandle_t`. Hence, the variable `engine_state` should be of the type `SemaphoreHandle_t`.

vendors/espressif/esp-idf/components/driver/rtc_module.c
mahavirj: @cobusve To keep compatibility of version of FreeRTOS used in ESP-IDF this change is required (https://github.com/espressif/esp-idf/blob/release/v3.3/components/freertos/include/freertos/timers.h#L118). This is approved change and will reflect in public ESP-IDF github repository soon.

cobusve: Ok Mahavir, with your assurance that this is not diverging from the esp-idf release I can approve it for you.

PR #1570, opened 2019-11-28
fix for ESP32 Nimble: prvBTSetAdvData assumes exactly 1 UUID to be ad…

### LINKED ISSUE #1569: ESP32 Nimble: prvBTSetAdvData assumes exactly 1 UUID to be advertised
**Describe the bug**
In `iot_ble_hal_gap.c`, function `prvBTSetAdvData()` never uses the `xNbServices` parameter and assumes there is exactly 1 UUID to be advertised if `pxServiceUuid` is not NULL.
At the same time, `_setAdvData` always gives a value to `pxServiceUuid` and sets `xNbServices` as 0 if there are no UUIDs to advertise.
This means that trying to advertise 0 UUIDs (e.g. to save advertisement space) does not work. Trying to advertise 2 UUIDs would also not work.

**System information**
- Which hardware board or part numbers? ESP32
- IDE used. Does not matter.
- Operating System. Does not matter.
- Version of Amazon FreeRTOS. master
- Project: Custom Application

**Expected behavior**
The xNBServices should be used to advertise the requested number of UUIDs.

**To reproduce**

Just use a custom advertising callback like the following:

**Code to reproduce the bug**
```
void IotBle_SetCustomAdvCb(IotBleAdvertisementParams_t *pAdvParams,
                           IotBleAdvertisementParams_t *pScanParams)
{
  pAdvParams->setScanRsp          = false;
  pAdvParams->includeTxPower      = false;
  pAdvParams->name.xType          = BTGattAdvNameShort;
  pAdvParams->name.ucShortNameLen = IOT_BLE_DEVICE_SHORT_LOCAL_NAME_SIZE;
  pAdvParams->appearance          = IOT_BLE_ADVERTISING_APPEARANCE;
  pAdvParams->minInterval         = IOT_BLE_ADVERTISING_CONN_INTERVAL_MIN;
  pAdvParams->maxInterval         = IOT_BLE_ADVERTISING_CONN_INTERVAL_MAX;
  pAdvParams->serviceDataLen      = 0;
  pAdvParams->pServiceData        = NULL;
  pAdvParams->manufacturerLen     = 0;
  pAdvParams->pManufacturerData   = NULL;
  pAdvParams->pUUID1 = (BTUuid_t *)&_advUUID;
  pAdvParams->pUUID2 = NULL;

  /// use the scan response to fit the full name of the device
  pScanParams->setScanRsp          = true;
  pScanParams->includeTxPower      = false;
  pScanParams->name.xType          = BTGattAdvNameComplete;
  pScanParams->name.ucShortNameLen = 0; // does not matter
  pScanParams->appearance          = IOT_BLE_ADVERTISING_APPEARANCE;
  pScanParams->minInterval         = IOT_BLE_ADVERTISING_CONN_INTERVAL_MIN;
  pScanParams->maxInterval         = IOT_BLE_ADVERTISING_CONN_INTERVAL_MAX;
  pScanParams->serviceDataLen      = 0;
  pScanParams->pServiceData        = NULL;
  pScanParams->manufacturerLen     = 0;
  pScanParams->pManufacturerData   = NULL;
  pScanParams->pUUID1 = NULL;
  pScanParams->pUUID2 = NULL;
}
```

**Additional context**
I have a patch that I will post soon as a PR

matteoscordino:
fix for ESP32 Nimble: prvBTSetAdvData assumes exactly 1 UUID to be advertised

Description
-----------
fixes #1569 :
In iot_ble_hal_gap.c, function prvBTSetAdvData() never uses the xNbServices parameter and assumes there is exactly 1 UUID to be advertised if pxServiceUuid is not NULL.
At the same time, _setAdvData always gives a value to pxServiceUuid and sets xNbServices as 0 if there are no UUIDs to advertise.
This means that trying to advertise 0 UUIDs (e.g. to save advertisement space) does not work. Trying to advertise 2 UUIDs would also not work.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
@@ -634,27 +634,54 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
       {
         if( pxServiceUuid[i].ucType == eBTuuidType16 )
         {
-          uuid16[fields.num_uuids16].u.type = BLE_UUID_TYPE_16;
-          uuid16[fields.num_uuids16].value = pxServiceUuid->uu.uu16;
-          fields.uuids16 = uuid16;
-          fields.num_uuids16++;
-          fields.uuids16_is_complete = 1;
+          if(fields.num_uuids16==0)
+          {
+            uuid16[fields.num_uuids16].u.type = BLE_UUID_TYPE_16;
+            uuid16[fields.num_uuids16].value = pxServiceUuid->uu.uu16;
+            fields.uuids16 = uuid16;
+            fields.num_uuids16++;
+            fields.uuids16_is_complete = 1;
+          }
+          else
+          {
+            /// there are more than 1 service of this type, but as per bluetooth core specification supplement v8,
+            /// page 10, only one service UUID per type is allowed. Mark this as incomplete.
+            fields.uuids16_is_complete = 0;
+          }
         }
         else if( pxServiceUuid[i].ucType == eBTuuidType32 )
         {
-          uuid32[fields.num_uuids32].u.type = BLE_UUID_TYPE_32;
-          uuid32[fields.num_uuids32].value = pxServiceUuid->uu.uu32;
-          fields.uuids32 = uuid32;
-          fields.num_uuids32++;
-          fields.uuids32_is_complete = 1;
+          if(fields.num_uuids32==0)
+          {
+            uuid32[fields.num_uuids32].u.type = BLE_UUID_TYPE_32;
+            uuid32[fields.num_uuids32].value = pxServiceUuid->uu.uu32;
+            fields.uuids32 = uuid32;
+            fields.num_uuids32++;
+            fields.uuids32_is_complete = 1;
+          }
+          else
+          {
+            /// there are more than 1 service of this type, but as per bluetooth core specification supplement v8,
+            /// page 10, only one service UUID per type is allowed. Mark this as incomplete.
+            fields.uuids32_is_complete = 0;
+          }
         }
         else if( pxServiceUuid[i].ucType == eBTuuidType128 )
         {
-          uuid128[fields.num_uuids128].u.type = BLE_UUID_TYPE_128;
-          memcpy( uuid128[fields.num_uuids128].value, pxServiceUuid->uu.uu128, sizeof( pxServiceUuid->uu.uu128 ) );
-          fields.uuids128 = uuid128;
-          fields.num_uuids128++;
-          fields.uuids128_is_complete = 1;
+          if(fields.num_uuids128==0)
+          {
+            uuid128[fields.num_uuids128].u.type = BLE_UUID_TYPE_128;
+            memcpy( uuid128[fields.num_uuids128].value, pxServiceUuid->uu.uu128, sizeof( pxServiceUuid->uu.uu128 ) );
+            fields.uuids128 = uuid128;
+            fields.num_uuids128++;
+            fields.uuids128_is_complete = 1;
+          }
+          else
+          {
+            /// there are more than 1 service of this type, but as per bluetooth core specification supplement v8,
+            /// page 10, only one service UUID per type is allowed. Mark this as incomplete.
+            fields.uuids128_is_complete = 0;
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


matteoscordino:
made the change you suggested, it was a quick one

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -535,12 +535,13 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
...
                             size_t xNbServices )
 {
+    const size_t c_max_uuid_per_adv_message = 2;
     struct ble_hs_adv_fields fields;
     const char * name;
     int xESPStatus;
-    ble_uuid16_t uuid16 = { 0 };
-    ble_uuid32_t uuid32 = { 0 };
-    ble_uuid128_t uuid128 = { 0 };
+    ble_uuid16_t uuid16[c_max_uuid_per_adv_message];
```
ravibhagavandas: You don't need the array and `c_max_uuid_per_adv_message` as there can be at max one uuid of each type.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -624,32 +624,65 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
...
     }
 
-    if( pxServiceUuid != NULL )
+    fields.num_uuids16=0;
+    fields.num_uuids32=0;
+    fields.num_uuids128=0;
+    for (size_t i = 0; i < xNbServices; i++)
     {
-        if( pxServiceUuid->ucType == eBTuuidType16 )
+      if( &(pxServiceUuid[i]) != NULL )
```
ravibhagavandas: This check is not needed as  &(pxServiceUuid[i]) can never be NULL.

It makes sense to check for `if( pxServiceUuid != NULL )` before iterating through the array here `for (size_t i = 0; i < xNbServices; i++)`

matteoscordino: `pxServiceUuid[i]` can never be NULL only if we assume the calling code has no bugs.
I think the check is needed as a form of defensive coding, because the caller might have passed a wrong `xNbServices` (i.e. one that is inconsistent with `pxServiceUuid`).
e.g. if the caller passed no services in `pxServiceUuid`, but set `xNbServices` to `1`, removing this check would result in an out-of-bounds memory access.

ravibhagavandas: @matteoscordino 
Array out of bounds cannot be verified using the method you mentioned. Example if array `pxServiceUuid` points to address say `0x1000` and each element size is `K` then  `&pxServiceUuid[N+1]` will be pointing to address `(0x1000 + ( N+1) * K` and this will never be `null(0)`. This would work if `pxServiceUuid is NULL` and user passes `xNbServices > 0` but a better check would be `pxServiceUuid  != NULL` in terms of readability.

matteoscordino: As you prefer, I will make the change.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -624,32 +624,65 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
         fields.svc_data_uuid128_len = usServiceDataLen;
     }
 
+    fields.num_uuids16=0;
+    fields.num_uuids32=0;
+    fields.num_uuids128=0;
     if( pxServiceUuid != NULL )
     {
-        if( pxServiceUuid->ucType == eBTuuidType16 )
+      for (size_t i = 0; i < xNbServices; i++)
```
gkwicker: The code in this loop contains 3 copies of identical operations that differ only in the type they are applied to:

- Set type
- Set value
- Set UUID
- Increment Number of UUIDs of this type
- Set the is_complete flag for this type of UUID

Please refactor this code so that this pattern is only done once in the loop.  An easy way to do this would be with some small utility functions for the UUID16, UUID32, and UUID128 types that this loop calls based on the value of `pxServiceUuid[i].ucType`.

matteoscordino:
Fair point, that code is not optimal.
However, it was there in that form before the PR, it's not code I wrote,
and it is not affected by the code I wrote to fix the issue that I reported.

I think refactoring is out of scope of this PR.


Le mar. 10 déc. 2019 à 23:05, Gary Wicker <notifications@github.com> a
écrit :

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
matteoscordino: Fair point, that code is not optimal.
However, the structure of the code was there in that form before the PR, it's not code I wrote,
and its structure is not affected by the code I wrote to fix the issue that I reported.


I think refactoring that code is out of scope of this PR, whose initial intent was fixing a bug I found in AFR.

PR #1582, opened 2019-12-05
 Ensure make_proof_makefiles.py runs on python 3.8 f4f947b 

karkhaz:
This commit fixes the make_proof_makefiles.py script, which uses a class that was removed from the standard library in python 3.8. The script now detects which python version it's running on, and uses the correct class either way.


karkhaz:
the second commit is the interesting one; the first one is just whitespace fixes.

markrtuttle:

markrtuttle:

PR #1585, opened 2019-12-06
CBMC Windows fixes

karkhaz:
One of the commits in this PR improves the performance of cbmc-viewer on windows by actually excluding the correct directories. The other fixes the common Makefile used by CBMC proofs so that it can also be used on Windows.


markrtuttle:

PR #1586, opened 2019-12-06
OTA fixes

tgsong:
Description
-----------

This PR addresses some issues in current OTA code,

- `aws_iot_ota_mqtt.h` should be included when MQTT is selected in data **or control**.
- When HTTP send fails, the error callback will be invoked directly, we need to set the error status in http downloader. Before this case is not handled.
- Now the "connection" field is always checked when receiving a HTTP response to see if we need to connect. Before if error happens during the response callback routine, this will not be executed.
- Fix the way that momentum is handled in OTA downloads. We should always increase the momentum after sending the request. Before the momentum is only increased if the request function returns failure. However, both MQTT and HTTP requests are async and it could return success but we still fail to receive the response and the OTA agent would just stuck here forever. Instead we can just rely on `prvProcessDataHandler` to reset the momentum, which is called after we successfully get and process a response.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -949,39 +949,31 @@ static OTA_Err_t prvRequestDataHandler( OTA_EventData_t * pxEventData )
         /* Start the request timer. */
         prvStartRequestTimer( otaconfigFILE_REQUEST_WAIT_MS );
 
-        /* Request data blocks. */
-        xErr = xOTA_DataInterface.prvRequestFileBlock( &xOTA_Agent );
-
-        if( xErr != kOTA_Err_None )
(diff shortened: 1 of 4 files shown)


libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -952,36 +952,26 @@ static OTA_Err_t prvRequestDataHandler( OTA_EventData_t * pxEventData )
         /* Request data blocks. */
         xErr = xOTA_DataInterface.prvRequestFileBlock( &xOTA_Agent );
```
ravibhagavandas: Should this be within `if( xOTA_Agent.ulRequestMomentum < otaconfigMAX_NUM_REQUEST_MOMENTUM )` ?
Otherwise we sent request for the last file block and still mark it as failed before the response for the last retry.

tgsong: Make sense, fixed.

PR #1592, opened 2019-12-07
BLE Hal API modifications

ravibhagavandas:
BLE Hal API modifications

Description
-----------

* Change advertisement timeout to 16 bit value as per spec, handle this in ESP and nordic.
* Implementation for pxConfigureMTU()
* Fix API pxAdvStatusCb() to pass adapter interface 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/bluedroid/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/bluedroid/iot_ble_hal_gatt_server.c
@@ -92,6 +92,8 @@ static BTStatus_t prvBTSendResponse( uint16_t usConnId,
                                      BTStatus_t xStatus,
                                      BTGattResponse_t * pxResponse );
 
+static BTStatus_t prvBTConfigureMtu( uint8_t ucServerIf, uint16_t usMtu );
+
 static BTGattServerInterface_t xGATTserverInterface =
 {
     .pxRegisterServer     = prvBTRegisterServer,
@@ -109,7 +111,8 @@ static BTGattServerInterface_t xGATTserverInterface =
     .pxStopService        = prvBTStopService,
     .pxDeleteService      = prvBTDeleteService,
     .pxSendIndication     = prvBTSendIndication,
-    .pxSendResponse       = prvBTSendResponse
+    .pxSendResponse       = prvBTSendResponse,
+    .pxConfigureMtu       = prvBTConfigureMtu
 };
 
 /*-----------------------------------------------------------*/
@@ -787,6 +790,22 @@ BTStatus_t prvBTSendResponse( uint16_t usConnId,
     }
 
     return xReturnStatus;
+}
+
+/*-----------------------------------------------------------*/
+
+static BTStatus_t prvBTConfigureMtu( uint8_t ucServerIf, uint16_t usMtu )
+{
+    BTStatus_t xStatus = eBTStatusSuccess;
+    esp_err_t xESPStatus;
+
+    xESPStatus = esp_ble_gatt_set_local_mtu( usMtu );
+    if( xESPStatus != ESP_OK )
+    {
+        xStatus = eBTStatusFail;
+    }
+
+    return xStatus;
 }
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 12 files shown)


vendors/espressif/boards/esp32/ports/ble/bluedroid/iot_ble_hal_gatt_server.c
```diff
@@ -791,6 +795,24 @@ BTStatus_t prvBTSendResponse( uint16_t usConnId,
 
 /*-----------------------------------------------------------*/
 
+static BTStatus_t prvBTConfigureMtu( uint8_t ucServerIf,
```
qiutongs: ucServerIf is not used. Is it expected?

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -667,6 +671,18 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
         /* fixme: set adv_params->high_duty_cycle accordingly */
     }
 
+    if( pxParams->bSetScanRsp == false )
+    {
+        if( pxParams->usTimeout != 0 )
+        {
+            lAdvDurationMS = ( int32_t ) ( pxParams->usTimeout * IOT_BLE_ADVERTISING_DURATION_UNITS ); /* Multiply pxParams->usTimeout value by units = 10ms */
```
qiutongs: Not understand this math computation. What is the unit here?

vendors/espressif/boards/esp32/ports/ble/bluedroid/iot_ble_hal_gatt_server.c
ravibhagavandas: Yes that's expected. Bluedroid API doesn't need the GATT server interface to set the local MTU so parameter is unused for the port.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
ravibhagavandas: This is as per BLE spec. The input integer value is multiplied by 10ms to get the actual timeout value for advertisement.

qiutongs: I see.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -43,6 +43,11 @@
 
 BTBleAdapterCallbacks_t xBTBleAdapterCallbacks;
 static struct ble_gap_adv_params xAdv_params;
+
+#define IOT_BLE_ADVERTISING_DURATION_UNITS    ( 10 )
```
gkwicker: 10 is a quantity rather than a unit of measure, suggest renaming to `IOT_BLE_ADVERTISING_DURATION_MS` since I think the units are milliseconds based on the comment on line 678 below.

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gap.c
gkwicker: The comment on line 678 is redundant and will become incorrect if the value of `IOT_BLE_ADVERTISING_DURATION_UNITS` is ever removed.  I would delete it.

vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_gap.c
```diff
@@ -838,14 +839,16 @@ ret_code_t prvBTAdvDataConvert( ble_advdata_t * xAdvData,
             else
             {
                 /* extract company_identifier from input manufacturer data
-                because nordic sdk has separate field for company_identifier */
-                for (uint8_t i = 0; i < iot_ble_hal_gapADVERTISING_COMPANY_ID_SIZE; i++) {
-                  companyId = companyId | ((uint16_t) pcManufacturerData[i] << (i * 8));
+                 * because nordic sdk has separate field for company_identifier */
+                for( uint8_t i = 0; i < iot_ble_hal_gapADVERTISING_COMPANY_ID_SIZE; i++ )
```
gkwicker: Would add comments about the size of the company ID string (I believe it's 4 bytes based on the rest of the code) and the size of `companyId` since it's not encoded in that variable name.

ravibhagavandas: Added comment regarding the company identifier field and the variable.

PR #1595, opened 2019-12-10
Add eWiFiSecurityWPA2_ent into Wi-Fi Security types

WenChieh-Feng:
Add eWiFiSecurityWPA2_ent into Wi-Fi Security types

Description
-----------
This commit adds eWiFiSecurityWPA2_ent into WIFISecurity structure to
let vendor be able to support Wi-Fi WPA2 enterprise security in the
future.

Checklist:
----------
- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/wifi/include/iot_wifi.h
+++ b/libraries/abstractions/wifi/include/iot_wifi.h
@@ -65,6 +65,7 @@ typedef enum
     eWiFiSecurityWEP,         /**< WEP Security. */
     eWiFiSecurityWPA,         /**< WPA Security. */
     eWiFiSecurityWPA2,        /**< WPA2 Security. */
+    eWiFiSecurityWPA2_ent,    /**< WPA2 Enterprise Security. */
     eWiFiSecurityNotSupported /**< Unknown Security. */
 } WIFISecurity_t;
 


sathishm6:
@dan4thewin @gkwicker  - For review.

libraries/abstractions/wifi/include/iot_wifi.h
```diff
@@ -65,6 +65,7 @@ typedef enum
     eWiFiSecurityWEP,         /**< WEP Security. */
     eWiFiSecurityWPA,         /**< WPA Security. */
     eWiFiSecurityWPA2,        /**< WPA2 Security. */
+    eWiFiSecurityWPA2_ent,    /**< WPA2 Enterprise Security. */
```
dan4thewin: Hi, @WenChieh-Feng .
I can understand having the above enum, but there's no code that makes use of this new enum.  Using this enum in a config may result in non-working wifi.

gkwicker: Also, where is the client configuration for WPA2-Enterprise configured and stored?  That isn't in this PR.

yuhui-zheng:
Adding another security type could potentially break our existing test case. An email is sent to the group to exercise the logic. A better approach, along with this change, is to update port specific WiFi implementation to check security type in API implementation before calling into driver.

libraries/abstractions/wifi/include/iot_wifi.h
sathishm6: Hi @gkwicker  - This PR is only for the changes that are required in the common interface (iot_wifi.h) to support enterprise security feature on Cypress platform. The actual implementation will be part of the code which will be under vendors/cypress folder, that will available in our upcoming release

sathishm6:
> Adding another security type could potentially break our existing test case. An email is sent to the group to exercise the logic. A better approach, along with this change, is to update port specific WiFi implementation to check security type in API implementation before calling into driver.

Agree. All the port specific implementation should be updated to validate the arguments and return error to the caller if the port doesn't support `eWiFiSecurityWPA2_ent`

yuhui-zheng:
@sathishm6 

While we are investigating what's the process to update all ports, could you elaborate why a new security type is needed? 

I'm on an other thread with Mike, and I'm assuming this PR is related to what he's working on. From the code he shared with me, I do not see ```eWiFiSecurityWPA2_ent``` is used in your ```iot_wifi.c```. (whd_tosecurity() and whd_fromsecurity() need to be updated in your ```iot_wifi.c``` at least, I did not check driver code.) If I'm understanding correctly, it is AP authenticates SSID/password/security type combination, not station. So a new security type does not have huge functional impact for what's coming, would that be a correct statement?

sathishm6:
> @sathishm6
> 
> While we are investigating what's the process to update all ports, could you elaborate why a new security type is needed?
> 
> I'm on an other thread with Mike, and I'm assuming this PR is related to what he's working on. From the code he shared with me, I do not see `eWiFiSecurityWPA2_ent` is used in your `iot_wifi.c`. (whd_tosecurity() and whd_fromsecurity() need to be updated in your `iot_wifi.c` at least, I did not check driver code.) If I'm understanding correctly, it is AP authenticates SSID/password/security type combination, not station. So a new security type does not have huge functional impact for what's coming, would that be a correct statement?

@yuhui-zheng  - eWiFiSecurityWPA2_ent security type is used for connecting the device to an enterprise security Wi-Fi network. This features is  going to be supported in Cypress platform as part of the upcoming AFR release we plan to make. With that release, the corresponding changes that are required to support enterprise security feature in iot_wifi.c will also show up. The primary reason for creating this PR is to overcome AmazonFreeRTOSSourceIntegrity failures we see in our Qual tests. In summary, without this change showing up on the AFR mainline, we can't pass AmazonFreeRTOSSourceIntegrity test. And if we don't pass AmazonFreeRTOSSourceIntegrity tests, we can't get through the Vendor Qual which will have all the relevant pieces of the software for enterprise security feature support.

yuhui-zheng:
@sathishm6  Given we have a pending release, I'm moving this PR to release-candidate branch to run with our infra. The hope is to get this enum into the coming release, thus to unblock the qualification. https://github.com/aws/amazon-freertos/pull/1630

As for API change or port change, I'm creating an internal issue to keep track of this.

PR #1596, opened 2019-12-10
test: Change I2S IOCTL test parameters

hesham-mahmoud:
This change provide commonly used I2S paramerts for this IOCTL
configuration test that we can assume it's safely going to be supported
by any I2S.

Why:
This is needed by MTK as the used parameters currently are not supported
by their chip.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_i2s.c
+++ b/libraries/abstractions/common_io/test/test_iot_i2s.c
@@ -512,6 +512,9 @@ TEST( TEST_IOT_I2S, AFQP_IotI2SIoctl )
     IotI2SClkPolarity_t xPolarity;
     uint32_t ulBusState;
     uint32_t ulRxTxBytes;
+    uint32_t ulFreqVals[2] = {8000, 16000};
+    uint32_t ulFrameLenVals[2] = {32, 64};
+    uint32_t ulDataLenVals[2] = {8, 16};
 
     /* Open i2s to initialize hardware */
     xI2SHandle = iot_i2s_open( ltestIotI2sInputInstance );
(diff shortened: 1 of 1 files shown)


hesham-mahmoud:
@cobusve  @vjjunnut @qiutongs
Please have a look.

libraries/abstractions/common_io/test/test_iot_i2s.c
```diff
@@ -572,7 +575,11 @@ TEST( TEST_IOT_I2S, AFQP_IotI2SIoctl )
        xTestConfig.xI2SWsPolarity = eI2SFallingEdge;
 
        /* Validate different frequencies */
-       xTestConfig.ulI2SFrequency++;
+       if( xTestConfig.ulI2SFrequency != ulFreqVals[0] ) {
+          xTestConfig.ulI2SFrequency = ulFreqVals[0];
+       } else {
+          xTestConfig.ulI2SFrequency = ulFreqVals[1];
+       }
```
qiutongs: I don't understand what this "if block" really wants to test.

hesham-mahmoud: If the running I2S Frequency is the value at index [0] of ulFreqVals, change it to the other one at index [1] before you apply it with SetConfig.
But if it's different, use the one at index [0].

We want to set a different parameter than the running configuration to see it is reflected back in the ioctl GetConfig call after we do SetConfig.

qiutongs: At line 578-582, you set the xTestConfig, but there is no eI2SSetConfig ioctl call after it.
At line 583, xTestConfig is passed into ioctl eI2SGetConfig and its value will be overwritten. Then line 578-582 is meaningless.

hesham-mahmoud: Good catch, many of the `eI2SGetConfig` used should be `eI2SSetConfig`. I'll update it.

hesham-mahmoud: @qiutongs , the commit is updated now. Please have a look.

PR #1600, opened 2019-12-11
Fix Infinite loop in BLE disconnect

leegeth:
Fix Infinite loop in BLE disconnect

Description
-----------
Shadow demo and MQTT demo on Nordic were not completing on Nordic.
The root cause was that the demo was completing but it was ending up in
an infinite loop in BLE disconnect logic and thus blocking the lower
priority logging task . This prevented logs of demo completion to be not
printed.
Solution is done by exiting the infinite loop in a condition when there
is error.

Checklist:
----------
- [ x] I have tested my changes.

### DIFF
--- a/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
+++ b/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
@@ -517,9 +517,10 @@ BTStatus_t IotBle_Off( void )
                 status = _BTInterface.pBTLeAdapterInterface->pxDisconnect( _BTInterface.adapterIf,
                                                                            &bdAddr,
                                                                            connId );
+
                 if( status != eBTStatusSuccess )
                 {
-                    IotLogError("Failed disconnect");
+                    IotLogError( "Failed disconnect" );
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/ble/src/iot_ble_gap.c
```diff
@@ -514,9 +514,15 @@ BTStatus_t IotBle_Off( void )
...
             if( pConnInfo != NULL )
             {
-                ( void ) _BTInterface.pBTLeAdapterInterface->pxDisconnect( _BTInterface.adapterIf,
+                status = _BTInterface.pBTLeAdapterInterface->pxDisconnect( _BTInterface.adapterIf,
                                                                            &bdAddr,
                                                                            connId );
+
+                if( status != eBTStatusSuccess )
+                {
+                    IotLogError( "Failed disconnect" );
```
qiutongs: Nitpick, can you also log the status?

leegeth: Done

dan4thewin:
To benefit future readers, it might be nice to have a comment about the life cycle of pConnInfo, i.e., does disconnect remove it from the linked list, etc.

leegeth:
> To benefit future readers, it might be nice to have a comment about the life cycle of pConnInfo, i.e., does disconnect remove it from the linked list, etc.

Sure. That's a good suggestion. I have talked to Ravi about it and we will create a SIM to handle it.

PR #1615, opened 2019-12-13
ulDNSHandlePacket() - check length before calculating payload size

gkwicker:
Detect integer overflow in `ulDNSHandlePacket()`.

-----------
This PR fixes an integer overflow case introduced in [an earlier commit](https://github.com/aws/amazon-freertos/commit/93c0aba73fe69cd3b067941d83d3d5d4dfebaabd#diff-ab35dd7e800d6d3ecb19dfa767aeacd4R1615) and detected by a failing [`ulDNSHandlePacket()`](https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/standard/freertos_plus_tcp/test/iot_test_freertos_tcp.c#L291) test.  See [here](https://github.com/aws/amazon-freertos/blob/release-candidate/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c#L1615) for another example of this overflow detection logic.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
@@ -916,18 +916,25 @@ for testing purposes, by the module iot_test_freertos_tcp.c
 uint32_t ulDNSHandlePacket( NetworkBufferDescriptor_t *pxNetworkBuffer )
 {
 DNSMessage_t *pxDNSMessageHeader;
-size_t uxPayloadSize = pxNetworkBuffer->xDataLength - sizeof( UDPPacket_t );
+size_t uxPayloadSize;
 
-	if( uxPayloadSize >= sizeof( DNSMessage_t ) )
-	{
-		pxDNSMessageHeader =
-			( DNSMessage_t * ) ( pxNetworkBuffer->pucEthernetBuffer + sizeof( UDPPacket_t ) );
+    /* Only proceed if the payload length indicated in the header
+    appears to be valid. */
+    if( pxNetworkBuffer->xDataLength >= sizeof( UDPPacket_t ) )
+    {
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -916,18 +916,25 @@ for testing purposes, by the module iot_test_freertos_tcp.c
...
+
+            /* The parameter pdFALSE indicates that the reply was not expected. */
+            prvParseDNSReply( ( uint8_t * ) pxDNSMessageHeader,
+                uxPayloadSize,
+                pdFALSE );
+        }
+    }
 
 	/* The packet was not consumed. */
 	return pdFAIL;
```
dan4thewin: Call me crazy, but why have a return value if it can never be anything other than pdFAIL?

gkwicker: Even though this handler never consumes the buffer, the IP stack expects handling code to return a status value indicating whether or not the buffer was consumed.

htibosch:
Hello @gkwicker I think that your change doesn't add anything.
Note my comment in the source code:

~~~c
The value of 'xDataLength' was proven to be at least the size of a
UDP packet in prvProcessIPPacket()
~~~

When `xProcessReceivedUDPPacket()` is called, you don't need to test this again:

~~~c
    if( pxNetworkBuffer->xDataLength >= sizeof( UDPPacket_t ) )
~~~

When you follow the flow of the program, you will see that the statement is always true.

Now for the call-back ( a rarely used feature in FreeRTOS+TCP ), I had forgotten to subtract the size of the UDP packet. I think that you can safely drop this PR #1615.

Thanks, Hein

PR #1616, opened 2019-12-13
Fix: GGD Test failures on ST

muneebahmed10:
Description
-----------
GGD_JSONRequestGetFile() would not receive the entire file
in one call on some boards. This adds a loop to receive the
file in chunks if the request was not completed.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
+++ b/libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
@@ -1030,7 +1030,7 @@ TEST( Full_GGD, JSONRequestGetFile )
             xStatus = prvGGD_JSONRequestGetFileLoop( ulBufferSize,
                                                      &ulByteRead,
                                                      &xJSONFileRetrieveCompleted,
-                                                     ulJSONFileSize - 1 ); /* Remove one byte to the expected JSON file size. */
+                                                     ulJSONFileSize - 1 ); /* Remove one byte from the expected JSON file size. */
         }
 
         TEST_ASSERT_EQUAL_INT32( pdFAIL, xStatus );
--- a/libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
+++ b/libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
@@ -47,6 +47,7 @@
 #define ggdTestJSON_PORT_ADDRESS_1         1234
 #define ggdTestJSON_PORT_ADDRESS_3         4321
 #define ggdTestLOOP_NUMBER                 10
+#define ggdTestMAX_REQUEST_LOOP_COUNT      2
 
 #define ggdJSON_FILE_GROUPID               "GGGroupId"
 #define ggdJSON_FILE_THING_ARN             "thingArn"
@@ -276,6 +277,7 @@ TEST( Full_GGD, GetIPandCertificateFromJSON )
     uint32_t ulBufferSize = testrunnerBUFFER_SIZE;
     uint32_t ulByteRead;
     BaseType_t xJSONFileRetrieveCompleted;
+    uint32_t ulRequestLoopCounter = 0;
 
     if( TEST_PROTECT() )
     {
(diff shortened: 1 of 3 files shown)


libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
```diff
@@ -318,12 +320,18 @@ TEST( Full_GGD, GetIPandCertificateFromJSON )
...
+
+            do
+            {
+                xStatus = GGD_JSONRequestGetFile( &xSocket,
+                                                  &cBuffer[ ulByteRead ],
+                                                  ulBufferSize - ulByteRead,
+                                                  &ulByteRead,
+                                                  &xJSONFileRetrieveCompleted,
+                                                  ulJSONFileSize );
+            } while( ( xStatus == pdPASS ) && ( xJSONFileRetrieveCompleted != pdTRUE ) && ( ulBufferSize - ulByteRead ) > 0 && ( ++ulRequestLoopCounter < ggdTestMAX_REQUEST_LOOP_COUNT ) );
```
gkwicker: Is there any way in which `ulBufferSize < ulByteRead`?  Imagine something isn't working properly – should we check to make sure this isn't the case before continuing in this loop?  See [this PR](https://github.com/aws/amazon-freertos/pull/1615) for an example of where not checking for this caused a problem.

libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
```diff
@@ -1004,12 +1023,17 @@ TEST( Full_GGD, JSONRequestGetFile )
...
-                                              cBuffer,
-                                              ulBufferSize,
-                                              &ulByteRead,
-                                              &xJSONFileRetrieveCompleted,
-                                              ulJSONFileSize - 1 ); /* Remove one byte to the expected JSON file size. */
+            ulRequestLoopCounter = 0;
+
+            do
+            {
+                xStatus = GGD_JSONRequestGetFile( &xSocket,
```
gkwicker: Two questions:

1. This logic looks a little different than the other two blocks, but they seem to be doing the same thing.  Is there any reason for the difference?
1. Since we have 3 blocks of code that look almost the same, could this operation be moved to a static utility function?

libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
muneebahmed10: Good catch, I think checking `ulBufferSize > ulByteRead` is more appropriate since unsigned will always be > 0

libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
muneebahmed10: 1. Good catch, it should be similar to the other three.
2. Created a static helper function per your suggestion.

libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
```diff
@@ -47,6 +47,7 @@
 #define ggdTestJSON_PORT_ADDRESS_1         1234
 #define ggdTestJSON_PORT_ADDRESS_3         4321
 #define ggdTestLOOP_NUMBER                 10
+#define ggdTestMAX_REQUEST_LOOP_COUNT      2
```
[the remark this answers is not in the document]
muneebahmed10: For the specific board this PR is meant to address, two iterations are required to retrieve the document from the socket. For every other board for which these tests passed, it was done in one go. I increased it to 3, and it can be increased in the future if a new board requires more iterations.

libraries/freertos_plus/aws/greengrass/test/aws_test_greengrass_discovery.c
```diff
@@ -102,6 +109,28 @@ TEST_GROUP_RUNNER( Full_GGD )
...
+
+    do
+    {
+        xStatus = GGD_JSONRequestGetFile( &xSocket,
+                                          &cBuffer[ *pulByteRead ],
+                                          ulBufferSize - *pulByteRead,
+                                          pulByteRead,
+                                          pxJSONFileRetrieveCompleted,
+                                          ulJSONFileSize );
+    } while( ( xStatus == pdPASS ) && ( *pxJSONFileRetrieveCompleted != pdTRUE ) && ( ulBufferSize > *pulByteRead ) && ( ++ulRequestLoopCounter < ggdTestMAX_REQUEST_LOOP_COUNT ) );
```
abhidixi11: Do we have to use ggdTestMAX_REQUEST_LOOP_COUNT ?  We are checking pxJSONFileRetrieveCompleted and ulBufferSize > pulByteRead, should these checks be sufficient to read the entire file and terminate if there is an error ?

muneebahmed10: With our code, no, and that's why there's no loop counter for the similar code in the [library](https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/aws/greengrass/src/aws_greengrass_discovery.c#L193). However, since this is part of a test, I thought that we should guard against the possibility that a change to `GGD_JSONRequestGetFile` might break it in a way such that those variables don't get updated properly.

PR #1618, opened 2019-12-13
fIX: Shutdown OTA agent so that it reconnects after max errors

pvyawaha:
Description
-----------

OTA shutdown so that OTA demo can reconnect when maximum retries are completed.

Checklist:
----------

- [ ] My code is Linted.


libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -916,7 +914,7 @@ static OTA_Err_t prvInitFileHandler( OTA_EventData_t * pxEventData )
             prvStopRequestTimer();
 
             /* Send close file event. */
```
ravibhagavandas: Update comment

pvyawaha: Updated

PR #1621, opened 2019-12-14
Avoid "Error: No OTA data buffers available" on LPC54018

dan4thewin:
Tune number of buffers.

Same as https://github.com/aws/amazon-freertos/pull/1620 but for release-candidate.


vendors/nxp/boards/lpc54018iotmodule/aws_demos/config_files/aws_ota_agent_config.h
```diff
@@ -100,7 +100,7 @@
  * This configurations parameter sets the maximum number of static data buffers used by
  * the OTA agent for job and file data blocks received.
  */
-#define otaconfigMAX_NUM_OTA_DATA_BUFFERS       2U
+#define otaconfigMAX_NUM_OTA_DATA_BUFFERS       10U
```
pvyawaha: Just noting that as the file block size is 4 KB we are reserving 40 Kb for data buffers.

PR #1623, opened 2019-12-14
Common ECC608A folder

lundinc2:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


libraries/abstractions/pkcs11/ecc608a/iot_pkcs11_secure_element.c
```diff
@@ -220,6 +216,17 @@ ATCAIfaceCfg cfg_ateccx08a_i2c_default = {
     .rx_retries             = 20
 };
 
+/** \brief default configuration for Kit protocol over the device's async interface */
+ATCAIfaceCfg cfg_ateccx08a_kithid_default = {
+    .iface_type         = ATCA_HID_IFACE,
+    .devtype            = ATECC508A,
```
dcgaws: Why do we need this?

lundinc2: This is the ecc608 configuration used the windows port. I made a copy paste error here, and will update with the correct HID configuration.

lundinc2: Also I think we should add a flag to make sure we don't waste memory compiling the other port's ecc608a configuration.

lundinc2: @dcgaws I fixed the copy paste error, and added macros to only compile the correct ecc608a config.

Issue #1555, opened 2019-11-26
Can not link the Xilinx demo project


htibosch:
**Describe the bug**
When compiling the Xilinx aws_demo project, linking fails, because of the command length.

**System information**
- Which hardware board or part numbers?
  Xilinx Zynq

- IDE used
  Xilinx Software Development Kit, release 2019.1

- Operating System
  Windows 10

- Version of Amazon FreeRTOS
  Todays version

- Project
  aws_demo

**To reproduce**

1.Import the aws_demo project into the work space
2. Press on Build All

**Additional context**
The problem is caused by the fact that cmd.exe can only accept a parameter length of 8192 bytes.
There seems to be [a solution](https://mcuoneclipse.com/2015/03/29/solving-the-8192-character-command-line-limit-on-windows), but I didn't get that working with Xilinx SDK. Mentioned Make.exe is too old.

The link command is 9891 bytes long :
~~~
    arm-none-eabi-gcc -mcpu=cortex-a9...
~~~

Thanks, Hein

abhidixi11:
Hello Hein,
Thanks for reaching out, apologies for delay in responding. 
I'm trying to understand if there is any difference between your setup and how build on our internal nightly test server works. I don't have Xilinx setup with me, so I'm attaching a log file for the recent nightly build, I'll try to get more detail logs on Monday if you are still stuck by then.
I'm assuming you are following getting started guide:
https://docs.aws.amazon.com/freertos/latest/userguide/getting_started_xilinx.html
Here is build log:
[xlinxLogs.txt](https://github.com/aws/amazon-freertos/files/3906779/xlinxLogs.txt)

htibosch:
Hello, and thanks for the links.
Now I followed the "getting started" guidelines, but this build also ends in a command that is 11,708 bytes long, too long for Windows' cmd.exe.
In a meanwhile I'm installing XSDK 2019.1 on Ubuntu, so I can continue my work.
I have a pending PR for the NetworkInterface for Xilinx, and want to test it in the demo and test applications.
But still I'd like to get it built on Windows as well.

abhidixi11:
Thanks for update, I'll check  tomorrow if anyone here has seen character limit issue with XSDK. Meanwhile, I'm not sure if you have seen this thread from Xilinx forum, which talks about the same issue, may be you can create static libraries of .o files as mentioned there. 
[...]

PR #1628, opened 2019-12-15
FreeRTOS+TCP : Revision of Xilinx Network Interface

htibosch:
Description
-----------
It is a long time that the network interface of Xilinx had an update in AWS/FreeRTOS repo.
All changes have been tested for a long time.

Here a summary:

Source file `x_emacpsif_physpeed.c` :

- Stopped using `xil_printf()`

- Solved some `#include` problems

- Make PHY detection better:
  - Start from address 0, not from 32
  - More informative output: "XEmacPs detect_phy: C9150010 at address 0"

- Replaced `sleep( 1 )` with the FreeRTOS API `vTaskDelay( 2 )`.

Source file `x_emacpsif_hw.c` :

- Solved some `#include` problems

- Some cleaning ( use headers in stead of local declarations of functions and variables )

Source file `x_emacpsif_dma.c` :

- Solved some `#include` problems

- Stopped using the macro's `ipTOTAL_ETHERNET_FRAME_SIZE` and `RX_BUFFER_ALIGNMENT`

Several **bugs are solved** by reading back registers in several places:

- In `emacps_send_handler()` : read-back the `XEMACPS_TXSR` registers in ISR context after clearing it
- In `emacps_send_message()` : read-back the `XEMACPS_NWCTRL` register after setting it
- In `emacps_recv_handler()` : read-back the `XEMACPS_RXSR` register in ISR context after clearing it
- In `emacps_check_rx()` : read-back the `address` field of a DMA descriptor to make sure it is set.

- The user is now free to use `ipconfigUSE_LINKED_RX_MESSAGES` or not. The option may increase performance because incoming network packets are linked together in a list and send as 1 queue message to the IP-task.

- Cleaned up some code:

~~~c
-static NetworkBufferDescriptor_t *ethMsg = NULL;
-static NetworkBufferDescriptor_t *ethLast = NULL;
~~~

These variables are now declared the stack.

- The ICMP-checksum offloading for outgoing packets didn't work. `xNetworkInterfaceOutput()` will now correct this.

- Added the function `prvMonitorResources()` that helps users monitor the essential resources while developing. The code is only included if FreeRTOS+TCP logging is used ( `ipconfigHAS_PRINTF` ).

- The PHY status: when a packet come in while the PHY was low, it is assume that it is high. However, the variable should be updated:

~~~c
    /* Indicate that the Link Status is high, so that
    xNetworkInterfaceOutput() can send packets. */
    ulPHYLinkStatus |= niBMSR_LINK_STATUS;
~~~

NB. I had problems linking the project within Eclipse on Windows 10. When calling the linker, the command line got larger than 8 KB ( 10 or 11 KB ), which was a problem for Windows.
I found a solution and described it [here](https://github.com/aws/amazon-freertos/issues/1555).

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
@@ -38,6 +38,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"
 #include "FreeRTOS_IP_Private.h"
+#include "FreeRTOS_ARP.h"
 #include "NetworkBufferManagement.h"
 #include "NetworkInterface.h"
 
@@ -50,9 +51,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 /* Provided memory configured as uncached. */
 #include "uncached_memory.h"
 
-#ifndef	BMSR_LINK_STATUS
-	#define BMSR_LINK_STATUS            0x0004UL
+#ifndef niEMAC_HANDLER_TASK_PRIORITY
+	/* Define the priority of the task prvEMACHandlerTask(). */
+	#define niEMAC_HANDLER_TASK_PRIORITY	configMAX_PRIORITIES - 1
 #endif
+
+#define niBMSR_LINK_STATUS         0x0004uL
 
 #ifndef	PHY_LS_HIGH_CHECK_TIME_MS
 	/* Check if the LinkSStatus in the PHY is still high after 15 seconds of not
@@ -83,6 +87,13 @@ FreeRTOSConfig.h as configMINIMAL_STACK_SIZE is a user definable constant. */
 	#define configEMAC_TASK_STACK_SIZE ( 2 * configMINIMAL_STACK_SIZE )
 #endif
 
+#if( ipconfigZERO_COPY_RX_DRIVER == 0 || ipconfigZERO_COPY_TX_DRIVER == 0 )
+	#error Please define both 'ipconfigZERO_COPY_RX_DRIVER' and 'ipconfigZERO_COPY_TX_DRIVER' as 1
+#endif
+
+#if( ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM == 0 || ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM == 0 )
+	#warning Please define both 'ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM' and 'ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM' as 1
+#endif
 /*-----------------------------------------------------------*/
 
 /*
@@ -95,6 +106,10 @@ static BaseType_t prvGMACWaitLS( TickType_t xMaxTime );
  * A deferred interrupt handler for all MAC/DMA interrupt sources.
  */
 static void prvEMACHandlerTask( void *pvParameters );
+
+#if ( ipconfigHAS_PRINTF != 0 )
+    static void prvMonitorResources( void );
+#endif
 
 /*-----------------------------------------------------------*/
 
@@ -188,7 +203,7 @@ const TickType_t xWaitLinkDelay = pdMS_TO_TICKS( 7000UL ), xWaitRelinkDelay = pd
 		possible priority to ensure the interrupt handler can return directly
 		to it.  The task's handle is stored in xEMACTaskHandle so interrupts can
 		notify the task when there is something to process. */
-		xTaskCreate( prvEMACHandlerTask, "EMAC", configEMAC_TASK_STACK_SIZE, NULL, configMAX_PRIORITIES - 1, &xEMACTaskHandle );
+		xTaskCreate( prvEMACHandlerTask, "EMAC", configEMAC_TASK_STACK_SIZE, NULL, niEMAC_HANDLER_TASK_PRIORITY, &xEMACTaskHandle );
 	}
 	else
 	{
@@ -206,7 +221,27 @@ const TickType_t xWaitLinkDelay = pdMS_TO_TICKS( 7000UL ), xWaitRelinkDelay = pd
 
 BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxBuffer, BaseType_t bReleaseAfterSend )
 {
-	if( ( ulPHYLinkStatus & BMSR_LINK_STATUS ) != 0 )
+	#if( ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM != 0 )
+	{
+	ProtocolPacket_t *pxPacket;
+
+		/* If the peripheral must calculate the checksum, it wants
+		the protocol checksum to have a value of zero. */
+		pxPacket = ( ProtocolPacket_t * ) ( pxBuffer->pucEthernetBuffer );
+		if( pxPacket->xICMPPacket.xIPHeader.ucProtocol == ipPROTOCOL_ICMP )
+		{
+		IPHeader_t *pxIPHeader = &( pxPacket->xUDPPacket.xIPHeader );
+
+			pxPacket->xICMPPacket.xICMPHeader.usChecksum = ( uint16_t )0u;
+			pxIPHeader->usHeaderChecksum = 0u;
+			pxIPHeader->usHeaderChecksum = usGenerateChecksum( 0UL, ( uint8_t * ) &( pxIPHeader->ucVersionHeaderLength ), ipSIZE_OF_IPv4_HEADER );
+			pxIPHeader->usHeaderChecksum = ~FreeRTOS_htons( pxIPHeader->usHeaderChecksum );
+
+			usGenerateProtocolChecksum( (uint8_t*)&( pxPacket->xUDPPacket ), pxBuffer->xDataLength, pdTRUE );
+		}
+	}
+	#endif /* ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM */
+	if( ( ulPHYLinkStatus & niBMSR_LINK_STATUS ) != 0 )
 	{
 		iptraceNETWORK_INTERFACE_TRANSMIT();
 		emacps_send_message( &xEMACpsif, pxBuffer, bReleaseAfterSend );
@@ -249,7 +284,7 @@ BaseType_t xReturn;
 		}
 		ulPHYLinkStatus = ulReadMDIO( PHY_REG_01_BMSR );
 
-		if( ( ulPHYLinkStatus & BMSR_LINK_STATUS ) != 0 )
+		if( ( ulPHYLinkStatus & niBMSR_LINK_STATUS ) != 0 )
 		{
 			xReturn = pdTRUE;
 			break;
@@ -281,7 +316,7 @@ BaseType_t xGetPhyLinkStatus( void )
 {
 BaseType_t xReturn;
 
-	if( ( ulPHYLinkStatus & BMSR_LINK_STATUS ) == 0 )
+	if( ( ulPHYLinkStatus & niBMSR_LINK_STATUS ) == 0 )
 	{
 		xReturn = pdFALSE;
 	}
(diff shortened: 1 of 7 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
```diff
@@ -206,7 +221,27 @@ const TickType_t xWaitLinkDelay = pdMS_TO_TICKS( 7000UL ), xWaitRelinkDelay = pd
...
+		/* If the peripheral must calculate the checksum, it wants
+		the protocol checksum to have a value of zero. */
+		pxPacket = ( ProtocolPacket_t * ) ( pxBuffer->pucEthernetBuffer );
+		if( pxPacket->xICMPPacket.xIPHeader.ucProtocol == ipPROTOCOL_ICMP )
+		{
+		IPHeader_t *pxIPHeader = &( pxPacket->xUDPPacket.xIPHeader );
+
+			pxPacket->xICMPPacket.xICMPHeader.usChecksum = ( uint16_t )0u;
+			pxIPHeader->usHeaderChecksum = 0u;
+			pxIPHeader->usHeaderChecksum = usGenerateChecksum( 0UL, ( uint8_t * ) &( pxIPHeader->ucVersionHeaderLength ), ipSIZE_OF_IPv4_HEADER );
```
dcgaws: The purpose of this block of code is that the #define above indicates that the driver itself should calculate the checksum, correct?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
```diff
@@ -56,7 +56,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #endif
 #define TX_OFFSET				ipconfigPACKET_FILLER_SIZE
 
-#define RX_BUFFER_ALIGNMENT	14
+#define dmaRX_TX_BUFFER_SIZE			1536
```
dcgaws: Should this instead be a common #define for Ethernet?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
```diff
@@ -170,6 +167,11 @@ BaseType_t xHigherPriorityTaskWoken = pdFALSE;
 
 	xemacpsif = (xemacpsif_s *)(arg);
 
+	/* This function is called from an ISR. The driver has already cleared the
```
dcgaws: What driver? I thought this is the driver.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
```diff
@@ -422,8 +433,12 @@ int head = xemacpsif->rxHead;
 					addr |= XEMACPS_RXBUF_WRAP_MASK;
 				}
 				/* Clearing 'XEMACPS_RXBUF_NEW_MASK'       0x00000001 *< Used bit.. */
-				xemacpsif->rxSegments[ head ].address = addr;
 				xemacpsif->rxSegments[ head ].flags = 0;
+				xemacpsif->rxSegments[ head ].address = addr;
+				if (xemacpsif->rxSegments[ head ].address)
+				{
+					// Just to read it
```
dcgaws: What is the meaning of this comment?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
```diff
@@ -480,7 +495,7 @@ XStatus init_dma(xemacpsif_s *xemacpsif)
 	xTxSize = ipconfigNIC_N_TX_DESC * sizeof( xemacpsif->txSegments[ 0 ] );
 
 	/* Also round-up to 4KB */
-	xemacpsif->uTxUnitSize = ( ipTOTAL_ETHERNET_FRAME_SIZE + 0x1000ul ) & ~0xffful;
+	xemacpsif->uTxUnitSize = ( dmaRX_TX_BUFFER_SIZE + 0x1000ul ) & ~0xffful;
```
dcgaws: Why round up to 4k?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_physpeed.c
```diff
@@ -163,19 +161,17 @@ int phy_detected = 0;
...
-		if ((phy_reg != 0xFFFF) &&
-			((phy_reg & PHY_DETECT_MASK) == PHY_DETECT_MASK)) {
-			/* Found a valid PHY address */
-			FreeRTOS_printf( ("XEmacPs detect_phy: PHY detected at address %d.\r\n",
-																	phy_addr));
-			FreeRTOS_printf( ("XEmacPs detect_phy: PHY detected.\n" ) );
+	u16 id_lower, id_upper;
+	u32 phy_addr, id;
+
+	for (phy_addr = 0; phy_addr < 32; phy_addr++) {
```
dcgaws: What's the significance of 32?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_physpeed.c
```diff
@@ -338,32 +334,32 @@ unsigned get_IEEE_phy_speed(XEmacPs *xemacpsp)
 			break;
 	}
 #endif
-	xil_printf("Waiting for PHY to complete autonegotiation.\r\n");
+	FreeRTOS_printf( ( "Waiting for PHY to complete autonegotiation.\n" ) );
 
 	XEmacPs_PhyRead(xemacpsp, phy_addr, IEEE_STATUS_REG_OFFSET, &status);
 	while ( !(status & IEEE_STAT_AUTONEGOTIATE_COMPLETE) ) {
-		sleep(1);
+		vTaskDelay( 2 );
```
dcgaws: Should this delay value be a configurable #define?

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
htibosch: I changed it to: 

~~~c
if (xemacpsif->rxSegments[ head ].address)
{
    /* Make sure that the value has reached the peripheral by reading it back. */
}

It works like a memory barrier.
~~~

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
```diff
@@ -50,10 +51,13 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
 #include "uncached_memory.h"
 
-#ifndef	BMSR_LINK_STATUS
-	#define BMSR_LINK_STATUS            0x0004UL
+#ifndef niEMAC_HANDLER_TASK_PRIORITY
+	/* Define the priority of the task prvEMACHandlerTask(). */
+	#define niEMAC_HANDLER_TASK_PRIORITY	configMAX_PRIORITIES - 1
 #endif
 
+#define niBMSR_LINK_STATUS         0x0004uL
```
gkwicker: Since [MII is a standard](https://en.wikipedia.org/wiki/Media-independent_interface), we should collect all of the corresponding defines in an mii.h include file and refer to that.  It looks like MII definitions are being defined in every driver which uses them.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
gkwicker: The comments on lines 228/229 make it sound like the driver is using Zynq checksum offloading (CSO).  The code here is using 0 for the ICMP checksum at line 235 but generating an IP checksum at line 237 (after explicitly clearing it on line 236).  Wouldn't the CSO calculate the IP checksum?  I'm not that familiar with this feature, but [this Xilinx post](https://forums.xilinx.com/t5/Ethernet/How-to-enable-AXI-Ethernet-LWIP-Full-checksum-offload/m-p/895249) seems to indicate that you wouldn't need line 237 if CSO was enabled.  Please clarify the comments and adjust the code if needed.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c
gkwicker: +1

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_physpeed.c
gkwicker: There are [up to 32 phy slaves on the MDIO bus](https://en.wikipedia.org/wiki/Management_Data_Input/Output). To Dan's point, some links/pointers to applicable standards documents are helpful in code like this.  Also, if this is standard MDIO code we could probably make it a common module at some point, but that's beyond the scope of this PR.

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_physpeed.c
gkwicker: +1.  Also, how does this code handle a plug/unplug during operation?

htibosch: I added a define `MINIMUM_SLEEP_TIME` with a value of 2.
As you know, a `vTaskDelay(1)` could return witin a few µS when called just before a clock-tick.
`

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c
htibosch: As stated here below, Xilinx is standing alone because it doesn't use the module `phyHandling.c` ( because it has a 1 Gbps PHY ).

yuhui-zheng:
I ran into this PR while browsing around. My comment below is not related to this PR, but rather the port.

@htibosch While we were doing the recent FreeRTOS release, we found two driver files for Zynq seemed to be missing -- ```uncached_memory.{c, h}``` (see https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/Zynq). ```uncached_memory.h``` is included in [```x_emacpsif_dma.c```](https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/x_emacpsif_dma.c#L46), and in the same file we are calling functions prototyped in ```uncached_memory.h```. So it seems I need to merge back those two missing files. Please comment back if you have concern or further info.

htibosch:
Hello yuhui-zheng,

There are two copies of each file you mention, and they are part of the Xilinx application code:

vendors/xilinx/boards/microzed/aws_**demos**/application_code/xilinx_code/uncached_memory.c
vendors/xilinx/boards/microzed/aws_**demos**/application_code/xilinx_code/uncached_memory.h
vendors/xilinx/boards/microzed/aws_**tests**/application_code/xilinx_code/uncached_memory.c
vendors/xilinx/boards/microzed/aws_**tests**/application_code/xilinx_code/uncached_memory.h

I don't see any good reason why equal copies were placed into two directories.

Originally, the location of the files was in the network porting directory:

libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq

We could move the files back there

yuhui-zheng:
@htibosch — Thank you for the information. I'm voting for putting ```uncached_memory.{c, h}``` files back under ```portable/NetworkInterface/Zynq```. Since I felt strongly about it, I guess I'll send a PR. 

----------------


The rationale:
I've looked at the dependency. DMA appears to be essential to [the network driver on this port](https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq/NetworkInterface.c#L182) (I'm skipping examples where EMAC Tx uses DMA), DMA itself depends on uncached memory implementation. So, I think it's better to group them together.

It does seem quite arbitrary about where to cut the line, since uncached memory depends on other Xilinx components e.g. MMU etc. and software library implementation e.g. memcpy.c in the same directory. If the question may be about why not putting all dependencies in the same place...  I think hardware implementation related is nice to be out of driver directory (e.g. MMU implementation may be different in each version of the BSP, though interface would be the same), and common library is nice to be excluded as well (e.g. memcpy/memset).

htibosch:
@yuhui-zheng moved uncached_memory.{c, h} to the network port directory, which is very good.
I will now close this PR and come up with a new commit that is rebased.

PR #1639, opened 2019-12-17
Added macro to disable common folder hooks

lundinc2:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/demo_runner/iot_demo_freertos.c
+++ b/demos/demo_runner/iot_demo_freertos.c
@@ -404,10 +404,11 @@ void runDemoTask( void * pArgument )
 
 /*-----------------------------------------------------------*/
 
-/* It is recommended to implement hooks that use platform specific APIs. This allows 
+/* It is recommended to implement hooks that use platform specific APIs. This allows
  * for better error messages and recovery. Should platform specific hooks be implemented,
  * add this macro to iot_config.h to avoid compiling these symbols.*/
 #ifndef iotconfigUSE_PORT_SPECIFIC_HOOKS
+
 /**
  * @brief Warn user if pvPortMalloc fails.
  *
@@ -418,16 +419,16 @@ void runDemoTask( void * pArgument )
  * configTOTAL_HEAP_SIZE configuration constant in FreeRTOSConfig.h.
  *
  */
-void vApplicationMallocFailedHook()
-{
-    configPRINT_STRING( ( "ERROR: Malloc failed to allocate memory\r\n" ) );
-    taskDISABLE_INTERRUPTS();
-
-    /* Loop forever */
-    for( ; ; )
+    void vApplicationMallocFailedHook()
     {
+        configPRINT_STRING( ( "ERROR: Malloc failed to allocate memory\r\n" ) );
+        taskDISABLE_INTERRUPTS();
+
+        /* Loop forever */
+        for( ; ; )
+        {
+        }
     }
-}
 
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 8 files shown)


yuhui-zheng:
Thank you for this PR -- this cleans up a lot of tech debt and clarifies what needs to be port specific. For a broader audience:

https://www.freertos.org/a00016.html 
- Since each port can use whatever heap implementation vendor prefers, ```vApplicationMallocFailedHook``` is better port specific. 
- For ```vApplicationStackOverflowHook```, https://www.freertos.org/Stacks-and-stack-overflow-checking.html this clarifies why it needs to be port specific.

lundinc2:
@dcgaws

> The net result of this change is that we're making a bunch of exact copies of two functions. Why is that better?

Since these functions are defined in tests/common/iot_test_freertos.c and demos/common/iot_demo_freertos.c they override platform specific symbols. Ideally these hooks would be implemented properly for every platform but this copy and paste is a step in the right direction.

As an example, for our Microchip port these symbols are defined three times: https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/microchip/boards/curiosity_pic32mzef/aws_tests/application_code/microchip_code/rtos_hooks.c#L55 and here https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/microchip/boards/curiosity_pic32mzef/aws_tests/application_code/main.c#L346
 
When really they should only be defined once, in the port layer of our co debase.

Here is another example using the Espressif port.

https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/espressif/esp-idf/components/esp32/panic.c#L117

This hook is a lot more informative and uses a platform specific API, but it gets overridden by hooks defined in the common code section.

lundinc2:
@

> The net result of this change is that we're making a bunch of exact copies of two functions. Why is that better?

Since these functions are defined in tests/common/iot_test_freertos.c and demos/common/iot_demo_freertos.c they override platform specific symbols. Ideally these hooks would be implemented properly for every platform but this copy and paste is a step in the right direction.

As an example, for our Microchip port these symbols are defined three times: https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/microchip/boards/curiosity_pic32mzef/aws_tests/application_code/microchip_code/rtos_hooks.c#L55 and here https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/microchip/boards/curiosity_pic32mzef/aws_tests/application_code/main.c#L346
 
When really they should only be defined once, in the port layer of our co debase.

Here is another example using the Espressif port.

https://github.com/aws/amazon-freertos/blob/fb6bf8e1980d0156c0e072ad5f284ebc997ea48a/vendors/espressif/esp-idf/components/esp32/panic.c#L117

This hook is a lot more informative and uses a platform specific API, but it gets overridden by hooks defined in the common code section.

dcgaws:
The ESP32 stack overflow hook is equivalent to ours: it writes the error text and task name to the console and then blocks forever. Ironically, the Microchip implementation is less functional: it blocks forever, but doesn't write anything to the console. 

Regarding the reference to https://www.freertos.org/Stacks-and-stack-overflow-checking.html, I read that as asserting that stack overflow detection is port specific. But, as our implementation seems to be demonstrating, I don't see an indication that the hook is necessarily port specific.

As an alternative, could we try an implementation where we provide a single version of the hooks with Amazon FreeRTOS? If the target silicon framework has one, use it instead. Otherwise, use ours.

lundinc2:
The problem of putting it in a common place to be overridden, is in the linking stage the correct symbol still might not be selected. Here is some info from ARM on how symbols are resolve when linking libraries. http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka15833.html. I believe the best solution is to put the hooks all in port specific section so we can guarantee the correct hook was compiled in.

I will note that the ESP32 hook is better, because it does not allocate memory, as internally configPRINTF calls malloc. This causes a hard fault when malloc failed hook is called.

mahavirj:
> The ESP32 stack overflow hook is equivalent to ours: it writes the error text and task name to the console and then blocks forever.

Its quite different, it does not use full-fledged `printf` (since context for stack overflow hook might be ISR so available stack is limited) but low level UART APIs for debug prints and then followed by call to `abort` to generate dummy exception which can help in getting entire backtrace information under `idf_monitor` utility.

> As an alternative, could we try an implementation where we provide a single version of the hooks with Amazon FreeRTOS? If the target silicon framework has one, use it instead. Otherwise, use ours.

This sounds good, as long as platform has ability for selection (through say platform specific `config_files`), it should achieve the purpose.

PR #1663, opened 2020-01-03
Fix BLE enable disable sequence

### LINKED ISSUE #1576: xBLEStackTeardown in main.c
Can we fix `xBLEStackTeardown` when NimBLE is used, because in NimBLE we dont have `esp_bt_controller_mem_release`:
https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/aws_demos/application_code/main.c#L195

with something like that:
```
        esp_err_t xBLEStackTeardown( void )
        {
            BTStatus_t xRet = IotBle_Off(); // 
            const BTInterface_t* xBTinter = BTGetBluetoothInterface(); 
            xBTinter->pxDisable();
            xBTinter->pxBtManagerCleanup();
            return ESP_OK;
        }
```
esp32 master commit

ravibhagavandas:
Fix enable/disable sequence in BLE to support disabling and re-enabling stack at runtime.

Description
-----------
* Refactor IotBle_On() and IotBle_Off() functions to be symmetrical - enable/teardown GATT services and BLE stack.
* Example usage in demo runner to disable non-used networks for demo . (Disable BLE stack to save power and reclaim RAM after WiFi provisioning when WiFi is connected).
* Tested on ESP32 and Nordic.

PR also fixes #1576

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -47,6 +47,7 @@ static struct ble_gatt_svc_def espServices[ MAX_SERVICES + 1 ];
 static BTService_t * afrServices[ MAX_SERVICES ];
 static uint16_t serviceCnt = 0;
 static SemaphoreHandle_t xSem;
+static bool semInited;
 bool xSemLock = 0;
 uint16_t gattOffset = 0;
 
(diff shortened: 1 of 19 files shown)


vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -727,69 +766,41 @@ static uint16_t prvCountDescriptor( BTService_t * pxService,
...
-                break;
-
-            default:
-                break;
+                vPortFree( ( void * ) pCharacteristic->descriptors );
+            }
         }
-    }
 
-    if( pSvc->characteristics != NULL )
```
cobusve: Do we really not need to check this any more? What if the malloc failed halfway though the creation, do we leak in that case?

ravibhagavandas: The check is now at line 780 before iterating through the list of characteristics and descriptors.

PR #1667, opened 2020-01-06
declare variables outside "for loop" in spi tests

qiutongs:
Description
-----------

Declaring variables in the "for loop" is not allowed by TI-gcc compiler.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_spi.c
+++ b/libraries/abstractions/common_io/test/test_iot_spi.c
@@ -268,6 +268,7 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncAssisted )
     uint8_t ucRxBuf[ 16 ] = { 0 };
     char cMsg[ 50 ] = { 0 };
     size_t xBytesRx;
+    int i = 0, j = 0, len = 0;
 
     /* If unit test and assisted test have different spi slave, make sure select slave function in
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -268,6 +268,7 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncAssisted )
     uint8_t ucRxBuf[ 16 ] = { 0 };
     char cMsg[ 50 ] = { 0 };
     size_t xBytesRx;
+    int i = 0, j = 0, len = 0;
```
lundinc2: Prefer size specific integers, either 32, 8, or size_t

haipengd: Sorry late comment. I see similar code for other test, e.g. test_iot_flash.c, do we also need to correct all the test files?

lundinc2: It's a nit, so i'm not sure. I think to keep with  the code style then yes we would

qiutongs: I think doing the way Carl mentioned is good only because of consistent coding style, in this case. So let's not worry it now.

But please fix files if you see variables are declared inside "for loop"

PR #1669, opened 2020-01-07
Fix several issues in I2C tests after testing on TI board

qiutongs:
Description
-----------

- remove unnecessary NO_STOP in test AFQP_IotI2CReadAsyncFailReadTwice
- move iot_i2c_set_callback after iot_i2c_write_sync in test AFQP_IotI2CCancelReadSuccess
- add configurable IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH to limit the length of write buffer; this causes fault on TI
    - provide a default value for IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH 
    - vendor can override it in "test_iot_config"

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_i2c.c
+++ b/libraries/abstractions/common_io/test/test_iot_i2c.c
@@ -619,10 +619,6 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CReadAsyncFailReadTwice )
         lRetVal = iot_i2c_ioctl( xI2CHandle, eI2CSetSlaveAddr, &uctestIotI2CSlaveAddr );
         TEST_ASSERT_EQUAL( IOT_I2C_SUCCESS, lRetVal );
 
-        /* Set i2c configuration */
-        lRetVal = iot_i2c_ioctl( xI2CHandle, eI2CSendNoStopFlag, NULL );
-        TEST_ASSERT_EQUAL( IOT_I2C_SUCCESS, lRetVal );
-
         /* read from i2c device */
         lRetVal = iot_i2c_read_async( xI2CHandle, ucBuffer, sizeof( ucBuffer ) );
         TEST_ASSERT_EQUAL( IOT_I2C_SUCCESS, lRetVal );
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -921,7 +917,7 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CWriteAsyncFailWriteTwice )
 {
     IotI2CHandle_t xI2CHandle;
     int32_t lRetVal;
-    uint8_t writeVal[] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF };
+    uint8_t writeVal[ IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH ] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal };
```
cmcgcc: Same here.

libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -871,7 +867,7 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CWriteAsyncFailIoctl )
...
```
cmcgcc: Is one byte too few here? Write may finish before the next api call especially in the multi-task env.

qiutongs: IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH is configurable. So the writeVal array is actually initialize to the size of what you want. The first two bytes are xtestIotI2CDeviceRegister and uctestIotI2CWriteVal, and the rest of writeVal will be filled 0. I think 0 is fine because the value doesn't matter

cmcgcc: The problem is user may neglect to set more bytes and get wrong result if it is not enforced here.

qiutongs: I see your concern. I plan to give a default value to IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH and vendor can override.

cmcgcc: Is it possible to give something more than 2? Will it affect TI?

qiutongs: See the latest code, I set to 10.

PR #1673, opened 2020-01-07
test: Expand on how we enable/disable peripheral tests

hesham-mahmoud:
This change provide:
1- Using macros to enable/diable running tests for peripherals
2- Default the macros to "enable", to avoid forgetting about setting it
for a test
3- Unify the macro used to enable the test or not with the old macro
that would set the number of tests to run.
3- Collect all variables that can be changed by the peripheral
configuration function into one header.

Why:
- Simplifying the header exposed to the vendors
- Reduce macros needed for each peripheral

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/iot_test_common_io.c
+++ b/libraries/abstractions/common_io/test/iot_test_common_io.c
@@ -57,7 +57,7 @@ TEST_GROUP_RUNNER( Common_IO )
     size_t i = 0;
 
     #if ( IOT_TEST_COMMON_IO_UART_SUPPORTED == 1 )
-        for( i = 0; i < UART_TEST_SET; i++ )
+        for( i = 0; i < IOT_TEST_COMMON_IO_UART_SUPPORTED; i++ )
         {
             SET_TEST_IOT_UART_CONFIG( i );
             RUN_TEST_GROUP( TEST_IOT_UART );
@@ -65,7 +65,7 @@ TEST_GROUP_RUNNER( Common_IO )
     #endif
 
     #if ( IOT_TEST_COMMON_IO_I2C_SUPPORTED == 1 )
-        for( i = 0; i < I2C_TEST_SET; i++ )
+        for( i = 0; i < IOT_TEST_COMMON_IO_I2C_SUPPORTED; i++ )
         {
             SET_TEST_IOT_I2C_CONFIG( i );
             RUN_TEST_GROUP( TEST_IOT_I2C );
@@ -73,7 +73,7 @@ TEST_GROUP_RUNNER( Common_IO )
     #endif
 
     #if ( IOT_TEST_COMMON_IO_SPI_SUPPORTED == 1 )
-        for( i = 0; i < SPI_TEST_SET; i++ )
+        for( i = 0; i < IOT_TEST_COMMON_IO_SPI_SUPPORTED; i++ )
         {
             SET_TEST_IOT_SPI_CONFIG( i );
             RUN_TEST_GROUP( TEST_IOT_SPI );
--- a/libraries/abstractions/common_io/test/iot_test_common_io.c
+++ b/libraries/abstractions/common_io/test/iot_test_common_io.c
@@ -56,7 +56,7 @@ TEST_GROUP_RUNNER( Common_IO )
 {
     size_t i = 0;
 
-    #if ( IOT_TEST_COMMON_IO_UART_SUPPORTED == 1 )
+    #ifdef IOT_TEST_COMMON_IO_UART_SUPPORTED
         for( i = 0; i < IOT_TEST_COMMON_IO_UART_SUPPORTED; i++ )
         {
             SET_TEST_IOT_UART_CONFIG( i );
@@ -64,7 +64,7 @@ TEST_GROUP_RUNNER( Common_IO )
         }
     #endif
 
-    #if ( IOT_TEST_COMMON_IO_I2C_SUPPORTED == 1 )
+    #ifdef IOT_TEST_COMMON_IO_I2C_SUPPORTED
         for( i = 0; i < IOT_TEST_COMMON_IO_I2C_SUPPORTED; i++ )
         {
             SET_TEST_IOT_I2C_CONFIG( i );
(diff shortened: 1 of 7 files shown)


libraries/abstractions/common_io/test/test_iot_internal.h
```diff
@@ -33,6 +33,190 @@
...
+
+/*
+ * The variables below are used in the tests. Generally. they should be
+ * set in SET_TEST_IOT_RESET_CONFIG() defined in the respective platform
+ * directory. If not set, the default value set at the variable
+ * definition is used.
+ */
+
+/* Uart */
+extern uint8_t ustestIotUartPort;                       /* the index of the UART that will be tested */
```
gkwicker: Since this is a `uint8_t`, it should be named `uctestIotUartPort` according to the [FreeRTOS Coding Style Guide.](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html)

libraries/abstractions/common_io/test/test_iot_internal.h
```diff
@@ -33,6 +33,190 @@
...
+
+/* GPIO */
+extern int32_t ltestIotGpioPortA;                       /* The 1st GPIO port used in the loop back test */
+extern int32_t ltestIotGpioPortB;                       /* The 2nd GPIO port used in the loop back test */
+extern uint16_t ustestIotGpioConfig;                    /* ustestIotGpioConfig is used for configuring GPIO test
+                                                         * user must set this variable for testing different GPIOs
+                                                         *
+                                                         * default ustestIotGpioConfig = 0 means reading logic GPIO# 0
+                                                         *
+                                                         * bit[7:0] for port
```
gkwicker: Are these bit definitions defined anywhere?  If so, just refer to the symbols here rather than defining them in comments.

libraries/abstractions/common_io/test/iot_test_common_io.c
```diff
@@ -57,23 +57,23 @@ TEST_GROUP_RUNNER( Common_IO )
     size_t i = 0;
 
     #if ( IOT_TEST_COMMON_IO_UART_SUPPORTED == 1 )
-        for( i = 0; i < UART_TEST_SET; i++ )
+        for( i = 0; i < IOT_TEST_COMMON_IO_UART_SUPPORTED; i++ )
```
qiutongs: Thanks for doing this.

libraries/abstractions/common_io/test/test_iot_internal.h
```diff
@@ -33,6 +33,190 @@
 
 #pragma once
 
+
+/*
+ * Test iteration count macros that platforms should set
+ * in their repective "test_iot_config.h" file. If not set
+ * there, the default value of 1 is set here.
+ */
```
qiutongs: I am OK with this. @cobusve  What do you think?

libraries/abstractions/common_io/test/iot_test_common_io.c
haipengd: With such change: for( i = 0; i < IOT_TEST_COMMON_IO_UART_SUPPORTED; i++ )
the loop will be actually always 1, so do we still need foor loop here.

libraries/abstractions/common_io/test/test_iot_internal.h
hesham-mahmoud: They are defined as macros in test_iot_gpio.c
Here is how they look in that file, for your reference:

```
#define testIotGPIO_TEST_PORT_MASK  ( 0x00FF )
#define testIotGPIO_TEST_DIR_BIT    ( 8 )
#define testIotGPIO_TEST_DIR_MASK   ( 1 << testIotGPIO_TEST_DIR_BIT )
#define testIotGPIO_TEST_IRQ_BIT    ( 9 )
#define testIotGPIO_TEST_IRQ_MASK   ( 1 << testIotGPIO_TEST_IRQ_BIT )
#define testIotGPIO_TEST_VALUE_BIT  ( 10 )
#define testIotGPIO_TEST_VALUE_MASK ( 1 << testIotGPIO_TEST_VALUE_BIT )
#define testIotGPIO_TEST_READ_DELAY ( 50 )
```
All the macros above except the last one are mapped to this bits in the comment under review.

libraries/abstractions/common_io/test/iot_test_common_io.c
hesham-mahmoud: No, this same macro will mean how many tests you want to run, the same meaning of the current `UART_TEST_SET`

libraries/abstractions/common_io/test/test_iot_internal.h
hesham-mahmoud: OK, will update it.

libraries/abstractions/common_io/test/iot_test_common_io.c
haipengd: But we have #if ( IOT_TEST_COMMON_IO_UART_SUPPORTED == 1 ), right?
We only check if equals 1, if not 1, it won't run at all.

hesham-mahmoud: Good catch, I'll do the needed change.

libraries/abstractions/common_io/test/test_iot_internal.h
gkwicker: Quick reminder about this

libraries/abstractions/common_io/test/test_iot_internal.h
gkwicker: Can you add a reference to this file and/or these macros in the comments here?  The point is to make it so that someone reading this code understands where these definitions actually exist; this will help them from getting out of sync.

libraries/abstractions/common_io/test/test_iot_internal.h
hesham-mahmoud: Working on it. Thanks for your reminder.

hesham-mahmoud: Done.

libraries/abstractions/common_io/test/test_iot_internal.h
hesham-mahmoud: Done.

libraries/abstractions/common_io/test/iot_test_common_io.c
hesham-mahmoud: Fixed, please check.

PR #1674, opened 2020-01-07
Fix potential segfault (dereferencing a pointer before checked for NULL)

alfred2g:
Committer: Alfred Gedeon <alfred2g@hotmail.com>

Description
-----------
Fix a potential segfault, by trying to dereference a pointer before checking if it is NULL

Checklist:
----------

- [ ] My code is Linted.


libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -412,11 +412,6 @@ int32_t SOCKETS_Recv( Socket_t xSocket,
 {
     ss_ctx_t * ctx = ( ss_ctx_t * ) xSocket;
 
-    if( ( ctx->status & SS_STATUS_CONNECTED ) != SS_STATUS_CONNECTED )
-    {
-        return SOCKETS_ENOTCONN;
-    }
-
     if( SOCKETS_INVALID_SOCKET == xSocket )
```
gkwicker: Note that `SOCKETS_INVALID_SOCKET != NULL`, (definition [here](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/secure_sockets/include/iot_secure_sockets.h#L110)) so the commit message is incorrect.  Let's change the commit message to "Validate socket handle before dereferencing."

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -464,6 +464,11 @@ int32_t SOCKETS_Send( Socket_t xSocket,
         return SOCKETS_EINVAL;
     }
 
+    if( ( ctx->status & SS_STATUS_CONNECTED ) != SS_STATUS_CONNECTED )
```
qiutongs: It is a bug: ctx is uninitialized in line 455.

qiutongs: You want to do "ctx = ( ss_ctx_t * ) xSocket" first.

gkwicker: Two issues:

1. `ctx` has not been assigned yet, that doesn't happen until line 472.
1. The commit message does not describe this change; `xSocket != SOCKETS_INVALID_SOCKET` was guaranteed by line 457.  If this is a change we want to make, it should be in a separate commit/PR with a message like "Check for disconnected socket in SOCKETS_Send."

PR #1682, opened 2020-01-09
Copy latest common io changes to master

qiutongs:
Description
-----------

Copy latest common io changes to master:
- no other changes except formatting iot_uart.h

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/include/iot_uart.h
+++ b/libraries/abstractions/common_io/include/iot_uart.h
@@ -60,9 +60,10 @@
  */
 typedef enum
 {
-    eUartCompleted = IOT_UART_SUCCESS,            /*!< UART operation completed successfully. */
-    eUartLastWriteFailed = IOT_UART_WRITE_FAILED, /*!< UART driver returns error when performing write operation. */
-    eUartLastReadFailed = IOT_UART_READ_FAILED,   /*!< UART driver returns error when performing read operation. */
+    eUartWriteCompleted,        /*!< UART operation write completed successfully. */
+    eUartReadCompleted,         /*!< UART operation read completed successfully. */
+    eUartLastWriteFailed,       /*!< UART driver returns error when performing write operation. */
+    eUartLastReadFailed,        /*!< UART driver returns error when performing read operation. */
 } IotUARTOperationStatus_t;
 
 /**
@@ -80,8 +81,8 @@ typedef enum
  */
 typedef enum
 {
-    eUartStopBitsOne, /*!< UART stop bits is 1. */
-    eUartStopBitsTwo, /*!< UART stop bits is 2. */
+    eUartStopBitsOne, /*!< UART stop bits is 1 bit per frame. */
+    eUartStopBitsTwo, /*!< UART stop bits is 2 bit per frame. */
 } IotUARTStopBits_t;
 
 /**
@@ -100,14 +101,14 @@ typedef void ( * IotUARTCallback_t )( IotUARTOperationStatus_t xStatus,
 /**
  * @brief The UART descriptor type defined in the source file.
  */
-struct                       IotUARTDescriptor_t;
+struct                       IotUARTDescriptor;
 
 /**
  * @brief IotUARTHandle_t is the handle type returned by calling iot_uart_open().
  *        This is initialized in open and returned to caller. The caller must pass
  *        this pointer to the rest of APIs.
  */
-typedef struct IotUARTDescriptor_t * IotUARTHandle_t;
+typedef struct IotUARTDescriptor * IotUARTHandle_t;
 
 /**
  * @brief Ioctl requests for UART HAL.
@@ -159,7 +160,7 @@ typedef struct
  *  IotUARTHandle_t xOpen;
  *  int32_t lRead, lWrite, lClose;
  *  BaseType_t xCallbackReturn;
- *  uint8_t ucPort = 1; /* Each UART peripheral will be assigned an integer.
+ *  uint8_t ucPort = 1; // Each UART peripheral will be assigned an integer.
  *
  *  xOpen = iot_uart_open( ucPort );
  *  if( xOpen != NULL )
--- a/libraries/abstractions/common_io/include/iot_uart.h
+++ b/libraries/abstractions/common_io/include/iot_uart.h
@@ -60,10 +60,10 @@
  */
 typedef enum
 {
-    eUartWriteCompleted,        /*!< UART operation write completed successfully. */
-    eUartReadCompleted,         /*!< UART operation read completed successfully. */
-    eUartLastWriteFailed,       /*!< UART driver returns error when performing write operation. */
-    eUartLastReadFailed,        /*!< UART driver returns error when performing read operation. */
+    eUartWriteCompleted,  /*!< UART operation write completed successfully. */
+    eUartReadCompleted,   /*!< UART operation read completed successfully. */
+    eUartLastWriteFailed, /*!< UART driver returns error when performing write operation. */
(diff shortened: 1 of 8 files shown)


libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -80,8 +81,8 @@ typedef enum
  */
 typedef enum
 {
-    eUartStopBitsOne, /*!< UART stop bits is 1. */
-    eUartStopBitsTwo, /*!< UART stop bits is 2. */
+    eUartStopBitsOne, /*!< UART stop bits is 1 bit per frame. */
```
dcgaws: "UART character stop is one bit."

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -80,8 +81,8 @@ typedef enum
...
+    eUartStopBitsTwo, /*!< UART stop bits is 2 bit per frame. */
```
dcgaws: "UART character stop is two bits."

libraries/abstractions/common_io/include/iot_uart.h
```diff
@@ -159,7 +160,7 @@ typedef struct
  *  IotUARTHandle_t xOpen;
  *  int32_t lRead, lWrite, lClose;
  *  BaseType_t xCallbackReturn;
- *  uint8_t ucPort = 1; /* Each UART peripheral will be assigned an integer.
+ *  uint8_t ucPort = 1; // Each UART peripheral will be assigned an integer.
```
dcgaws: Why switch to C++ comment style here? I'm fine with deferring to Uncrustify, as long as we're consistent.

libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -806,7 +807,7 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CWriteAsyncSuccess )
     IotI2CHandle_t xI2CHandle;
     int32_t lRetVal;
     uint8_t writeVal1[] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal };
-    uint8_t writeVal2[] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal, uctestIotI2CWriteVal, uctestIotI2CWriteVal };
+    uint8_t writeVal2[ IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH ] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal, uctestIotI2CWriteVal };
```
dcgaws: Is this change because you want extra space after the initialized portion of the buffer? Is there a comment somewhere that explains why?

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -318,12 +320,14 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncAssisted )
...
-        TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
+        #ifdef IOT_TEST_COMMON_IO_SPI_SLAVE_SELECT_SUPPORTED
+            lRetVal = iot_spi_select_slave( ulAssistedTestIotSpiInstance, ulAssistedTestIotSpiSlave );
+            TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
+        #endif
     }
 
     /* Create a string with read bytes and print it to console. */
-    for( int i = 0, j = 0, len = sizeof( ucRxBuf ); i < len; i++ )
+    for( i = 0, j = 0, len = sizeof( ucRxBuf ); i < len; i++ )
```
dcgaws: Could this loop better ensure that a future change doesn't cause us to write past the end of cMsg?

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -470,12 +475,14 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadAsyncAssisted )
...
```
dcgaws: Please refactor this into shared code.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -547,11 +554,12 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_WriteSyncAssisted )
     uint8_t ucTxBuf[ 16 ] = { 0 };
     char cMsg[ 50 ] = { 0 };
     size_t xBytesTx;
+    int i = 0, j = 0, len = 0;
 
     srand( xTaskGetTickCount() );
 
     /* Generate random tx bytes and load them to a string to print later. */
-    for( int i = 0, j = 0, len = sizeof( ucTxBuf ); i < len; i++ )
+    for( i = 0, j = 0, len = sizeof( ucTxBuf ); i < len; i++ )
```
dcgaws: Ditto.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -995,11 +1011,12 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_TransferAsyncAssisted )
     uint8_t ucTxBuf[ 8 ] = { 0 };
     char cMsg[ 50 ] = { 0 };
     int32_t lLoop = 0;
+    int i = 0, len = 0;
 
     srand( xTaskGetTickCount() );
 
     /* Generate random tx bytes and load them to a string to print later. */
-    for( int i = 0, len = sizeof( ucTxBuf ); i < len; i++ )
+    for( i = 0, len = sizeof( ucTxBuf ); i < len; i++ )
```
dcgaws: Do we for sure want randomness in this test? If so, unless we're seeding rand() with something like clock time, I bet that the stream is the same each time.

libraries/abstractions/common_io/include/iot_uart.h
cobusve: Uncrustify does not check examples inside of comments AFAIK. I think the problem is that C -comment style with /* cannot be nested, which means that if yo uwant to put an example in your header comment block you have to do it this way or it does not work. I am not aware of a better way to do this today.

qiutongs: Yeah, this is just comment of example code. Nested "/* */" doesn't work

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: I think we do. The tests want to send random bytes to external device.
See the line above, we already have srand( xTaskGetTickCount() )

libraries/abstractions/common_io/test/test_iot_i2c.c
qiutongs: I made this change because 4 bytes might be too large for some case (TI) 
I probably should write it to below, implying the IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH >= 2
```
uint8_t writeVal2[ IOT_TEST_COMMON_IO_I2C_MAX_WRITE_LENGTH ] = { xtestIotI2CDeviceRegister, uctestIotI2CWriteVal };
```

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: ok

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: ok

qiutongs:
Need to address comments from Dan. Will do that on feature/common-io branch

PR #1693, opened 2020-01-15
Cover different types of carriage return and modify the method to get…

cmcgcc:
… absolute path

Description
-----------
[Problem]
The carriage return style ('\r\n') some vendor used confuses current scripts. The keywords used as criteria could be parsed correctly.
The way to get absolute path can not work with the top script test_iot_assisted_tests.py.

[Solution]
Remove '\r' from the string read from serial. And take away '\r' as part of parsing keyword.
Use os.path.dirname(os.path.realpath(__file__)) instead of os.path.abspath(sys.path[0]) to get abolute path. Because sys.path[0] may not be the path of current script if the file is imported.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_scripts/adc/test_iot_adc_test.py
+++ b/libraries/abstractions/common_io/test/test_scripts/adc/test_iot_adc_test.py
@@ -24,18 +24,18 @@
 # http://www.FreeRTOS.org
 
 
-
-
 import serial
 import csv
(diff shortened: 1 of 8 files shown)


cmcgcc:
Please help for code review. @vjjunnut @cobusve @brianfr @dongh @satyaprk-amazon @qiutongs @lundinc2

PR #1694, opened 2020-01-16
Fix SegFault: ss_ctx_t context is shared and sent to a user callback,…

alfred2g:
… in the meantime if the socket is closed and subsequently freed, the callback task still holds a reference that was previously deleted

Description
-----------
The changes consists of adding a reference count to the number of users of the context, and is only freed once the count reaches zero

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
+++ b/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
@@ -116,13 +116,16 @@ static int8_t sockets_allocated = socketsconfigDEFAULT_MAX_NUM_SECURE_SOCKETS;
 #define TICK_TO_US( _t_ )    ( ( _t_ ) * 1000 / configTICK_RATE_HZ * 1000 )
 
 /*-----------------------------------------------------------*/
+
 /*
  * @brief Sockets close
  */
-static void prvSocketsClose( ss_ctx_t *ctx )
+static void prvSocketsClose( ss_ctx_t * ctx )
 {
     uint32_t ulProtocol;
+
     sockets_allocated++;
+
     /* Clean-up application protocol array. */
     if( NULL != ctx->ppcAlpnProtocols )
     {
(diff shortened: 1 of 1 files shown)


libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -114,6 +117,69 @@ static int8_t sockets_allocated = socketsconfigDEFAULT_MAX_NUM_SECURE_SOCKETS;
...
+    vPortFree( ctx );
+}
+
+/*
+ * @brief Decrement ctx refcount and call release function if the count is 1 (
+ *        last user of the ctx)
+ */
+static void prvDecrementRefCount( ss_ctx_t * ctx )
+{
+    if( Atomic_Decrement_u32( &ctx->ulRefcount ) == 1 )
```
lundinc2: If the count is at 1, we would want to close the socket, as the last remaining reference has called prvDecrementRefCount. Should this not be == 0? As Atomic_Decrement_u32( x ) , would return ( x - 1 )?

alfred2g: it would return x, then decrement x, thats why i am checking for 1
static portFORCE_INLINE uint32_t Atomic_Decrement_u32( uint32_t volatile * pAddend )
{
    uint32_t ulCurrent;

    ATOMIC_ENTER_CRITICAL();

    ulCurrent = *pAddend;

    *pAddend -= 1;

    ATOMIC_EXIT_CRITICAL();

    return ulCurrent;
}

lundinc2: Ah i understand. Thanks for the explanation!

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -220,7 +286,7 @@ static void vTaskRxSelect( void * param )
             /*ctx->rx_callback = NULL; */
```
gkwicker: Please delete commented-out code on lines 286, 288, and 296 while we're in here.

PR #1695, opened 2020-01-16
Fix advertising slave connection interval in Nordic

ravibhagavandas:
Fix advertising slave connection interval in Nordic

Description
-----------
Nordic port was not including slave connection interval range in advertisement.
PR fixes this.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_gap.c
+++ b/vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_gap.c
@@ -870,6 +870,10 @@ ret_code_t prvBTAdvDataConvert( ble_advdata_t * xAdvData,
             IotLogError( "Invalid params in advertisement/scan response: either min connection interval or max connection interval is zero" );
             xErrCode = NRF_ERROR_INVALID_PARAM;
         }
+        else
+        {
+            /* Do nothing. User have set min interval and max interval to 0 to not include this in advertisment. */
+        }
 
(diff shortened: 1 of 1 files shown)


vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_gap.c
```diff
@@ -850,8 +850,26 @@ ret_code_t prvBTAdvDataConvert( ble_advdata_t * xAdvData,
     {
         /* The first two UUID go to the initial advertisement packet */
         xAdvData->uuids_complete = xCompleteUUIDS;
-        xAdvData->p_slave_conn_int->max_conn_interval = pxParams->ulMaxInterval;
-        xAdvData->p_slave_conn_int->min_conn_interval = pxParams->ulMinInterval;
+
+        if( ( pxParams->ulMaxInterval != 0 ) && ( pxParams->ulMinInterval != 0 ) )
+        {
```
alfred2g: if both values (ulMaxInterval and ulMinInterval) are 0's, both the if and the else will not be executed.
(not sure if it is a possibility)

ravibhagavandas: if both values are zeros, it means the connection interval won't be included in advertisement, so its a no-action.

yuhui-zheng: I'd have an else branch and call it out it is intended to do nothing...

ravibhagavandas: Good callout. Added an empty else.

PR #1696, opened 2020-01-16
Fix DHCP option Client-identifier

shubhamkulkarni97:
Description
-----------
DHCP option 61 i.e. `Client-Identifier` consist of type-value pairs. In FreeRTOS TCP stack, value is assigned, while the type is missing.

Hence, type is added in Client-identifier option, which has value 0x1. `0x1` corresponds to `Ethernet`.

Checklist:
----------

- [x] My code is Linted.


mahavirj:
This is per 9.14 section of RFC at https://tools.ietf.org/html/rfc2132 and observed in the context of https://github.com/espressif/esp-idf/issues/4060

htibosch:
Hi Shubham, thank you for this PR. I tested the changes in one of my boards and indeed it looks good, and conform the RFC.
In fact you repaired a bug that has gone unnoticed for many years.
Hein

PR #1707, opened 2020-01-22
CBMC Queue Proofs -- clean up

yuhui-zheng:
Description
-----------
The overall goal is to clean up kernel proofs in this repo, before we bump kernel version. 

The proof should look like "declare variables, assumptions, and call APIs".

Checklist:
----------

- [ ] My code is Linted.


yuhui-zheng:
Dry run. This is to see if the fix is breaking other proofs, since I'm updating stubs as well.

yuhui-zheng:
Will raise the same PR towards FreeRTOS/FreeRTOS repo.

PR #1720, opened 2020-02-04
test: unit test lwip secure_sockets

alfred2g:
test: unit test lwip secure_sockets

Description
-----------
Add Unit testing framework with a cmake build 
Add Unit test with mocking to lwip secure sockets
Add CoveCoverage with gcov/lcov and html generated coverage files

Checklist:
----------

- [X] My code is Linted.

### DIFF
--- a/libraries/abstractions/posix/include/FreeRTOS_POSIX/pthread.h
+++ b/libraries/abstractions/posix/include/FreeRTOS_POSIX/pthread.h
@@ -243,7 +243,7 @@ int pthread_barrier_wait( pthread_barrier_t * barrier );
  */
 int pthread_create( pthread_t * thread,
                     const pthread_attr_t * attr,
-                    void *( *startroutine )( void * ),
+                    void * ( *startroutine )( void * ),
                     void * arg );
 
 /**
--- a/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
+++ b/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
@@ -282,7 +282,6 @@ static void vTaskRxSelect( void * param )
 
         if( FD_ISSET( s, &read_fds ) )
         {
-            configASSERT( ctx->rx_callback );
             ctx->rx_callback( ( Socket_t ) ctx );
         }
     }
@@ -290,6 +289,7 @@ static void vTaskRxSelect( void * param )
     prvDecrementRefCount( ctx );
     vTaskDelete( NULL );
 }
+
 
 /*-----------------------------------------------------------*/
 
@@ -400,7 +400,6 @@ int32_t SOCKETS_Connect( Socket_t xSocket,
     struct sockaddr_in sa_addr = { 0 };
     int ret;
 
-    sa_addr.sin_family = pxAddress->ucSocketDomain ? pxAddress->ucSocketDomain : AF_INET;
     sa_addr.sin_addr.s_addr = pxAddress->ulAddress;
     sa_addr.sin_port = pxAddress->usPort;
 
@@ -447,12 +446,14 @@ int32_t SOCKETS_Connect( Socket_t xSocket,
         }
         else
         {
-            configPRINTF( ( "TLS_Connect fail (0x%x, %s)\n", ( unsigned int ) -status, ctx->destination ? ctx->destination : "NULL" ) );
+            configPRINTF( ( "TLS_Connect fail (0x%x, %s)\n",
+                            ( unsigned int ) -status,
+                            ctx->destination ? ctx->destination : "NULL" ) );
         }
     }
     else
     {
-        configPRINTF( ( "Invalid ip socket\n" ) );
+        configPRINTF( ( "LwIP connect fail %d %d\n", ret, errno ) );
     }
 
     return SOCKETS_SOCKET_ERROR;
@@ -569,7 +570,6 @@ int32_t SOCKETS_Close( Socket_t xSocket )
 {
     ss_ctx_t * ctx;
 
-
     if( SOCKETS_INVALID_SOCKET == xSocket )
     {
         return SOCKETS_EINVAL;
@@ -579,7 +579,6 @@ int32_t SOCKETS_Close( Socket_t xSocket )
     ctx->state = SST_RX_CLOSING;
 
     lwip_close( ctx->ip_socket );
-
     prvDecrementRefCount( ctx );
 
     return SOCKETS_ERROR_NONE;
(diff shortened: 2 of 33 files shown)


libraries/abstractions/posix/include/FreeRTOS_POSIX/pthread.h
```diff
@@ -243,7 +243,7 @@ int pthread_barrier_wait( pthread_barrier_t * barrier );
  */
 int pthread_create( pthread_t * thread,
                     const pthread_attr_t * attr,
-                    void *( *startroutine )( void * ),
+                    void * ( *startroutine )( void * ),
```
cobusve: Did this change come out of uncrustify? I do not like changing the pthread code in a PR about units tests, do we really need this change in this PR?

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -210,11 +210,6 @@ static BaseType_t prvNetworkRecv( void * pvContext,
 
     ctx = ( ss_ctx_t * ) pvContext;
 
-    if( 0 > ctx->ip_socket )
-    {
-        return SOCKETS_SOCKET_ERROR;
-    }
-
     int ret = lwip_recv( ctx->ip_socket,
```
cobusve: Please remind me why we had to remove these? If I recall this was dead/unreachable code right?
I would actually prefer if we can separate out the fixes to the code from the adding of the tests but it may be too late to do that easily? 

Think about the case where we find out this change was bad, if we want to roll it back we will now have to roll back the entire PR.

Also if the PR does not change the code, only tests it is low risk to include it in master, but if it changes code the risk is much higher and this will hold up the review process.

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
```diff
@@ -0,0 +1,1334 @@
+/*
+ * Amazon FreeRTOS HTTPS Client V1.1.1
```
cobusve: Is this version number in this header correct? Does not look like it?

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
```diff
@@ -0,0 +1,1334 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://aws.amazon.com/freertos
+ * http://www.FreeRTOS.org
+ */
+#include <stdbool.h>
+#include <pthread.h>
+
```
cobusve: Does this always depend on pthread?

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
```diff
@@ -0,0 +1,1334 @@
...
+    connectSocket( s );
+    return s;
+}
+/* ======================  TESTING SOCKETS_Send  ============================ */
+
+/*!
+ * @brief A happy send case with normal sockets
+ */
+void test01_SecureSockets_send_successful( void )
+{
```
cobusve: I want to make sure we establish patterns here that in the long run will be stable as these will become examples of how to do it. Do you think it is wise to have a test number as part of the test name like this? Keeping these numbers sequential and tracable seems to be hard e.g. when I delete a test in the middle somewhere, does that require me to rename all the tests? If not will it not look like there is a missing test?

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
```diff
@@ -0,0 +1,1334 @@
...
+    TEST_ASSERT_EQUAL_INT( SOCKETS_ENOTCONN, ret );
+    deinitSocket( so );
+}
+
+/* =====================  TESTING SOCKETS_Socket  =========================== */
+
+/*!
+ * @brief calling SOCKETS_Socket with various invalid arguments
+ *        expects SOCKETS_INVALID_SOCKET
+ */
```
cobusve: In general I would like to see us increase the verbosity of the description of the test. Specifically I would like to explicitly have a section for every test that describes "Purpose of this test"

libraries/c_sdk/aws/shadow/include/types/aws_iot_shadow_types.h
```diff
@@ -471,7 +471,7 @@ typedef struct AwsIotShadowDocumentInfo
              * This only needs to be set if #AWS_IOT_SHADOW_FLAG_WAITABLE is passed to
              * @ref shadow_function_get.
              */
-            void *( *mallocDocument )( size_t );
+            void * ( *mallocDocument )( size_t );
         } get; /**< @brief Valid members for @ref shadow_function_get. */
```
cobusve: Same for this and a couple of other files, is this an uncrustify problem? This does bloat this PR with lots of files?

libraries/abstractions/posix/include/FreeRTOS_POSIX/pthread.h
alfred2g: This change is to simulate better the environment, since the secure_sockets code is calling xTaskCreate to create a thread to notify the library user of changes on the descriptor.
I tried to make it simple before, instead of launching a thread, i just call the function, but i was impossible to control without changing the code itself, and the address sanitizers complained about usage after free.

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
alfred2g: this condition always evaluate to true
i created another PR to address it

libraries/abstractions/posix/include/FreeRTOS_POSIX/pthread.h
alfred2g: i removed the dependency over pthread by creating a library, and will have to be implemented per supported platform, current implementation is for linux with pthread

libraries/c_sdk/aws/shadow/include/types/aws_iot_shadow_types.h
alfred2g: Yes our uncrustify config file doesn't like the * to be attached to anything, it depends on the version i was using as well 1.7 instead of 1.6

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
alfred2g: i recopied it from secure_sockets.c

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
alfred2g: nope, just removed this dependency, by creating a wrapper for linux

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
alfred2g: it was easier for me to find failed tests, but super annoying to renumber them
i removed the numbers..

libraries/abstractions/secure_sockets/utest/secure_sockets_utest.c
alfred2g: I will try to find an appropriate doxygen tag for that and add more verbose descriptions

PR #1727, opened 2020-02-05
Add sdio apis to common io.

qizengqizeng:
Add sdio apis to common io.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/include/iot_sdio.h
+++ b/libraries/abstractions/common_io/include/iot_sdio.h
@@ -1,0 +1,720 @@
+/*
+ * Amazon FreeRTOS Common IO V1.0.0
+ * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://aws.amazon.com/freertos
+ * http://www.FreeRTOS.org
+ */
+
+
+/**
+ * @file    iot_sdio.h
+ * @brief   This file contains all the SDIO HAL API definitions.
+ *          Here we define SDIO driver as a protocol layer driver which builds on top of SDMMC
+ *          host driver. SDMMC host driver operates at physical layer and provide APIs to send
+ *          commands to the slave device and receive command responses, send and receive data,
+ *          and handle error conditions on the bus.
+ *
+ *          SD/SDIO/MMC driver can be built on top of SDMMC host. This interface only define
+ *          the APIs for SDIO applications, the interface can be expanded to support SD/MMC either
+ *          by expanding the APIs in this file or adding separate files for SD/MMC
+ *
+ * @design consideration
+ *          SDIO HAL APIs performs protocol layer tasks between sdio host and sdio card.
+ *          Some APIs perform a single task and some APIs perform a sequence of tasks.
+ *          For examples, iot_sdio_io_read_direct() API performs a single task of reading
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/include/iot_sdio.h
```diff
@@ -0,0 +1,720 @@
...
+ * @{
+ */
+
+/**
+ * @brief The return codes for the functions in SDIO
+ */
+ /*!< SDIO operation completed successfully. */
+#define IOT_SDIO_SUCCESS                             ( 0 )
+/*!< SDIO operation failed. */
+#define IOT_SDIO_FAIL                                ( 1 )
```
qiutongs: In which case, is this general failure error code returned?

libraries/abstractions/common_io/include/iot_sdio.h
```diff
@@ -0,0 +1,720 @@
...
+#define IOT_SDIO_SWITCH_HIGH_SPEED_FAIL              ( 5 )
+/*!< If the card is initialized in 3.3V signaling,
+     and card support high speed mode (SHS ==1),
+     host can issue CMD52 in RAW mode, setting EHS
+     bit in CCCR to one to switch to high speed mode
+     . This error code indicates that the sdio card
+     does not supports high speed mode. The bus clock
+     rate will be set to the default 25MHz in this
+     case. */
+#define IOT_SDIO_HIGH_SPEED_MODE_NOT_SUPPORTED       ( 6 )
```
qiutongs: We can replace this with a generic "NOT SUPPORTED" error code.

libraries/abstractions/common_io/include/iot_sdio.h
```diff
@@ -0,0 +1,720 @@
...
+ *                                               to eSdioFunction7.
+ * @param[in]   xIOFunctionCallback              IOFunctionCallback to be set for I/O function eFunc.
+ * @param[in]   pvIOFunctionCallbackUserContext  User context to be passed when IOFunctionCallback is
+ *                                               invoked.
+ *
+ * @return
+ *   - IOT_SDIO_SUCCESS on success.
+ *   - IOT_SDIO_INVALID_VALUE if any parameter is invalid.
+ */
+int32_t iot_sdio_set_io_function_callback(IotSdioSlotHandle_t const pxSdioHandle,
```
qiutongs: What are some use cases that applications do in side this IotSdioIOFunctionCallback_t callback?

libraries/abstractions/common_io/include/iot_sdio.h
```diff
@@ -0,0 +1,720 @@
...
+ * @param[in]   pxSdioHandle         Handle to sdio card slot instance returned in
+ *                                   iot_sdio_open().
+ * @param[out]  bpCardPresenceStatus True if card is present.
+ *                                   False is card is not present.
+ *
+ * @return
+ *   - IOT_SDIO_SUCCESS on successful check.
+ *   - IOT_SDIO_INVALID_VALUE if any parameter is invalid.
+ */
+int32_t iot_sdio_check_card_presence_status(IotSdioSlotHandle_t const pxSdioHandle,
```
qiutongs: This seems fit into ioctl to me.

libraries/abstractions/common_io/include/iot_sdio.h
qizengqizeng: It is used in IOCTL. I updated the doc for IOCTL.

libraries/abstractions/common_io/include/iot_sdio.h
qizengqizeng: This is needed since it is one of the specific reasons why switch speed could fail.  This is different from API not supported on certain platform implementation.  I will add IOT_SDIO_NOT_SUPPORTED for that case.

libraries/abstractions/common_io/include/iot_sdio.h
qizengqizeng: In this callback, each I/O function specific application can perform I/O function specific handling of transfer completed IRQ.

libraries/abstractions/common_io/include/iot_sdio.h
qizengqizeng: will make ioctl.

libraries/abstractions/common_io/include/iot_sdio.h
qiutongs: No, in iot_sdio_ioctl, when some request is not supported, we return generic "NOT SUPPORT". This eSDIOSwitchToHighSpeed also applies.

I can understand IOT_SDIO_SWITCH_HIGH_SPEED_FAIL is a specific one for high speed mode but "NOT SUPPORT" is not

libraries/abstractions/common_io/include/iot_sdio.h
qiutongs: I see.

libraries/abstractions/common_io/include/iot_sdio.h
qizengqizeng: How do I distinguish the following:
1.  Older SD controller don't support high speed at all.
2. Newer SD controller support high speed, but the SDIO Card inserts doesn't support high speed.
3. For whatever reason, the implementer decide to not to implement this API and simply want to return "NOT_SUPPORTED"

Do I use the same return code for all 3 scenarios?  It don't seem to help debugging very much.

qiutongs: I see your point now. I didn't think from the point of debugging. 
My opinion is that all the 3 cases are same to users. Do users need to distinguish or debug? 
I will let you make the decision.

qizengqizeng: I will change "IOT_SDIO_HIGH_SPEED_MODE_NOT_SUPPORTED" to "IOT_SDIO_LOW_SPEED_CARD"

qizengqizeng:
Unable to amend.  Abandoned and create new one.

qiutongs:
> Unable to amend. Abandoned and create new one.

When you amend the commit, it overrides the existing commit and then your local and remote diverges. That’s why you do a force push to override remote as well.  
A better approach is to just do normal commit in the PR. To clean the commits history, we do “squash and merge” as the final step.

PR #1730, opened 2020-02-06
Change uxPendedTicks to xPendedTicks in all proofs

karkhaz:
This ensures that CBMC proofs are able to build after the name-change in commit 23839bb.


yuhui-zheng:
Apologies... Didn't realize we have proofs to be fixed as well...

yuhui-zheng:
I'm going to close this PR and continue your work in https://github.com/aws/amazon-freertos/pull/1754.

PR #1735, opened 2020-02-07
Reduce Microchip C Runtime Heap Size and Increase FreeRTOS Heap Size

lundinc2:
This change is necessary because FreeRTOS based projects should not use the C runtime heap. Instead
they should use the FreeRTOS managed Heap. I have changed the C runtime heap allocation from
170KB to 10KB, and in turn increased the FreeRTOS heap size by 160KB.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.


gkwicker:
This is failing microchip builds.

lundinc2:
@gkwicker I have fixed the regression.

PR #1736, opened 2020-02-07
Change name from "Amazon FreeRTOS" to "FreeRTOS"

navinns:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


demos/demo_runner/iot_demo_runner.c
```diff
@@ -1,6 +1,6 @@
 /*
- * Amazon FreeRTOS V201912.00
- * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ * FreeRTOS V201912.00
```
lundinc2: Shouldn't the version number get changed as well?

tgsong: version number is supposed to be updated by another tool.

lundinc2: Ah  I see.

sbammi:
Please change "Amazon FreeRTOS Qualifcation Program (AFQP)" to "Device Qualification Program for FreeRTOS" This will match with the change of test names from AFQP_xx to DQP_FR_xx. Thanks

demos/demo_runner/iot_demo_runner.c
cobusve: I think if we are going to change every file in the repo it is better to do it once than to do it twice. Lets update the version number as well in this repo. It will never be possible to separate these 2 changes as they all affect the same lines ...

PR #1746, opened 2020-02-11
Cypress projects kernel dependency fix.

yuhui-zheng:
Description
-----------
In previous releases, we "accidentally" committed a kernel port in kernel third party directory. This is unnecessary since third party port is essentially the same as ```GCC/ARM_CRx_No_GIC```. In this kernel release, we removed the redundant third party port. And in this PR, I'm trying to make the projects take dependency on the correct kernel port. 

Additional compiler flags are added for floating point related. Refer to https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html.

Checklist:
----------

- [ ] My code is Linted.


yuhui-zheng:
Will reach out Cypress later in regards to the fix proposed in this PR.

PR #1749, opened 2020-02-12
Fixed Hard Fault in ST and TI. Reverted secure element key generation regression.

lundinc2:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
+++ b/demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
@@ -1028,10 +1028,6 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
                 configPRINTF( ( "Warning: could not clean-up old crypto objects. %d \r\n", xResult ) );
             }
         }
-        else
-        {
-            xResult = CKR_ARGUMENTS_BAD;
-        }
     #endif /* if ( pkcs11configIMPORT_PRIVATE_KEYS_SUPPORTED == 1 ) */
 
     /* If a client certificate has been provided by the caller, attempt to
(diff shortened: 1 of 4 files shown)


vendors/st/boards/stm32l475_discovery/aws_demos/config_files/iot_pkcs11_config.h
```diff
@@ -83,6 +83,15 @@
...
 #define pkcs11configJITP_CODEVERIFY_ROOT_CERT_SUPPORTED    0
 
+/**
+ * @brief Disable ECDSA crypto algorithm (i.e. mechanism) support in this
+ * PKCS #11 module.
+ *
+ * Set this to 1 when the host doesn't support ECDSA for network authentication
+ * (e.g. in the TLS protocol implementation).
+ */
+#define pkcs11configSUPPRESS_ECDSA_MECHANISM               1
```
dcgaws: For TI, below, you've got the equivalent change for Test as well as Demo. But for STM32, you've only got it for Demo. Is that intentional?

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
```diff
@@ -1028,10 +1028,6 @@ CK_RV xProvisionDevice( CK_SESSION_HANDLE xSession,
                 configPRINTF( ( "Warning: could not clean-up old crypto objects. %d \r\n", xResult ) );
             }
         }
-        else
```
dcgaws: The change that's being reverted here was an attempt to fix another bug, correct? Do you have an alternative fix for that other bug?

vendors/st/boards/stm32l475_discovery/aws_demos/config_files/iot_pkcs11_config.h
lundinc2: ST already had this configuration, that is why this issue did not exist on the test project of ST.

demos/dev_mode_key_provisioning/src/aws_dev_mode_key_provisioning.c
lundinc2: In hindsight this was not the right thing to do, I fixed the bug accidentally, but this code was correct. The real issue was actually the call to generatekey, which tried to generate an ECDSA key on ST and TI, causing a hard fault.

PR #1763, opened 2020-02-17
Test: Fix wrong no of bytes assert in spi assisted tests

cmcgcc:
Description
-----------
Why: The byte count in the no of bytes assertion is incorrect.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_spi.c
+++ b/libraries/abstractions/common_io/test/test_iot_spi.c
@@ -311,7 +311,7 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncAssisted )
 
         lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetRxNoOfbytes, &xBytesRx );
         TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
-        TEST_ASSERT_EQUAL( xBytesRx, 4 );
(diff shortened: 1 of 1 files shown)


cmcgcc:
Please help for code review. @vjjunnut @cobusve @brianfr @dongh @qiutongs @lundinc2 @haipengd

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -311,7 +311,7 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_ReadSyncAssisted )
 
         lRetVal = iot_spi_ioctl( xSPIHandle, eSPIGetRxNoOfbytes, &xBytesRx );
         TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
-        TEST_ASSERT_EQUAL( xBytesRx, 4 );
+        TEST_ASSERT_EQUAL( xBytesRx, sizeof( ucRxBuf ) );
```
dan4thewin: It looks to me like this changes an assertion that the variable is 4 to an assertion that the variable is 16.  Unless I'm reading this wrong, the assert either had to be failing before, or must fail with this change.  Have you run this and seen all test cases passing?

cmcgcc: Yes. It is verified. It was commented in the internal branch because the api was not ready. So the wrong count of bytes was found out in the first check in.

PR #1776, opened 2020-02-20
Free mbedtls resources in TLS_Cleanup().

gkwicker:
Description
-----------

`TLS_Cleanup()` was not calling `prvFreeContext()` for unsuccessful TLS handshakes.  This can cause a memory leak if the necessary credentials [can't be loaded](https://github.com/aws/amazon-freertos/blob/master/libraries/freertos_plus/standard/tls/src/iot_tls.c#L553).  This change ensures that mbedtls resources are freed on `TLS_Cleanup()` even if the handshake didn't complete.  Note that PKCS11 resources are freed independently in the case where [`C_OpenSession()` fails](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c#L1028).

Checklist:
----------

- [X] I have tested my changes. No regression in existing tests (tested on esp32 with Full_TCP)
- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/tls/src/iot_tls.c
+++ b/libraries/freertos_plus/standard/tls/src/iot_tls.c
@@ -129,7 +129,8 @@ static void prvFreeContext( TLSContext_t * pxCtx )
             ( CK_INVALID_HANDLE != pxCtx->xP11Session ) )
         {
             pxCtx->pxP11FunctionList->C_CloseSession( pxCtx->xP11Session ); /*lint !e534 This function always return CKR_OK. */
-	}
+        }
+
         pxCtx->xTLSHandshakeSuccessful = pdFALSE;
     }
 }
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/standard/tls/src/iot_tls.c
```diff
@@ -122,9 +122,11 @@ static void prvFreeContext( TLSContext_t * pxCtx )
         mbedtls_ssl_free( &pxCtx->xMbedSslCtx );
         mbedtls_ssl_config_free( &pxCtx->xMbedSslConfig );
 
-        /* Cleanup PKCS#11. */
-        if( ( NULL != pxCtx->pxP11FunctionList ) &&
-            ( NULL != pxCtx->pxP11FunctionList->C_CloseSession ) )
+        /* Cleanup PKCS11 only if the handshake succeeded. */
+        if( ( pdTRUE == pxCtx->xTLSHandshakeSuccessful ) &&
```
dcgaws: prvFreeContext can be called as a result of a handshake error. In that case, there will be a P11 context, but HandshakeSuccessful will be false.

PR #1777, opened 2020-02-21
Explicit typecast of UL to (void*)

AniruddhaKanhere:
Explicit typecast of UL to (void*)

Description
-----------
NOTE: This is specific to Microchip (Not sure about others though). The timer_t defined in the library finally converges to unsigned long.
Initialization was making pointer from integer without any kind of typecasting. This makes the typecast ambiguous and dependent on the compiler.
* Explicitly casted the variables to ( void * )
* Cannot cast them directly to TimerHandle_t since during compile time the structure itself may not be defined. Just a placeholder for the structure exists in the header-file timer.h during compile time.
* Uncrustify

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_posix/source/FreeRTOS_POSIX_timer.c
+++ b/libraries/freertos_plus/standard/freertos_plus_posix/source/FreeRTOS_POSIX_timer.c
@@ -152,11 +152,11 @@ int timer_create( clockid_t clockid,
 
 int timer_delete( timer_t timerid )
 {
-    /* Cannot use TimerHandle_t for casting since the structure itself is not defined when 
+    /* Cannot use TimerHandle_t for casting since the structure itself is not defined when
      * this file is being compiled. There exists just a place-holder for the structure in
      * the header file timers.h */
     TimerHandle_t xTimerHandle = ( void * ) timerid;
-    
+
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/standard/freertos_plus_posix/source/FreeRTOS_POSIX_timer.c
```diff
@@ -152,7 +152,11 @@ int timer_create( clockid_t clockid,
 
 int timer_delete( timer_t timerid )
 {
-    TimerHandle_t xTimerHandle = timerid;
+    /* Cannot use TimerHandle_t for casting since the structure itself is not defined when
+     * this file is being compiled. There exists just a place-holder for the structure in
+     * the header file timers.h */
+    TimerHandle_t xTimerHandle = ( void * ) timerid;
```
AniruddhaKanhere: The comment explains the choice for using (void*) instead of TimerHandle_t for typecasting.

cobusve: I do not think you can create an auto variable like this without the type being complete, so I do not think this is right ... I would expect that you would cast there to TimerHandle_t, if you can create the variable you should be able to do the cast.

PR #1778, opened 2020-02-21
Address issues in NXP port

gkwicker:
- Serialize all access to the WiFi module via semaphore
- Toggle WiFi power at startup
- Decrease secure socket test timeout to 10s

Description
-----------

Checklist:
----------

- [X] I have tested my changes. No regression in existing tests (Ran Full_TCP on NXP LPC54018 IoT Module)
- [ ] My code is Linted.

### DIFF
--- a/vendors/nxp/boards/lpc54018iotmodule/ports/wifi/iot_wifi.c
+++ b/vendors/nxp/boards/lpc54018iotmodule/ports/wifi/iot_wifi.c
@@ -40,6 +40,10 @@
  */
 #define wifiMAX_DNS_NAME_LENGTH    253
 
+/* Define the delay used when toggling the WiFi shield power.
+ */
+#define wifiSHIELD_POWER_TOGGLE_TIME_MSEC    500
+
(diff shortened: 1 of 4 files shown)


vendors/nxp/boards/lpc54018iotmodule/ports/wifi/iot_wifi.c
```diff
@@ -307,43 +344,66 @@ WIFIReturnCode_t WIFI_On( void )
...
+    result = ( A_STATUS ) WIFISHIELD_Init();
+
+    if( A_OK != result )
+    {
         return eWiFiFailure;
+    }
+
+    /* Power off the WLAN and wait 500ms */
+    WIFISHIELD_PowerUp( 0 );
+    vTaskDelay( MSEC_TO_TICK( 500 ) );
```
dcgaws: Since we're changing their hard-coded delay value, should we also change it to a #define?

gkwicker: Good idea, will change.

PR #1783, opened 2020-02-24
Implement Common IO on ST, for UART, SPI, I2C

qiutongs:
Description
-----------

Implement Common IO on ST, for UART, SPI, I2C

Checklist:
----------

- [ ] My code is Linted.


vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+{
+    IotI2CDescriptor_t * pI2cDescriptor = pxI2CPeripheral;
+    HAL_StatusTypeDef halStatus = HAL_OK;
+
+    if( ( pI2cDescriptor == NULL ) ||
+        ( _isFlagNotSet( pI2cDescriptor, _FLAG_OPENED ) ) ||
+        ( pucBuffer == NULL ) ||
+        ( xBytes == 0 ) )
+    {
+        return IOT_I2C_INVALID_VALUE;
```
dcgaws: Are we trying to restrict ourselves to a single return for MISRA compliance?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+            break;
+
+        case eI2CSetMasterConfig:
+
+            if( pvBuffer == NULL )
+            {
+                return IOT_I2C_INVALID_VALUE;
+            }
+
+            /* TODO: here it assumes the state is busy if it is not ready, but there it may be in other state. */
```
dcgaws: Where is "there"?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+/*-------------ST API Implementation Overrides---------------*/
+
+/**
+ * @brief Initializes low level hardware
+ */
+void HAL_I2C_MspInit( I2C_HandleTypeDef * hi2c )
+{
+    GPIO_InitTypeDef xGpio_init_structure;
+
+    /* Configure the GPIOs */
```
dcgaws: Coding style calls for comments to be complete sentences.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+
+    DISCOVERY_I2Cx_CLK_DISABLE();
+}
+
+void HAL_I2C_MasterTxCpltCallback( I2C_HandleTypeDef * hi2c )
+{
+    IotI2CDescriptor_t * pDescriptor = _toDescriptor( hi2c );
+    BaseType_t higherPriorityTaskWoken = pdFALSE;
+
+    /* Subtract not transferred yet. */
```
dcgaws: This is confusing. We're in a transmit-complete callback, but we're decreasing the number of bytes that have been transferred. Why?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+            return operation == _I2C_READ_OP ? IOT_I2C_READ_FAILED : IOT_I2C_WRITE_FAILED;
+
+        case HAL_BUSY:
+            return IOT_I2C_BUSY;
+
+        case HAL_TIMEOUT:
+            return IOT_I2C_BUS_TIMEOUT;
+    }
+
+    /* TODO: it is better to assert(0) here. */
```
dcgaws: Why?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+    return IOT_I2C_SUCCESS;
+}
+
+static int32_t _doMasterTransfer( IotI2CDescriptor_t * pI2cDescriptor,
+                                  uint8_t * const pucBuffer,
+                                  size_t xBytes,
+                                  HAL_StatusTypeDef ( * pMasterSync )( I2C_HandleTypeDef * hi2c,
```
dcgaws: Why aren't these function prototypes based on shared definitions?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+        if( ongoingTrans )
+        {
+            /* There is an ongoing transaction; end the transaction. */
+            halStatus = pMasterSequentialIT( &( pI2cDescriptor->xHandle ), pI2cDescriptor->usSlaveAddr, pucBuffer, xBytes, I2C_LAST_FRAME );
+        }
+        else
+        {
+            if( pMasterSync != NULL )
+            {
+                /* No ongoing transaction; simple make a blocking call. */
```
dcgaws: "simply"

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,679 @@
...
+                    }
+                    else
+                    {
+                        LL_SPI_SetTransferBitOrder( pxSpi->Instance, SPI_FIRSTBIT_LSB );
+                    }
+
+                    switch( ( ( IotSPIMasterConfig_t * ) pvBuffer )->eMode )
+                    {
+                        case eSPIMode0:
+                            LL_SPI_SetClockPolarity( pxSpi->Instance, SPI_POLARITY_LOW );
```
dcgaws: Could these blocks each be collapsed into one call instead of two via logical OR?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,679 @@
...
+    }
+    else if( ( pvBuffer == NULL ) || ( xBytes == 0 ) )
+    {
+        lError = IOT_SPI_INVALID_VALUE;
+    }
+    else
+    {
+        SPI_HandleTypeDef * pxSpi = pxSPIPeripheral->pxSpiContext->pxSpi;
+
+        if( HAL_SPI_GetState( pxSpi ) == HAL_SPI_STATE_BUSY_TX )
```
dcgaws: Since it's synchronous, should it try to block until the state is no longer busy?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,679 @@
...
+    {
+        SPI_HandleTypeDef * pxSpi = pxSPIPeripheral->pxSpiContext->pxSpi;
+
+        if( HAL_SPI_Abort( pxSpi ) != HAL_OK )
+        {
+            lError = IOT_SPI_BUS_BUSY;
+        }
+        else if( HAL_SPI_DeInit( pxSpi ) != HAL_OK )
+        {
+            lError = IOT_SPI_BUS_BUSY;
```
dcgaws: If this happens, is the app expected to retry this API call?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,679 @@
...
+        }
+    }
+
+    return lError;
+}
+/*-----------------------------------------------------------*/
+
+void HAL_SPI_TxRxCpltCallback( SPI_HandleTypeDef * hspi )
+{
+    if( hspi->Instance == SPI1 )
```
dcgaws: Rather than require these if/else blocks, it seems like the respective user context pointer should be in the SPI instance handle structure.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,689 @@
...
+{
+    UART_HandleTypeDef * pxHuart; /**< STM UART handle for all the configuration. */
+    IRQn_Type eIrqNum;            /**< Interrupt number for the UART. */
+} STM32_UART_HalContext_t;
+
+/**
+ * @brief Statically initialized map of STM UART Handle for all 5 ports.
+ *
+ */
+static UART_HandleTypeDef xUartHandleMap[] =
```
dcgaws: This initialization could be done in approximately 80% fewer lines.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,689 @@
...
+IotUARTHandle_t iot_uart_open( int32_t lUartInstance )
+{
+    IotUARTHandle_t xHandle = NULL;
+    UART_HandleTypeDef * xHandleHAL = NULL;
+
+    if( ( lUartInstance < IOT_USART_PORTS ) && ( lUartInstance >= 0 ) )
+    {
+        xHandle = pxUarts[ lUartInstance ];
+
+        /*lint !e9079 !e9087 handle passed should be of UART_HandleTypeDef *. */
```
dcgaws: Looks like xHandle needs to be of type IoTUARTHandle_t.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,689 @@
...
+    if( ( pxUartPeripheral == NULL ) || ( pxUartPeripheral->sOpened != IOT_UART_OPENED ) )
+    {
+        lError = IOT_UART_INVALID_VALUE;
+    }
+    else
+    {
+        /*lint !e9079 !e9087 handle passed should be of UART_HandleTypeDef *. */
+        UART_HandleTypeDef * pxHuart = ( UART_HandleTypeDef * ) ( ( STM32_UART_HalContext_t * ) pxUartPeripheral->pvHalContext )->pxHuart;
+
+        /* Abort always returns HAL_OK */
```
dcgaws: What if the underlying driver implementation changes that behavior at some point?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,689 @@
...
+        }
+    }
+
+    return lError;
+}
+/*-----------------------------------------------------------*/
+
+void HAL_UART_ErrorCallback( UART_HandleTypeDef * huart )
+{
+    /* Propogate Error message back to user.
```
dcgaws: "Propagate error..."

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,689 @@
...
+
+        for( ; i < sizeof( xUartHandleMap ) / sizeof( UART_HandleTypeDef ); i++ )
+        {
+            if( huart->Instance == xUartHandleMap[ i ].Instance )
+            {
+                if( pxUarts[ i ]->xUartCallback != NULL )
+                {
+                    pxUarts[ i ]->xUartCallback( huart->ErrorCode, pxUarts[ i ]->pvUserCallbackContext );
+                }
+
```
dcgaws: It's not necessary to give the semaphore back, as per to the two ISR callbacks below?

vendors/st/boards/stm32l475_discovery/ports/common_io/test_iot_internal.c
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * IOT On-Target Unit Test Internal - LPC54018
```
dcgaws: Why does this say LPC?

vendors/st/boards/stm32l475_discovery/aws_tests/config_files/test_iot_config.h
```diff
@@ -0,0 +1,91 @@
+/*
```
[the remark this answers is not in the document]
qiutongs: Yes, This is file is Common I/O specific. And I agree with you. It was just renaming it might break code in ACE team. Let me check with them.

vendors/st/boards/stm32l475_discovery/aws_tests/config_files/test_iot_config.h
```diff
@@ -0,0 +1,91 @@
...
+ * Amazon FreeRTOS Common IO V1.0.0
+ * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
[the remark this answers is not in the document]
qiutongs: Yes, let me fix.

vendors/st/boards/stm32l475_discovery/aws_tests/config_files/test_iot_config.h
```diff
@@ -0,0 +1,91 @@
...
+#define I2C_TEST_SET                                         1
+
+/* 0xD4 is the slave address for on-board sensor. */
+const uint8_t i2cTestSlaveAddr[ I2C_TEST_SET ] = { 0xD4 };
+const uint8_t i2cTestDeviceRegister[ I2C_TEST_SET ] = { 0x73 };
+const uint8_t i2cTestWriteVal[ I2C_TEST_SET ] = { 0b01101010 };
+const uint8_t i2cTestInstanceIdx[ I2C_TEST_SET ] = { 1 };
+const uint8_t i2cTestInstanceNum[ I2C_TEST_SET ] = { 3 };
+
+/* Not used by tests in AFR. */
```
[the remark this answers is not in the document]
qiutongs: Yes, let me remove

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+#include "stm32l475e_iot01.h"
+
+/* Main includes. */
+#include "iot_i2c.h"
+
+/* FreeRTOS includes. */
+#include "semphr.h"
+
+/* Total number of I2C instances on this ST board. */
+#define _I2C_INSTANCE_NUM       ( 3 )
```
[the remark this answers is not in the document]
qiutongs: I might mixing the coding style of C-SDK where the naming of variable/macro accessible only in this file is prefixed with "_".

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+/* Used by the "flags" member of IotI2CDescriptor_t. */
+#define _FLAG_OPENED            ( 0x01 ) /* If the instance is already opened. */
+#define _FLAG_SLAVE_ADDRESS     ( 0x02 ) /* If the slave address has been set. */
+#define _FLAG_NO_STOP           ( 0x04 ) /* If the no stop has been requested. */
+
+/* Initializer of the "flags" member of IotI2CDescriptor_t. */
+#define _FLAG_INITIALIZER       ( 0 )
+
+/* Initial state if slave address is unset. */
+#define _UNSET_SLAVE_ADDRESS    ( 0xFFFF )
```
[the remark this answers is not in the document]
qiutongs: Sounds better. let me fix

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+    IotI2CCallback_t xCallback;    /* Callback function */
+    void * pvUserContext;          /* User context passed in callback */
+    SemaphoreHandle_t xSemphr;
+    uint16_t usSlaveAddr;          /* Slave Address */
+    uint16_t usTransmittedTxBytes; /* Number of Transmitted Bytes */
+    uint16_t usReceivedRxBytes;    /* Number of Received Bytes */
+    uint8_t flags;                 /* Bit flags to tract different states. */
+} IotI2CDescriptor_t;
+
+static IotI2CDescriptor_t _i2cContexts[ _I2C_INSTANCE_NUM ] =
```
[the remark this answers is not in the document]
qiutongs: If _I2C_INSTANCE_NUM is set to wrong value, then compile should warn. 
But maybe _I2C_INSTANCE_NUM constant is redundant. Probably I should do sizeof(_i2cContexts) / sizeof(IotI2CDescriptor_t)

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: I didn't follow some MISRA rules. Good callout. Let me create a SIM to track this issue.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Rephrased it.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Because XferCount is NOT the number of bytes that have been transferred. It is the number of bytes still left, not yet transferred. That is just how ST driver tracks state.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: I was thinking since _toI2cStatus is a private function, all the cases of "halStatus" parameter should be covered by the switch block above. So this line of code should be not reachable.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: This is defined by ST HAL, not controlled by us.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Thank you for reviewing this so carefully!

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: Actually, I feel this is a bug. The second call should be LL_SPI_SetClockPhase instead of LL_SPI_SetClockPolarity.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: This question made me think. I believe that is not the design goal of Common IO. We leave this decision to the upper layer.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: After second check, HAL_SPI_DeInit doesn't fail. I removed this if condition.

vendors/st/boards/stm32l475_discovery/ports/common_io/test_iot_internal.c
qiutongs: Copy-paste error. Fixed.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: Yes.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: I think I missed it. Fixed.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: Then our code might break in failed case. What do you suggest in such case? Should we handle the error code any way to be future-proof?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: This comment is confusing. Let me remove it.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: Refactoring...

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+ * @file iot_i2c.c
+ * @brief HAL i2c implementation on STM32L4 Discovery Board
+ */
+
+/* The config header is always included first. */
+#include "iot_config.h"
+
+/* ST Board includes. */
+#include "stm32l475xx.h"
+#include "stm32l4xx_hal.h"
```
lundinc2: Do we need all of these includes?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,677 @@
...
+ */
+
+/**
+ * @file iot_spi.c
+ * @brief HAL spi implementation on STM32L4 Discovery Board
+ */
+
+/* ST Board includes. */
+#include "stm32l475xx.h"
+#include "stm32l4xx_ll_spi.h"
```
lundinc2: Same comment as above, I am curious if we can reduce the headers being included

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,677 @@
...
+    IRQn_Type eIrqNum;
+} STM32_SPI_HalContext_t;
+
+typedef struct IotSPIDescriptor
+{
+    STM32_SPI_HalContext_t const * pxSpiContext; /* ST Handle */
+    IotSPIMasterConfig_t xConfig;                /* Master Configuration */
+    IotSPICallback_t xSpiCallback;               /* Callback function */
+    void * pvUserContext;                        /* User context passed in callback */
+    uint8_t sOpened;                             /* Bit flags to tract different states. */
```
lundinc2: Small typo here: tract -> track

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
dcgaws: Yes, that would seem to be a safer approach. Also, I noticed that we'll return IOT_UART_SUCCESS even if the HAL_UART_DeInit call fails. We should return an error status in the latter case as well.

vendors/st/boards/stm32l475_discovery/aws_tests/config_files/test_iot_config.h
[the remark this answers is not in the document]
dcgaws: Ok thanks. This seems like a situation where the proposed file name is too generic and is very likely to result in namespace conflicts for FreeRTOS.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Let me try to remove them if possible.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: ok.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,744 @@
...
+                                                                               uint8_t * pData,
+                                                                               uint16_t Size,
+                                                                               uint32_t XferOptions ) );
+
+/*--------------------API Implementation---------------------*/
+
+IotI2CHandle_t iot_i2c_open( int32_t lI2CInstance )
+{
+    /* Exceeded maximum instance number. */
+    if( lI2CInstance >= _I2C_INSTANCE_NUM )
```
lundinc2: Should this also check if lI2CInstance < 0, i seem to remember this used to take an unsigned int, but should be updated now that it is signed

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,636 @@
...
+    .sOpened               = IOT_UART_CLOSED,
+};
+
+static IotUARTHandle_t const pxUarts[] = { &xUart0, &xUart1, &xUart2, &xUart3, &xUart4 };
+
+IotUARTHandle_t iot_uart_open( int32_t lUartInstance )
+{
+    IotUARTHandle_t xHandle = NULL;
+
+    if( ( lUartInstance < IOT_USART_PORTS ) && ( lUartInstance >= 0 ) )
```
lundinc2: The other ports seem to be using <= to the total ports, i think we should align all three. This is not mentioned in the header file, so it seems to be up to the implementation as to which one to use

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Good catch. I should check <0.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: What do you mean? instance id should less than total instances number.

lundinc2: i2c and spi use <=, so those should then be changed to < if that is the case

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
lundinc2: This is what I am referencing in the UART open comment, it seems your answer there indicates that this should then be <, not <=

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,676 @@
...
+
+static IotSPIHandle_t const pxSpis[] = { &xSpi1, &xSpi2, &xSpi3 };
+
+/*--------------------API Implementation---------------------*/
+
+IotSPIHandle_t iot_spi_open( int32_t lSpiInstance )
+{
+    IotSPIHandle_t xHandle = NULL;
+
+    if( ( lSpiInstance >= 0 ) && ( lSpiInstance <= SPI_INSTANCE_NUM ) )
```
lundinc2: This should also be <

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_uart.c
qiutongs: I see what you are saying now. Fixed them.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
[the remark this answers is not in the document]
dcgaws: Right, you could declare the array size implicitly, eliminate _I2C_INSTANCE_NUM, and then any code that loops through the array will be more future proof.

vendors/st/boards/stm32l475_discovery/aws_tests/application_code/st_code/stm32l4xx_hal_msp.c
```diff
@@ -1,91 +1,211 @@
...
+        /* USER CODE BEGIN UART4_MspInit 0 */
+
+        /* USER CODE END UART4_MspInit 0 */
+        /* UART4 clock enable */
+        __HAL_RCC_UART4_CLK_ENABLE();
+
+        __HAL_RCC_GPIOA_CLK_ENABLE();
+        __HAL_RCC_GPIOB_CLK_ENABLE();
+
+        /**UART4 GPIO Configuration
```
dcgaws: Are these pin mapping comment blocks all still correct? I'm not convinced that they're easier to understand than the code itself. Perhaps they could be replaced with comments that are purely conversational language.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
dcgaws: Why not fix it now?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
dcgaws: The code should be refactored then.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
dcgaws: Some other piece of code presumably has implementations of these routines. Wouldn't these types need to be visible there, too?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
dcgaws: Good catch! This suggests that we don't have test code for this, though. Do we?

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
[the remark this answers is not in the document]
qiutongs: ok, let me fix it.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: I can fix the "single return". But leave other MISRA violation later.

vendors/st/boards/stm32l475_discovery/aws_tests/application_code/st_code/stm32l4xx_hal_msp.c
qiutongs: TX and RX are correct but CTS is not. Let me remove those comments.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Any suggestion? Assert looks the best to me.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Not quite understand your question. One of the implementation is this https://github.com/aws/amazon-freertos/blob/master/vendors/st/stm32l475_discovery/STM32L4xx_HAL_Driver/stm32l4xx_hal_i2c.c#L753

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_i2c.c
qiutongs: Agree. I can say "Enable GPIO clock". Other than that, I don't more knowledge about this.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: Yeah, our test cases still need more coverage, especially ioctl. I have a backlog issue for it. As the scope of this problem is bigger, I wouldn't solve it in this PR.

vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
qiutongs: Ideally yes. SPI_HandleTypeDef structure is of ST driver's HAL while xSpiCallback is defined by us. I didn't find a good map of mapping SPI_HandleTypeDef pointer to IotSPIHandle_t
For now, let's keep it as is.

PR #1784, opened 2020-02-24
Fix async test by separating read and write semaphore

qiutongs:
Description
-----------

Fix async test by separating read and write semaphore.

Test case AFQP_IotUARTWriteReadAsyncWithCallback failed on ST board because it waits for write to complete before making a read call. That incurred UART overrun error. 

So the right way is to start async write and async read at the same time and verify each of them completes. I solve this problem by separating it to two semaphores.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_uart.c
+++ b/libraries/abstractions/common_io/test/test_iot_uart.c
@@ -63,7 +63,7 @@
 /* Globals values which can be overwritten by the test
  * framework invoking these tests */
 /*-----------------------------------------------------------*/
-uint8_t ustestIotUartPort = 0; /** Default UART port for testing */
+uint8_t uctestIotUartPort = 0; /** Default UART port for testing */
 uint32_t ultestIotUartFlowControl = 0;
 uint32_t ultestIotUartParity = 0;
 uint32_t ultestIotUartWordLength = 0;
@@ -182,7 +182,7 @@ TEST( TEST_IOT_UART, AFQP_AssistedIotUARTWriteReadSync )
     IotUARTHandle_t xUartHandle;
     int32_t lRead, lWrite, lClose;
 
-    xUartHandle = iot_uart_open( ustestIotUartPort );
+    xUartHandle = iot_uart_open( uctestIotUartPort );
     TEST_ASSERT_NOT_EQUAL( NULL, xUartHandle );
 
     if( TEST_PROTECT() )
@@ -211,7 +211,7 @@ TEST( TEST_IOT_UART, AFQP_IotUARTIoctlGetSet )
     int32_t lIoctl, lClose, lTransferAmount;
     IotUARTConfig_t xgetConfigBuffer, xConfigBuffer, xOriginalConfig;
 
-    xUartHandle = iot_uart_open( ustestIotUartPort );
+    xUartHandle = iot_uart_open( uctestIotUartPort );
     TEST_ASSERT_NOT_EQUAL( NULL, xUartHandle );
 
     if( TEST_PROTECT() )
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_uart.c
```diff
@@ -123,8 +131,10 @@ TEST_GROUP( TEST_IOT_UART );
  */
 TEST_SETUP( TEST_IOT_UART )
 {
-    xtestIotUARTSemaphore = xSemaphoreCreateCountingStatic( 10, 0, &xtestIotUARTCompleted );
-    TEST_ASSERT_NOT_EQUAL( NULL, xtestIotUARTSemaphore );
+    xReadCompleteSemaphore = xSemaphoreCreateBinaryStatic( &xReadSemaphoreBuffer );
+    TEST_ASSERT_NOT_EQUAL( NULL, xReadCompleteSemaphore );
```
dcgaws: Can the above call fail?

libraries/abstractions/common_io/test/test_iot_uart.c
```diff
@@ -102,11 +104,17 @@ static IotUARTConfig_t xSampleConfig2 =
 static void prvReadWriteCallback( IotUARTOperationStatus_t xOpStatus,
                                   void * pvParams )
 {
-    IotUARTOperationStatus_t * xCallbackStatus = ( IotUARTOperationStatus_t * ) pvParams;
     BaseType_t xHigherPriorityTaskWoken;
 
-    *xCallbackStatus = xOpStatus;
-    xSemaphoreGiveFromISR( xtestIotUARTSemaphore, &xHigherPriorityTaskWoken );
+    if( xOpStatus == eUartReadCompleted )
```
dcgaws: This branch suggests that it might be cleaner to refactor this into two separate callbacks.

However, for example, what if the read operation has ended, but it failed? Will the same xOpStatus value be returned in that case?

libraries/abstractions/common_io/test/test_iot_uart.c
```diff
@@ -471,30 +481,25 @@ TEST( TEST_IOT_UART, AFQP_IotUARTWriteReadAsyncWithCallback )
...
-        xCallbackReturn = xSemaphoreTake( ( SemaphoreHandle_t ) &xtestIotUARTCompleted, testIotUART_DEFAULT_SEMPAHORE_DELAY );
-        TEST_ASSERT_EQUAL( pdTRUE, xCallbackReturn );
-
-        TEST_ASSERT_EQUAL( eUartWriteCompleted, xCallbackStatus );
-
         lRead = iot_uart_read_async( xUartHandle, cpBufferRead, testIotUART_BUFFER_LENGTH );
         TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lRead );
 
-        xCallbackReturn = xSemaphoreTake( ( SemaphoreHandle_t ) &xtestIotUARTCompleted, testIotUART_DEFAULT_SEMPAHORE_DELAY );
+        xCallbackReturn = xSemaphoreTake( xWriteCompleteSemaphore, testIotUART_DEFAULT_SEMPAHORE_DELAY );
```
dcgaws: Should we wait on the write semaphore before scheduling the read? How do we expect a real customer app to do this?

libraries/abstractions/common_io/test/test_iot_uart.c
qiutongs: Good question. After check FreeRTOS's doc, this call should not fail. So we can remove the TEST_ASSERT_NOT_EQUAL.

https://www.freertos.org/xSemaphoreCreateBinaryStatic.html

libraries/abstractions/common_io/test/test_iot_uart.c
qiutongs: In current API design, there is only one callback registered for both read and write. So I think it is not easy to do 2 separate callbacks for async_read and async_write.

https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/include/iot_uart.h#L205

If the read failed, it will be different error code in which case the semaphore will not be given. Then the test case that waits for the semaphore will fail. So I think this flow should work well.

libraries/abstractions/common_io/test/test_iot_uart.c
qiutongs: No. That is the exact point I am changing in this PR. This test case is designed to be a wire-loopback one so the read should be scheduled immediately after write. Otherwise, data comes without being read. That caused UART overrun error.

For real customer app, it is unlikely to do loopback. That case is test by assisted tests (e.g. with Raspberry Pi) that not covered in this PR.

cmcgcc:
looks good to me

PR #1804, opened 2020-03-03
Added Doxygen For PKCS #11.

lundinc2:
This commit syncs the Doxygen code that was used to generate the FreeRTOS.org PKCS #11 Doxygen.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/doc/config/pkcs11
+++ b/doc/config/pkcs11
@@ -1,0 +1,35 @@
+# Include common configuration options.
+@INCLUDE_PATH = doc/config
+@INCLUDE = common
+
+# Basic project information.
+PROJECT_NAME = "PKCS11"
+PROJECT_BRIEF = "PKCS11 Cryptoki Library"
+
+# Library documentation output directory.
+HTML_OUTPUT = pkcs11
+
+# Generate Doxygen tag file for this library.
+GENERATE_TAGFILE = doc/output/pkcs11.tag
+ENABLE_PREPROCESSING = YES
+MACRO_EXPANSION = YES
+PREDEFINED="CK_DECLARE_FUNCTION(return, name)=return name"
+
+# Directories containing library source code.
+INPUT = doc \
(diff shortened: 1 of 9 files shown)


doc/config/pkcs11
```diff
@@ -0,0 +1,35 @@
...
+        doc/lib \
+        libraries/3rdparty/pkcs11 \
+        libraries/abstractions/pkcs11 \
+        libraries/freertos_plus/standard/pkcs11/include \
+        libraries/freertos_plus/standard/pkcs11/src \
+        libraries/abstractions/pkcs11/mbedtls \
+        libraries/abstractions/pkcs11/include \
+        libraries/freertos_plus/standard/utils/include/ \
+        libraries/freertos_plus/standard/utils/src/ \
+        vendors/pc/boards/windows/ports/pkcs11
```
lundinc2: Windows port is used here, this file will vary between ports but should be okay to use (and has historically) for the purposes of generating the dox

doc/config/pkcs11
```diff
@@ -0,0 +1,35 @@
...
+INPUT = doc \
+        doc/lib \
+        libraries/3rdparty/pkcs11 \
+        libraries/abstractions/pkcs11 \
+        libraries/freertos_plus/standard/pkcs11/include \
+        libraries/freertos_plus/standard/pkcs11/src \
+        libraries/abstractions/pkcs11/mbedtls \
+        libraries/abstractions/pkcs11/include \
+        libraries/freertos_plus/standard/utils/include/ \
+        libraries/freertos_plus/standard/utils/src/ \
```
qiutongs: A little abnormal to me when the source folder has to be included.

lundinc2: I will create a ticket to track migrating the doxygen stuff to only header files.

PR #1815, opened 2020-03-05
Implement UART, I2C on TI

qiutongs:
Description
-----------

Implement UART, I2C on TI

Checklist:
----------

- [ ] My code is Linted.


libraries/abstractions/common_io/test/test_iot_uart.c
```diff
@@ -541,7 +542,11 @@ TEST( TEST_IOT_UART, AFQP_IotUARTCancel )
 
         /* Wait to make sure operation was really canceled. */
         xCallbackReturn = xSemaphoreTake( xWriteCompleteSemaphore, testIotUART_DEFAULT_SEMPAHORE_DELAY );
-        TEST_ASSERT_EQUAL( pdFALSE, xCallbackReturn );
+
+        lCancel = iot_uart_ioctl( xUartHandle, eGetTxNoOfbytes, &lTransferAmount );
+        TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lCancel );
+
+        TEST_ASSERT_TRUE( pdFALSE == xCallbackReturn || lTransferAmount < testIotUART_BUFFER_LENGTH );
```
lundinc2: IMO: I think this is should be two separate asserts, it may be difficult to debug which case failed in this assert.

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,391 @@
...
+                       IotI2CIoctlRequest_t xI2CRequest,
+                       void * const pvBuffer )
+{
+    int32_t ret = IOT_I2C_SUCCESS;
+
+    switch( xI2CRequest )
+    {
+        case eI2CSetMasterConfig:
+
+            if( pvBuffer == NULL )
```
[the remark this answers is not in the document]
lundinc2: Could potentially reduce the LOC by checking for null buffer and peripheral before the switch statement

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,391 @@
...
+
+        pxI2CPeripheral->transaction.writeBuf = NULL;
+        pxI2CPeripheral->transaction.readBuf = NULL;
+    }
+    else
+    {
+        pxI2CPeripheral->flags &= ~_FLAG_NO_STOP;
+
+        /* If async, invoke user's callback. */
+        if( ( mode == _MODE_ASYNC ) && ( pxI2CPeripheral->userCallback != NULL ) )
```
lundinc2: Is the expectation that the user callback is invoked when the i2c transaction is completed ?

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_i2c.c
```diff
@@ -0,0 +1,391 @@
...
+        {
+            pxI2CPeripheral->userCallback( eI2CCompleted, pxI2CPeripheral->userContext );
+        }
+    }
+
+    return ret;
+}
+
+static void _callbackInternal( I2C_Handle handle,
```
lundinc2: Handle and transaction are unused here, can those parameters just be removed?

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,478 @@
...
+    if( _isIdle( uartHandle->tiUartHandle ) )
+    {
+        return IOT_UART_NOTHING_TO_CANCEL;
+    }
+
+    UART_writeCancel( uartHandle->tiUartHandle );
+    UART_readCancel( uartHandle->tiUartHandle );
+
+    /* Have to sleep for a short period to let cancel finish. */
+    vTaskDelay( pdMS_TO_TICKS( 1000 ) );
```
lundinc2: I think it would be better to use a semaphore here. Waiting 1 second is pretty long. Reading the TI API, when you call cancel it invokes the registered callback (whether it be read or write). You could probably add some magic number or something to the callback, signaling you to release the semaphore in the callback.

lundinc2: Maybe setting some variable in the uartHandle

osloapps-max:
Chiming in here as I know this was an ongoing pull process.

 I'm currently working on full ports for the TI SimpleLink family. This include the CC32XX, MSP432, CC13X2 and CC26X2 family. My work would also cover the work being done here so it might be interesting to consider if  these ports are still interesting or not (considering there most likely will be duplicates).

Most of our (TI) drivers are are agnostic to the platform which means that as long as the ports are setup without using the device specific header files, the port will fit most devices. I know there is a few ports that will need to be done specific (as per API requirements) but I believe we can cover more than 80% of the common-io APIs with generic ports.

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_i2c.c
gkwicker: Question – are callbacks only for asynchronous operation?

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_uart.c
gkwicker: +1, also hard-coded delays in common code like this are usually a mistake.  If you can't have a structure member as @lundinc2 suggests, then at least use a `#define`.

vendors/ti/boards/cc3220_launchpad/ports/common_io/iot_uart.c
```diff
@@ -0,0 +1,478 @@
...
+               unsigned long baud = 0;
+               unsigned long tiConfig = 0;
+               unsigned long tiFlowMode = 0;
+
+               tiFlowMode = MAP_UARTFlowControlGet( hwAttrs->baseAddr );
+               MAP_UARTConfigGetExpClk( hwAttrs->baseAddr, freq.lo, &baud, &tiConfig );
+
+               pConfig->ucFlowControl = tiFlowMode == UART_FLOWCONTROL_NONE ? 0 : 1;
+               pConfig->ulBaudrate = baud;
+               pConfig->ucWordlength = ( uint8_t ) _indexOf( dataLength, sizeof( dataLength ), ( tiConfig & UART_CONFIG_WLEN_MASK ) ) + 5;
```
gkwicker, line 385: Note that `_indexOf()` returns `int8_t` and can return `-1`.  If that happens, `pConfig->ucWordLength` will be set to `0x4` here (cast to `uint8_t` gives `0xff`, then + 5).  Is that what we want?

qiutongs:
This PR is not useful anymore. TI is working on the implementation.

PR #1823, opened 2020-03-09
[BLE] Add set device name API

ravibhagavandas:
Add set device name API 

Description
-----------

Add a new API to set device name for BLE at run time. This facilitates reading the device name from a storage example flash and configuring different names for different devices without changing code.

Minor - Replaced some static variables (used only once) to functions to save RAM.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
+++ b/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
@@ -130,7 +130,7 @@ static const BTIOtypes_t xIO = IOT_BLE_INPUT_OUTPUT;
 static const bool bIsBondable = IOT_BLE_ENABLE_BONDING;
 static const bool bSecureConnection = IOT_BLE_ENABLE_SECURE_CONNECTION;
 
-static char bleDeviceName[ IOT_BLE_MAX_DEVICE_NAME_LENGTH + 1 ] = { 0 };
+static char bleDeviceName[ IOT_BLE_DEVICE_LOCAL_NAME_MAX_LENGTH + 1 ] = { 0 };
 
 static bool isBLEOn = false;
 
@@ -917,7 +917,7 @@ BTStatus_t IotBle_SetDeviceName( const char * pName,
 {
     BTStatus_t status;
 
-    if( length <= IOT_BLE_MAX_DEVICE_NAME_LENGTH )
+    if( length <= IOT_BLE_DEVICE_LOCAL_NAME_MAX_LENGTH )
     {
         /* Copy the device name to a buffer, so that device name does not change between
          * stack disable-enable cycles.
--- a/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
+++ b/libraries/c_sdk/standard/ble/src/iot_ble_gap.c
@@ -125,39 +125,14 @@ static IotBleAdvertisementParams_t _advParams =
 };
 
 
-const uint32_t usMtu = IOT_BLE_PREFERRED_MTU_SIZE;
-const BTIOtypes_t xIO = IOT_BLE_INPUT_OUTPUT;
-const bool bIsBondable = IOT_BLE_ENABLE_BONDING;
-const bool bSecureConnection = IOT_BLE_ENABLE_SECURE_CONNECTION;
+static const uint32_t usMtu = IOT_BLE_PREFERRED_MTU_SIZE;
+static const BTIOtypes_t xIO = IOT_BLE_INPUT_OUTPUT;
+static const bool bIsBondable = IOT_BLE_ENABLE_BONDING;
+static const bool bSecureConnection = IOT_BLE_ENABLE_SECURE_CONNECTION;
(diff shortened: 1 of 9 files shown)


libraries/c_sdk/standard/ble/src/iot_ble_gap.c
```diff
@@ -876,9 +900,63 @@ BTStatus_t IotBle_Init( void )
...
+        status = eBTStatusSuccess;
+    }
+    else
+    {
+        status = eBTStatusNoMem;
+    }
+
+    if( status == eBTStatusSuccess )
+    {
+        if( isBLEOn )
```
dachalco: In the header iot_ble.h:649 it states:

> API can be used to set the device name before turning on BLE by calling IotBle_On(), or when BLE is running. When BLE is advertising mode...

Does that mean the name can be adjusted while BLE is off? If so, I'm confused how this call could modify the name while BLE is off, given this call won't modify the name unless `isBLEOn == true`

libraries/c_sdk/standard/ble/src/iot_ble_gap.c
```diff
@@ -876,9 +900,63 @@ BTStatus_t IotBle_Init( void )
...
+{
+    BTStatus_t status;
+
+    if( length <= IOT_BLE_MAX_DEVICE_NAME_LENGTH )
+    {
+        /* Copy the device name to a buffer, so that device name does not change between
+         * stack disable-enable cycles.
+         */
+        strncpy( bleDeviceName, pName, length );
+        bleDeviceName[ length + 1 ] = '\0';
```
dachalco: ***This one is important:
```suggestion
        bleDeviceName[ length ] = '\0';
```
Given length is one-indexed, but the indexing is zero-indexed, this could go out of bounds.

ravibhagavandas: Buffer size is IOT_BLE_MAX_DEVICE_NAME_LENGTH + 1  (including null terminator) so it will not go out of bounds.

dachalco: Right, but say `length = IOT_BLE_MAX_DEVICE_NAME_LENGTH` and for sake of example max equals `4`. So then there's `IOT_BLE_MAX_DEVICE_NAME_LENGTH + 1 = 5` items alloced. This line could try to index 
`i = length + 1` --> `i = 4 + 1` -- > `i = 5`.  Then we're trying to index `array[5]` but that array only has 5 items i.e. max index of `array[4]`

libraries/c_sdk/standard/ble/src/iot_ble_gap.c
ravibhagavandas: API copies the device name to a buffer, so when BLE is turned on ( IotBle_On() ) it sets the stack property with the updated value.

libraries/c_sdk/standard/ble/src/iot_ble_gap.c
ravibhagavandas: That was updated to ` bleDeviceName[ length ]` in one of the commits in PR. Could you check the latest version of PR.

libraries/c_sdk/standard/ble/src/iot_ble_gap.c
dachalco: Ah, I see that now in IotBle_On. Got it

libraries/c_sdk/standard/ble/src/iot_ble_gap.c
dachalco: I see it! Awesome

libraries/c_sdk/standard/ble/test/iot_test_ble_end_to_end.c
```diff
@@ -107,13 +250,29 @@ TEST_GROUP_RUNNER( Full_BLE_END_TO_END_MQTT )
...
+
+TEST( Full_BLE_END_TO_END_CONNECTIVITY, EnableBLE )
+{
+    _BLEEnable();
+}
+
+TEST( Full_BLE_END_TO_END_CONNECTIVITY, SetDeviceName )
+{
+    TEST_ASSERT_EQUAL( eBTStatusSuccess,
+                       IotBle_SetDeviceName( IOT_BLE_DEVICE_COMPLETE_LOCAL_NAME, strlen( IOT_BLE_DEVICE_COMPLETE_LOCAL_NAME ) ) );
```
gkwicker: This test just checks whether or not `IotBle_SetDeviceName()` returns `eBTStatusSuccess`.  How do we know it actually did anything?

ravibhagavandas: This a system level test. The BLE peer will scan and connect to devices using the name set via this API and so the next test BLE_Enable() succeeds only if the device name is set correctly in the advertisement.

PR #1825, opened 2020-03-10
Logging function use cleaned up

AniruddhaKanhere:
Logging function use cleaned up

Description
-----------
The logging function was being guarded by multiple #ifs all over the code. Instead of doing that at every single place, a #define was used to guard the declaration and definition itself. It makes the code look cleaner and might help future users avoid mistakes.

Code tests:
- Custom Cmake (#define ipconfigHAS_PRINTF = 1). The tests show that shadow did not pass, but since that doesn't have anything to do with this change I have assumed that the code works. (Also other tests passed)
- Custom Cmake (#define ipconfigHAS_PRINTF = 0). The tests show that OTA did not pass, but since that doesn't have anything to do with this change I have assumed that the code works. (Also other tests passed)

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/esp32/NetworkInterface.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/esp32/NetworkInterface.c
@@ -34,7 +34,7 @@
 #include "esp_log.h"
 #include "esp_wifi.h"
 #include "esp_wifi_internal.h"
-prvMonitorResources();#include "tcpip_adapter.h"
+#include "tcpip_adapter.h"
 
 enum if_state_t {
     INTERFACE_DOWN = 0,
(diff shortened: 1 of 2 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/esp32/NetworkInterface.c
```diff
@@ -44,8 +44,13 @@ enum if_state_t {
...
 volatile static uint32_t xInterfaceState = INTERFACE_DOWN;
 
+/* protect the function declaration itself instead of using
+   #if everywhere.                                        */
 #if ( ipconfigHAS_PRINTF != 0 )
     static void prvMonitorResources();
+    #define PRINT_RESOURCE_LOGS prvMonitorResources();
+#else
+    #define PRINT_RESOURCE_LOGS
 #endif
```
cobusve: How about instead of this you leave the code as just `prvMonitorResources(); ` where it is being called, and here you do something like this.

```
#if ( ipconfigHAS_PRINTF != 0 )
    static void prvMonitorResources();	    static void prvMonitorResources();
#else
    #define prvMonitorResources()
#endif
```	

That way you can also remove the implementation of the function itself with the same guard on ipconfigHAS_PRINTF

AniruddhaKanhere: Done

PR #1838, opened 2020-03-16
Update logs to stringify error codes from mbedTLS

aggarw13:
Improve error logging for mbedTLS failures

Description
-----------
Existing error logs containing mbedTLS numeric error codes are not readable. This PR replaces those with strings by using the `mbedtls_strerror` utility

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -47,6 +47,7 @@
 #include "mbedtls/sha256.h"
 #include "mbedtls/base64.h"
 #include "threading_alt.h"
+#include "mbedtls/error.h"
 
 /* Credential includes. */
 #include "aws_clientcredential.h"
@@ -71,9 +72,9 @@ typedef int ( * pfnMbedTlsSign )( void * ctx,
                                   size_t hash_len,
                                   unsigned char * sig,
                                   size_t * sig_len,
-                                  int ( * f_rng )( void *,
-                                                   unsigned char *,
-                                                   size_t ),
+                                  int ( *f_rng )( void *,
+                                                  unsigned char *,
+                                                  size_t ),
                                   void * p_rng );
 
 /**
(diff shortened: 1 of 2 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3409,6 +3410,7 @@ CK_DECLARE_FUNCTION( CK_RV, C_Sign )( CK_SESSION_HANDLE xSession,
     CK_BBOOL xSignatureGenerated = CK_FALSE;
     uint8_t ecSignature[ pkcs11ECDSA_P256_SIGNATURE_LENGTH + 15 ]; /*TODO: Figure out this length. */
     int lMbedTLSResult;
+    char cMbedTlsErrorStr[ 134 ];
```
gkwicker: This increases the stack size quite a bit, I don't think we want to do this.

lundinc2: To add to this, this seems like a magic number, I think we should use some kind of macro to define the size, were does 134 come from?

lundinc2: @gkwicker perhaps we could use a global buffer, and protect it with a mutex? Or allocate a buffer on the heap in the case we do receive an error.

aggarw13: Yes, good suggestion to use a macro to avoid magic numbers. I came up with 134 as size of the largest error message in [error.c](https://github.com/aws/amazon-freertos/blob/master/libraries/3rdparty/mbedtls/library/error.c#L409)(hyperlinked to the longest error string) + 1 (for the NULL character)

aggarw13: I thought about using file-specific globals to avoid allocation the buffer on the stack, but wasn't sure whether that would require shared memory protection for thread-safety (in cases of multiple tls connections being alive on the platform)

gkwicker:
Closing, we will find a way to upstream a change to mbedtls for `const char*` log messages from error codes.

PR #1841, opened 2020-03-18
Add MQTT mock files for testing

gordonwang0:
Description
-----------
Imports the files needed to run the MQTT mocks from the C SDK.

Checklist:
----------

- [x] My code is Linted.


libraries/c_sdk/standard/mqtt/CMakeLists.txt
```diff
@@ -72,7 +73,7 @@ endif()
 afr_module_sources(
     ${AFR_CURRENT_MODULE}
     INTERFACE
-
+        "${test_dir}/mock/iot_tests_mqtt_mock.c"
```
aggarw13: Why include the mock within the mqtt library output?  Shouldn't the MQTT mock be defined as a separate library target?

gordonwang0: It isn't currently used, but will be when we change some of the integration tests.  This is in preparation for that.

The CMakeLists in this repo build the single aws_tests executable.  So if included here, it will be linkable for other library tests.

aggarw13: My suggestion is that we make an explicit target for the MQTT mock instead of including that in the MQTT tests' target, as the MQTT mock does not belong to the MQTT tests.
Any other library should be linking against the MQTT mock library target instead of the MQTT tests' target.

gordonwang0: The CMakeLists here are designed so that common test files are built into the aws_tests target, not as separate modules.

PR #1843, opened 2020-03-19
Rename serializer tests

muneebahmed10:
Description
-----------

Rename serializer unit tests

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/common/aws_test_runner.c
+++ b/tests/common/aws_test_runner.c
@@ -210,9 +210,9 @@ static void RunTests( void )
     #endif
 
     #if ( testrunnerFULL_SERIALIZER_ENABLED == 1 )
-        RUN_TEST_GROUP( Full_Serializer_CBOR );
-        RUN_TEST_GROUP( Full_Serializer_JSON );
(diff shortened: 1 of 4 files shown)


muneebahmed10:
> As these are all unit tests, should all files be renamed (as well) to contain the `_unit_` word?

I didn't see much point to this since these are all of the serializer tests, so there aren't any system tests to distinguish them from.

tests/common/aws_test_runner.c
```diff
@@ -210,9 +210,9 @@ static void RunTests( void )
     #endif
 
     #if ( testrunnerFULL_SERIALIZER_ENABLED == 1 )
```
aggarw13: Should this macro be changed to `testrunnerSERIALIZER_UNIT_ENABLED` to give the right picture about these tests in the reports generate/analysis done from the CI test results?

muneebahmed10: I don't think so. All of the other macros in this file are either of the form `testrunnerFULL...` or `testrunnerUTIL_PLATFORM...`, even when running unit tests

PR #1847, opened 2020-03-20
Change Default Winsim Test Port to 0

lundinc2:
Currently the winsim test project is configured to network adapter 3. If this adapter is incorrect for your windows PC, the project will seemingly
get stuck after successfully opening the network port. This issue is very hard to debug, as the project appears to just be frozen. If we configure this
network port to 0, the test project will be unable to open the network adapter, and print a helpful error message that helps the user configure it to the right
network adapter.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/pc/boards/windows/aws_tests/config_files/FreeRTOSConfig.h
+++ b/vendors/pc/boards/windows/aws_tests/config_files/FreeRTOSConfig.h
@@ -171,7 +171,7 @@ extern void vLoggingPrint( const char * pcMessage );
  * results in the wired network being used, while setting
  * configNETWORK_INTERFACE_TO_USE to 2 results in the wireless network being
  * used. */
-#define configNETWORK_INTERFACE_TO_USE       ( 3L )
+#define configNETWORK_INTERFACE_TO_USE       ( 0L )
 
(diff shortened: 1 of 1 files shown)


htibosch:
@lundinc2 : You're so quick, I even didn't have a chance to see the PR before it got merged.

This a very good idea, many people won't realise that the wrong network adapter has been pre-selected.

In addition to this, I'd like to see a change in the driver. It should only show a list of available adapters in case `configNETWORK_INTERFACE_TO_USE` equals zero, which is an error condition.
When an adapter has been selected, it is enough to show the name of that adapter only.
Note; the original version of FreeRTOS+TCP is now stored in [github.com/freertos/freertos](https://github.com/FreeRTOS/FreeRTOS/tree/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/WinPCap)

One step further would be not to define a number but a string: take the adapter which has "Realtek" in its name. That would overcome a second problem: the presence of some adapters changes over time.

lundinc2:
@htibosch thank you for this suggestion! I will take some time to raise this issue, and see about getting this implemented.

PR #1862, opened 2020-03-28
TEST: fix race condition in Unit test

alfred2g:
Fix Race Condition In Unit Test

Description
-----------
A race condition could occur in UT where a thread frees a shared variable, then it is used by the main thread which could cause a core dump

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/tests/unit_test/linux/CMakeLists.txt
+++ b/tests/unit_test/linux/CMakeLists.txt
@@ -46,8 +46,8 @@
                 )
     set_target_properties(utils PROPERTIES
             LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
-            COMPILE_FLAGS "-Wall -Og -pthread"
-            LINK_FLAGS "-Wall -Og -pthread -lpthread"
+            COMPILE_FLAGS "-Wall -Og -pthread -ggdb3"
+            LINK_FLAGS "-Wall -Og -pthread -lpthread -ggdb3"
         )
 
     # add unit test subdirectories here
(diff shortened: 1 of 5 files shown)


tests/unit_test/linux/CMakeLists.txt
```diff
@@ -46,8 +46,8 @@
                 )
     set_target_properties(utils PROPERTIES
             LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
-            COMPILE_FLAGS "-Wall -Og -pthread"
-            LINK_FLAGS "-Wall -Og -pthread -lpthread"
+            COMPILE_FLAGS "-Wall -Og -pthread -ggdb3"
```
gkwicker, line 49: Why are we enabling debugging (`-ggdb3`) without using `CMAKE_CFLAGS_DEBUG`?  Ideally we should be executing unit tests both with and without debug flags and using [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html) to control the build type.

alfred2g: when would you not want debugging information with the UT?

gkwicker: Ideally you will run your tests with and without debugging.  But the main issue here is that we're adding debugging support in a non-cmake standard way.  cmake gives you a standard method of doing this with `CMAKE_CFLAGS_DEBUG` and `CMAKE_BUILD_TYPE`, but we're not using that here.

alfred2g: Ah I see what you mean, good point, pushed another commit, thanks!

gkwicker: nice!

Issue #1859, opened 2020-03-26
common-io test cases will have cascading failures as resources are not freed up


osloapps-max:
**Describe the bug**
For most of the common-io test cases, the resources are not freed up following a failed test case. This will cause the following tests to fail as the resource would still be considered "in-use".

**System information**
- Which hardware board or part numbers? Texas Instruments SimpleLink family
- IDE used: none
- Operating System: Windows
- Version of FreeRTOS (run ``git describe --tags`` to find it): 201912.00-151-ga6a459b5e
- Project: common-io porting

**Expected behavior**
Using the test framework "Setup" and "Tear-down" stages to allocate and free the resource before and after a test case. Alternatively, use TEST_PROTECT() sections to make sure resources are freed up as expected. 

**To reproduce**

Run generic test group (for example ADC) and force an early error.

**Additional context**

I am currently working in ports for the Texas Instruments Simplelink family (CC32XX, MSP432, CC13X2 and CC26X2). I'm also available to do/provide work on implementing these fixes (moving to using the setup/ tear-down stages) as I'm fixing these issues anyway to be able to test the ports properly. 

Test cases for which pull-requests has been opened and the current status:
ADC - Merged (https://github.com/aws/amazon-freertos/pull/1880)
UART - Merged (https://github.com/aws/amazon-freertos/pull/1878)
SPI - Merged (https://github.com/aws/amazon-freertos/pull/1879)
GPIO - Merged (https://github.com/aws/amazon-freertos/pull/1882)
FLASH - Merged (https://github.com/aws/amazon-freertos/pull/1883)
Timer - Merged (https://github.com/aws/amazon-freertos/pull/1887)
Watchdog - Merged (https://github.com/aws/amazon-freertos/pull/1888)
I2C - Merged (https://github.com/aws/amazon-freertos/pull/1905)
PWM - Merged (https://github.com/aws/amazon-freertos/pull/1908)
tsensor - Review (https://github.com/aws/amazon-freertos/pull/1932)
RTC - Review (https://github.com/aws/amazon-freertos/pull/1977)

lundinc2:
@osloapps-max thank you for raising this issue and offer to help. @qiutongs and I are available should you need any help / assistance.

osloapps-max:
[...]

PR #1871, opened 2020-04-02
Common-IO UART Test Fixes

lundinc2:
Move UART async read to occur before call to async write in loopback test, for devices that don't buffer a write.
Add logic to UART cancel test for devices that do not support cancel.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.


osloapps-max:
@lundinc2 

I've done the same changes as part of my work to refactor the test cases based on the following issue:
https://github.com/aws/amazon-freertos/issues/1859

I have attached my version of the UART tests, maybe this would be interesting for you for this pull request?
[test_iot_uart.zip](https://github.com/aws/amazon-freertos/files/4421995/test_iot_uart.zip)

lundinc2:
@osloapps-max, 

Thanks! I took a look at those changes, they look great. Do you think you could open a pull request either to my fork's branch, or the master branch in the main repository so we can get those in? 

Preferably to the master branch, but which ever is easiest for you.

osloapps-max:
@lundinc2 
I'm currently doing similar work for all of the ports I'm doing for the TI SimpleLink family, I expect to have gone over all possible ports in about two weeks (as next week is easter holiday where I live).

If you are OK with waiting for this, I could provide a pull request with similar updates for most of the tests in there today (a few excluded like USB etc as I only update test cases I can actually run). 

If not (and you want them faster), I could try to get an pull request up with what I got today early next week :)

lundinc2:
@osloapps-max whatever is easiest for you. I do prefer smaller and frequent PRs, as they will be easier for me to review.

PR #1878, opened 2020-04-06
common-io: Refactoring and updates to UART test cases

osloapps-max:
Description
-----------
Refactored open/close to be done in setup/tear-down stages. 

Updated word length defines to number of bits.

Added checks for optional features such as flow-control support. 

Refactored loopback tests to perform read->write instead of write->read.

This is part of the work related to the following issue:

https://github.com/aws/amazon-freertos/issues/1859

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_uart.c
+++ b/libraries/abstractions/common_io/test/test_iot_uart.c
@@ -234,7 +234,7 @@ TEST( TEST_IOT_UART, AFQP_IotUARTIoctlGetSet )
         TEST_ASSERT_EQUAL( xSampleConfig2.ulBaudrate, xgetConfigBuffer.ulBaudrate );
 
         /* Test changing control flow config if specified by the test framework */
-        if (ultestIotUartFlowControl)
+        if( ultestIotUartFlowControl )
         {
             xConfigBuffer.ucFlowControl = xSampleConfig2.ucFlowControl;
             lIoctl = iot_uart_ioctl( xUartHandle, eUartSetConfig, &xConfigBuffer );
@@ -795,7 +795,7 @@ TEST( TEST_IOT_UART, AFQP_IotUARTOpenCloseCancelFuzzing )
         lCancel = iot_uart_cancel( NULL );
         TEST_ASSERT_EQUAL( IOT_UART_INVALID_VALUE, lCancel );
     }
-    
+
     lClose = iot_uart_close( xUartHandle_2 );
     TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lClose );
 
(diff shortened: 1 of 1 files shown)


osloapps-max:
@lundinc2 Pull request for my UART test refactoring discussed in https://github.com/aws/amazon-freertos/pull/1871

osloapps-max:
@qiutongs I reverted the open/close refactoring in favor of test_protect(), I was a bit quick on this one. I kept the "flow-control guard", moved some code into the "protected" section and re-ordered the async loopback read/write order.

libraries/abstractions/common_io/test/test_iot_uart.c
```diff
@@ -390,16 +393,16 @@ TEST( TEST_IOT_UART, AFQP_IotUARTWriteAsyncReadAsyncLoopbackTest )
...
+        TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lRead );
+
         lWrite = iot_uart_write_async( xUartHandle, cpBuffer, testIotUART_BUFFER_LENGTH );
         TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lWrite );
 
         lIoctl = iot_uart_ioctl( xUartHandle, eGetTxNoOfbytes, &lTransferAmount );
         TEST_ASSERT_EQUAL( IOT_UART_SUCCESS, lIoctl );
         TEST_ASSERT_GREATER_THAN( 0, lTransferAmount );
 
-        lRead = iot_uart_read_async( xUartHandle, cpBufferRead, testIotUART_BUFFER_LENGTH );
```
qiutongs, line 400: I see your point here. But I suppose it should work with write-then-read in an async way, even there is no hardware buffer for RX, assuming the full-duplex being supported. 

Because write_async should return and call read_async immediately. I might be wrong and I will confirm with team. Please also share your thoughts.

qiutongs:
> @qiutongs I reverted the open/close refactoring in favor of test_protect(), I was a bit quick on this one. I kept the "flow-control guard", moved some code into the "protected" section and re-ordered the async loopback read/write order.

Thanks for doing that. My only comment is around the async loopback read/write order.

libraries/abstractions/common_io/test/test_iot_uart.c
osloapps-max: I can tell you from experience that write -> read can not be expected to succeed. No matter the closeness of the calls, the write will in most systems start without a delay and as there will always be processing time to get into the read call and set it up, a success is not guaranteed (er even likely).

This is of course very device specific, depending on they underlying HW and driver implementations. All devices should however be able to support async read -> write.

qiutongs: What you said makes sense to me.

PR #1880, opened 2020-04-06
common-io: Refactoring ADC test cases

osloapps-max:
Description
-----------
Refactored tests to perform open/close during setup/tear-down phases.

This is part of the work related to the following issue: https://github.com/aws/amazon-freertos/issues/1859

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_adc.c
+++ b/libraries/abstractions/common_io/test/test_iot_adc.c
@@ -49,7 +49,10 @@ typedef struct IotAdcTestUserContext_s
 } testIotAdcTestUserContext_t;
 
 #define testIotADC_CH_DATA_BUF_LEN    5
+
+/* static global variables */
 static SemaphoreHandle_t xtestIotAdcTestSemaphore = NULL;
+static IotAdcHandle_t xAdcHandle;
 
 /* global variables for channel test control */
 /* ADC module instance, default to 0 */
@@ -60,9 +63,10 @@ uint8_t uctestIotAdcChListLen = 1;
 uint8_t * puctestIotAdcChList = NULL;
 /* ADC module instance for assisted test, need to be configured to the channel with external access */
 uint8_t ucAssistedTestIotAdcChannel = 0;
-
 /* ADC Channel Scan Wait time in ms */
 uint32_t ltestIotAdcTestChWaitTime = 1000;
+
+
 
 /* Define Test Group. */
 TEST_GROUP( TEST_IOT_ADC );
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_adc.c
```diff
@@ -82,6 +89,17 @@ TEST_SETUP( TEST_IOT_ADC )
...
 TEST_TEAR_DOWN( TEST_IOT_ADC )
 {
+    /* Release the ADC hardware during teardown */
+    int32_t lRetVal;
+
+    lRetVal = iot_adc_close( xAdcHandle );
+
+    if (lRetVal == IOT_ADC_NOT_OPEN)
+    {
+        return;
```
lundinc2, line 99: I prefer removing this return, and just guarding the assert with an if statement, as this is against the coding standard.

libraries/abstractions/common_io/test/test_iot_adc.c
```diff
@@ -132,14 +150,7 @@ static void prvAdcChCallback( uint16_t * pusConvertedData,
...
 {
-    IotAdcHandle_t xAdcHandle;
-    int32_t lRetVal;
-
-    xAdcHandle = iot_adc_open( uctestIotAdcInstance );
-    TEST_ASSERT_NOT_EQUAL( NULL, xAdcHandle );
-
-    lRetVal = iot_adc_close( xAdcHandle );
-    TEST_ASSERT_EQUAL( IOT_ADC_SUCCESS, lRetVal );
+    /* No content, test is covered by setup / teardown */
```
lundinc2, line 153: We can just delete this test case if so

libraries/abstractions/common_io/test/test_iot_adc.c
qiutongs: I would say we can simply just close without checking the return status:

```c
/* Intentionally not check the return status. */
iot_adc_close( xAdcHandle );
```

This is an example why I like using "Test_Protect()" to solving such cascading failures. Then the open/close logic can be contained in each test case clearly.

osloapps-max: I think I will do a quick re-spin of the test cases to go with the "test_protect()" solution that you prefer so that the look and feel remain the same.

qiutongs: Really appreciate that.

qiutongs:
One last thing is to format the code. Our team uses the [uncrustify tool](http://uncrustify.sourceforge.net/)

Please note to use this config: https://github.com/aws/amazon-freertos/blob/master/tools/uncrustify.cfg

PR #1882, opened 2020-04-07
common-io: Refactoring GPIO test cases

osloapps-max:
Description
-----------
* Refactor tests to use TEST_PROTECT sections. 
* Added checks for some functions that could be unsupported by the HW. 
* Added use of pull-up in read-back test for open-drain output
* Removed setup-phase code as setting GPIO functionality might not be supported as per API description. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_gpio.c
+++ b/libraries/abstractions/common_io/test/test_iot_gpio.c
@@ -148,7 +148,7 @@ static void testIotGpioOutputMode( IotGpioOutputMode_t eGpioOutputMode,
     IotGpioPull_t xGpioPullA = eGpioPullNone;
 
     /* When testing open-drain, use pull-up on port A (as the output can not driven high) */
-    if (eGpioOutputMode == eGpioOpenDrain)
+    if( eGpioOutputMode == eGpioOpenDrain )
     {
         xGpioPullA = eGpioPullUp;
     }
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_gpio.c
```diff
@@ -135,6 +106,9 @@ TEST_SETUP( TEST_IOT_GPIO )
  */
 TEST_TEAR_DOWN( TEST_IOT_GPIO )
 {
+    /* Make sure resources are freed up */
+    iot_gpio_close( xtestIotGpioHandleA );
```
qiutongs, line 110: Much cleaner, thanks.

libraries/abstractions/common_io/test/test_iot_gpio.c
```diff
@@ -148,29 +122,17 @@ TEST_GROUP_RUNNER( TEST_IOT_GPIO )
     TEST_ASSERT_NOT_EQUAL( NULL, xtestIotGpioSemaphore );
 
     RUN_TEST_CASE( TEST_IOT_GPIO, AFQP_IotGpioOpenClose );
-    testIotGpioClosePorts();
```
hydedong, line 151: reason we had this originally is to make sure ports are closed if previous test exit early because of assertion and not closing port. if we remove this, how do we make sure early test will not affect later ones?

hydedong: ok, I see each test case has setup and teardown, and we close port in tear down phase, is this true for every RUN_TEST_CASE? changes look ok if that's the case

qiutongs: Yes, that is true. That is how setup/teardown are supposed to be used.

Issue #1766, opened 2020-02-17
Sample/Demo code for WiFi SoftAP mode


teuteuguy:
Looking at the [documentation](https://docs.aws.amazon.com/freertos/latest/userguide/freertos-wifi.html) for the WiFi library there is the SoftAP mode.
Could you provide sample/demo code that shows how to setup wifi credentials on a device using this mode?

ie. Have mobile or laptop connect to the device SoftAP, show a page, enter the SSID and Password and reboot with the new credentials.

Thank you!

dnoal:
Up to recently, our support for wifi provisioning is based upon BLE.  SoftAP mode is supported in our wifi abstraction, but the complete functionality to support WiFi provisioning using that abstraction has not been supported.  We have started to make progress here.  On ESP32 specifically, Espressif has recently contributed two PRs (1519 and 1571) that are now merged into master.  Using those two PRs (read the details on configuration), it should be possible to use the Espressif documented method of configuring the wifi module to SoftAP mode, serving an IP address (via DHCP) to a mobile client, serving a simple web page to that client, entering the WiFI credentials from that client, and doing the tap dance to restore the wifi module to client/ station mode and using the entered WiFi credentials.  That approach is in the ESP32 documentation.  We are working on documenting that approach and providing a code sample, but you should be enabled now.  We are working to make WiFi based provisioning uniform across our supported platforms but that is not available at this time.  Note:  Some wifi modules (Microchip) act differently than the ESP32 Wifi, and can support wifi provisioning now.  Please advise on your success.

pisukeman:
Hi @dnoal ! thanks for your response. We are currently developing a FreeRTOS firmware on a ESP32 device.
Our customer wants a wifi provisioning system using a wifi SoftAP, because the wifi provisioning via BLE is using too much memory on the device because we also need to make some HTTP call (BLE+HTTP is consuming too much memory on the ESP32)

As i can see in your latest message you are working on it. Is this possible to have some early version of your work in order to make the wifi based provisioning? or some workaround that should do the trik?
If this is not possible, how can we manage to do the provisioning using BLE and also make an HTTP call to download certificates as it's explained in this aws doc? (https://aws.amazon.com/es/blogs/iot/ensure-secure-communication-with-aws-iot-core-using-the-certificate-vending-machine-reference-application/). 
[...]

PR #1886, opened 2020-04-08
Use LwIP as default networking stack on ESP32

shubhamkulkarni97:
Description
-----------
There are many advantages of using LwIP stack on ESP32.
* LwIP stack has lower memory consumption.
* LwIP supports zero copy receive path.
* Most of the features like WiFi provisioning require LwIP stack.

Considering these advantages, it will be useful to change default networking stack to LwIP.

Checklist:
----------

- [x] My code is Linted.


htibosch:
> LwIP stack has lower memory consumption.

Around December last year, I received an email saying that the people at ESP32 labs had found that using lwIP would save **15 KB** of RAM compared to FreeRTOS+TCP. Because of that, I studied FreeRTOS+TCP's memory consumption for the ESP32 in detail.

The FreeRTPS+TCP memory footprint is configurable. Please read [this post](https://github.com/aws/amazon-freertos/issues/1250#issuecomment-586172125) which explains how easy it is to save **30 KB** of RAM without affecting the performance.

People from the ESP32 labs have confirmed that the memory problems are solved now.

You may also have a look at [PR #44](https://github.com/FreeRTOS/FreeRTOS/pull/44) in which the `tcp_mem_stats` tool is presented. Using the tool results in a spreadsheet showing both static and dynamic memory usage.


> LwIP supports zero copy receive path.

FreeRTOS+TCP supports zero-copy in both direction, while using the normal BSD API's (unlike Netconn). We have never made much noise about this feature because the zero-copy methods are a bit tricky, they go beyond the standard way of doing things.

But yes, I wrote a HTTP- and a FTP-server that use a maximum of zero-copy techniques.
Also most FreeRTOS+TCP drivers use zero-copy. The ESP32 driver has not yet been adapted to use it.


> Most of the features like WiFi provisioning require LwIP stack.

I studied the FreeRTOS+TCP driver for the ESP32, and wrote this [PR 1626](https://github.com/aws/amazon-freertos/pull/1626) about it. it is a simple but good driver.

Could you mention what features that you miss with regards to WiFi?

mahavirj:
@htibosch 

Apart from memory usage in default configuration (and readily available zero copy receive path with `lwIP`), following is most commonly requested feature:

- Simultaneous multi interface and Posix API support (this blocks most commonly requested features for ESP32 which also includes SoftAP provisioning, see more at https://github.com/aws/amazon-freertos/issues/1766)

htibosch:
> memory usage in default configuration

The memory issue is solved, the default configuration was using too much RAM, like defining 60 Network Buffers.

> readily available zero copy receive path with lwIP

Also available within FreeRTOS+TCP.

> following is most commonly requested feature:
> Simultaneous multi interface

This extension, multiple interfaces and IPv6 will be out soon ( within months )

> Posix API support

Initially, FreeRTOS+TCP followed the pure BSD API design, which is like a subset of the POSIX API's.

We are studying other moves to make it more POSIX compatible.

We could add a POSIX translation header file defining things like this:

    #define SO_RCVTIMEO          FREERTOS_SO_RCVTIMEO
    #define socket               FreeRTOS_socket
    #define accept               FreeRTOS_accept

and add header files like `inet.h`, `netdb.h`, `socket.h` and `if.h`, just to make it look more POSIX.

New POSIX-compatible `select()` and/or `poll()` functions are on their way.

( it does have a `FreeRTOS_select()` but with a slightly different calling syntax ).

I am not sure if specific support for WiFi should be part of a TCP/IP stack. I think it is more part of the portable layer, the Network Interface.

One final pro of FreeRTOS+TCP: it cooperates perfectly with the FreeRTOS kernel.
It's structure and way of thinking are inspired by the kernel. It has no need for mutexes, and it makes sporadic use of critical sections because all logic is handled within the IP-task.

But feel free to use whatever TCP/IP stack you like. I only wanted to correct some misunderstandings that I heard from the ESP32 community.

dcgaws:
@pavanmr94 can you tell why we're getting a style check failure on this PR? Thank you.

pavanmr94:
```14:40:58 FAIL: libraries/abstractions/common_io/include/iot_usb_host.h (File size changed from 18964 to 18905)```

shubhamkulkarni97:
@pavanmr94 I have updated branch to latest aws:master

pavanmr94:
@dcgaws the style checks have passed. The build still fails (esp make)

dcgaws:
@shubhamkulkarni97 can you please check our make-based builds for your boards? Thank you.

shubhamkulkarni97:
@dcgaws I tried building aws_demos and aws_tests using make, the builds are successful when using FreeRTOS TCP stack as well as LwIP stack.

shubhamkulkarni97:
There was some delta in sdkconfig, due to which build failed. It should fixed now.

shubhamkulkarni97:
Are the build issues resolved? Can this PR be merged?

PR #1888, opened 2020-04-08
common-io: Refactoring Watchdog test cases

osloapps-max:
Description
-----------
* Refactored test cases to use TEST_PROTECT sections to guard against cascading failures.
* Moved to a single shared bark/bite callback with separate semaphores. This allows for the status to be evaluated as part of the callback instead of waiting for the semaphore to go back. 
* Added a WD reset in the case of a bite interrupt, this as some devices require a reset in order to escape the interrupt. 
* Added global test instance variable to allow the test framework to define which WD instance to test.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_watchdog.c
+++ b/libraries/abstractions/common_io/test/test_iot_watchdog.c
@@ -50,9 +50,17 @@
 /*-----------------------------------------------------------*/
 
 /*-----------------------------------------------------------*/
+
+/* Globals values which can be overwritten by the test
+ * framework invoking these tests */
+/*-----------------------------------------------------------*/
+uint8_t ucTestIotWatchdogInstance = 0; /** Default Flash instance for testing */
+
+/*-----------------------------------------------------------*/
 /** Static globals */
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_watchdog.c
```diff
@@ -49,10 +49,18 @@
...
+ * framework invoking these tests */
+/*-----------------------------------------------------------*/
+uint8_t ucTestIotWatchdogInstance = 0; /** Default Flash instance for testing */
+
 /*-----------------------------------------------------------*/
 /** Static globals */
 /*-----------------------------------------------------------*/
-static SemaphoreHandle_t xtestIotWatchdogSemaphore = NULL;
+static SemaphoreHandle_t xtestIotWatchdogBarkSemaphore = NULL;
+static SemaphoreHandle_t xtestIotWatchdogBiteSemaphore = NULL;
```
dan4thewin, line 63: What is bark and bite?

osloapps-max: As per the API description, Bark is the alert to "feed" the watchdog. Bite is the watchdog actually expiring.

PR #1892, opened 2020-04-10
Retry using exponential backoff upon connection failure with GGD system tests

yourslab:
Description
-----------

GGD_JSONRequestStart method creates a connection to a specified host and port by using the Sockets API. If that connection fails, then the test completely fails. This PR addresses this issue by wrapping the method around RETRY_EXPONENTIAL, which retries the method with exponential backoff if the expected status is not returned. 

Also considered retries for the library function instead but this masks the problem that really needs to be handled in the network layer.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
+++ b/libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
@@ -79,7 +79,7 @@ static BaseType_t prvGGD_JSONRequestGetFileLoop( uint32_t ulBufferSize,
 static BaseType_t prvGGD_JSONRequestStart( const char * pcHostAddress,
                                            uint16_t usGGDPort,
                                            const char * pcThingName,
-                                           Socket_t * pxSocket)
+                                           Socket_t * pxSocket);
 
 TEST_GROUP( GGD_System );
 
@@ -135,9 +135,9 @@ static BaseType_t prvGGD_JSONRequestGetFileLoop( uint32_t ulBufferSize,
 }
 
 static BaseType_t prvGGD_JSONRequestStart( const char * pcHostAddress,
-                                        uint16_t usGGDPort,
-                                        const char * pcThingName,
-                                        Socket_t * pxSocket )
+                                           uint16_t usGGDPort,
+                                           const char * pcThingName,
+                                           Socket_t * pxSocket )
 {
     BaseType_t xStatus;
 
@@ -162,9 +162,9 @@ TEST( GGD_System, JSONRequestAbort )
         GGD_JSONRequestAbort( &xSocket );
 
         prvGGD_JSONRequestStart( clientcredentialMQTT_BROKER_ENDPOINT,
-                              clientcredentialGREENGRASS_DISCOVERY_PORT,
-                              clientcredentialIOT_THING_NAME,
-                              &xSocket );
+                                 clientcredentialGREENGRASS_DISCOVERY_PORT,
+                                 clientcredentialIOT_THING_NAME,
+                                 &xSocket );
         GGD_JSONRequestAbort( &xSocket );
     }
     else
@@ -326,9 +326,9 @@ TEST( GGD_System, GetIPandCertificateFromJSON )
 
         xAutoSearchFlag = pdTRUE;
         xStatus = prvGGD_JSONRequestStart( clientcredentialMQTT_BROKER_ENDPOINT,
-                                        clientcredentialGREENGRASS_DISCOVERY_PORT,
-                                        clientcredentialIOT_THING_NAME,
-                                        &xSocket );
+                                           clientcredentialGREENGRASS_DISCOVERY_PORT,
+                                           clientcredentialIOT_THING_NAME,
+                                           &xSocket );
 
         if( xStatus == pdPASS )
         {
@@ -458,9 +458,9 @@ TEST( GGD_System, JSONRequestGetFile )
          *  @{
          */
         xStatus = prvGGD_JSONRequestStart( clientcredentialMQTT_BROKER_ENDPOINT,
-                                        clientcredentialGREENGRASS_DISCOVERY_PORT,
-                                        clientcredentialIOT_THING_NAME,
-                                        &xSocket );
+                                           clientcredentialGREENGRASS_DISCOVERY_PORT,
+                                           clientcredentialIOT_THING_NAME,
+                                           &xSocket );
 
         if( xStatus == pdPASS )
         {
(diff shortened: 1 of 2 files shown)


libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -166,6 +206,15 @@ TEST( GGD_System, GetGGCIPandCertificate )
                                                   cBuffer, /*lint !e971 can use char without signed/unsigned. */
                                                   ulBufferSize,
                                                   &xHostAddressData );
+            RETRY_EXPONENTIAL( xStatus - GGD_GetGGCIPandCertificate( clientcredentialMQTT_BROKER_ENDPOINT,
```
muneebahmed10, line 209: `xStatus - `? Should it be `=`? Also, the previous call to `GGD_GetGGCIPandCertificate` was not deleted so this will run twice.

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -166,6 +206,15 @@ TEST( GGD_System, GetGGCIPandCertificate )
...
+                                                                     clientcredentialGREENGRASS_DISCOVERY_PORT,
+                                                                     clientcredentialIOT_THING_NAME,
+                                                                     cBuffer, /*lint !e971 can use char without signed/unsigned. */
+                                                                     ulBufferSize,
+                                                                     &xHostAddressData ),
+                               pdPASS,
+                               IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY,
+                               IOT_TEST_GGD_CONNECT_RETRY_COUNT );
```
muneebahmed10, line 217: This is in a `for` loop that runs this function 10 times to see if it succeeds every time. If we're adding a retry to deal with network flakiness, then I don't think it still makes sense to require this function to succeed 10 times in a row.

yourslab: So do you suggest removing the retry or removing the loop?

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
yourslab: Must have had too much caffeine.

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
muneebahmed10: I meant removing the loop would be best here. I'm not sure why we want to have 10 consecutive successes here, and since this line is usually where this test fails, adding a retry would probably help. I think we should remove the loop and add a retry if everyone else is fine with it.

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -158,20 +197,20 @@ TEST( GGD_System, GetGGCIPandCertificate )
...
-                                                  cBuffer, /*lint !e971 can use char without signed/unsigned. */
-                                                  ulBufferSize,
-                                                  &xHostAddressData );
-
-            snprintf( cMsgBuffer, nBufferLength,
-                      "GGD_GetGGCIPandCertificate returned %d on iteration %d",
-                      ( int ) xStatus, ( int ) i );
-            TEST_ASSERT_EQUAL_INT32_MESSAGE( pdPASS, xStatus, cBuffer );
-        }
+        RETRY_EXPONENTIAL( xStatus - GGD_GetGGCIPandCertificate( clientcredentialMQTT_BROKER_ENDPOINT,
```
muneebahmed10, line 200: `=` again

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -158,20 +197,20 @@ TEST( GGD_System, GetGGCIPandCertificate )
...
+                                                                 cBuffer, /*lint !e971 can use char without signed/unsigned. */
+                                                                 ulBufferSize,
+                                                                 &xHostAddressData ),
+                           pdPASS,
+                           IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY,
+                           IOT_TEST_GGD_CONNECT_RETRY_COUNT );
+
+        snprintf( cMsgBuffer, nBufferLength,
+                  "GGD_GetGGCIPandCertificate returned %d on iteration %d",
+                  ( int ) xStatus, ( int ) i );
```
muneebahmed10, line 212: Doesn't seem to make sense anymore since it's not in a loop. I'm not sure what the purpose of `cMsgBuffer` was before this either since it doesn't seem to be used later. Is `i` even used anymore in this function?

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -31,15 +31,34 @@
...
+ */
+#ifndef IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY
+    #define IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY    ( ( uint32_t ) 1100 )
+#endif
+
+/**
+ * @brief The amount of times to retry a request if it fails.
+ */
+#ifndef IOT_TEST_GGD_CONNECT_RETRY_COUNT
+    #define IOT_TEST_GGD_CONNECT_RETRY_COUNT    ( 1 )
```
dan4thewin, line 54: If this is 1, does that mean there is no retry, but rather a single attempt?

yourslab: 1 means a single attempt. I shall comment that out. Thank you!

libraries/freertos_plus/aws/greengrass/test/aws_test_ggd_system.c
```diff
@@ -31,15 +31,34 @@
...
 
+/**
+ * @brief The initial delay in milliseconds that is doubled each retry of request.
+ */
+#ifndef IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY
+    #define IOT_TEST_GGD_INITIAL_CONNECTION_RETRY_DELAY    ( ( uint32_t ) 1100 )
+#endif
+
+/**
+ * @brief The amount of times to retry a connection if it fails with 1 being a single attempt.
```
muneebahmed10, line 51: Wording still seems a bit off to me. How would you feel about something like:
```
The number of times to try a connection. Values greater than 1 will retry on failure with an exponential backoff.
```

yourslab: Sounds good

PR #1893, opened 2020-04-10
Retry using exponential backoff upon connection failure with Defender system tests

yourslab:
If IotMqtt_Connect connection fails, then the test completely fails. This PR addresses this issue by wrapping the method around RETRY_EXPONENTIAL, which retries the method with exponential backoff if the expected status is not returned.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/aws/defender/test/system/aws_iot_tests_defender_system.c
+++ b/libraries/c_sdk/aws/defender/test/system/aws_iot_tests_defender_system.c
@@ -517,6 +517,9 @@ static IotMqttError_t _startMqttConnection( void )
     IotMqttError_t mqttError = IOT_MQTT_SUCCESS;
     IotMqttNetworkInfo_t mqttNetworkInfo = ( IotMqttNetworkInfo_t ) IOT_MQTT_NETWORK_INFO_INITIALIZER;
     IotMqttConnectInfo_t mqttConnectionInfo = ( IotMqttConnectInfo_t ) IOT_MQTT_CONNECT_INFO_INITIALIZER;
+    /* AWS IoT Service limits only allow 1 connection per MQTT client ID per second.
+     * Wait until 1100 ms have elapsed since the last connection. */
+    uint32_t periodMs = 1100;
 
     if( !_mqttConnectionStarted )
     {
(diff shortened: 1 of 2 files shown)


libraries/c_sdk/aws/defender/test/system/aws_iot_tests_defender_system.c
```diff
@@ -516,10 +533,16 @@ static IotMqttError_t _startMqttConnection( void )
         mqttConnectionInfo.pClientIdentifier = AWS_IOT_TEST_DEFENDER_THING_NAME;
         mqttConnectionInfo.clientIdentifierLength = ( uint16_t ) strlen( AWS_IOT_TEST_DEFENDER_THING_NAME );
 
-        mqttError = IotMqtt_Connect( &mqttNetworkInfo,
-                                     &mqttConnectionInfo,
-                                     1000,
-                                     &_mqttConnection );
+        /* AWS IoT Service limits only allow 1 connection per MQTT client ID per second.
+         * Wait until 1100 ms have elapsed since the last connection. */
+        uint32_t periodMs = 1100;
```
muneebahmed10, line 538: Since this file is from the C SDK, variables [should be](https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/main/guide_developer_styleguide.html) declared at the top of a function.

libraries/c_sdk/aws/defender/test/system/aws_iot_tests_defender_system.c
```diff
@@ -76,6 +79,23 @@
...
+ * @cond DOXYGEN_IGNORE
+ * Doxygen should ignore this section.
+ *
+ * Provide default values of test configuration constants.
+ */
+#ifndef IOT_TEST_MQTT_CONNECT_INITIAL_RETRY_DELAY
+    #define IOT_TEST_MQTT_CONNECT_INITIAL_RETRY_DELAY    ( 1100 )
+#endif
+#ifndef IOT_TEST_MQTT_CONNECT_RETRY_COUNT
+    #define IOT_TEST_MQTT_CONNECT_RETRY_COUNT            ( 1 )
```
gkwicker, line 92: I'd suggest a higher count here, for example 6 ([the value used in TCP tests.](https://github.com/aws/amazon-freertos/blob/master/tests/include/aws_test_tcp.h#L58)).

PR #1897, opened 2020-04-10
Make periodMs a config macro

yourslab:
Because we are now using periodMs throughout all system tests, it makes sense to make it a config macro in iot_config_common.h. 

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
+++ b/libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
@@ -233,7 +233,7 @@ static IotMqttError_t _mqttConnect( const IotMqttNetworkInfo_t * pNetworkInfo,
 
     int32_t retryCount = 0;
 
-    uint32_t periodMs = IOT_TEST_MQTT_CONNECT_INITIAL_RETRY_DELAY;
+    uint32_t periodMs = IOT_TEST_MQTT_CONNECT_INIT_RETRY_DELAY_MS;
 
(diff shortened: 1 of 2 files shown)


libraries/c_sdk/standard/mqtt/test/system/iot_tests_mqtt_system.c
```diff
@@ -60,13 +60,16 @@
...
 #ifndef IOT_TEST_MQTT_TIMEOUT_MS
-    #define IOT_TEST_MQTT_TIMEOUT_MS             ( 5000 )
+    #define IOT_TEST_MQTT_TIMEOUT_MS                     ( 5000 )
 #endif
 #ifndef IOT_TEST_MQTT_TOPIC_PREFIX
-    #define IOT_TEST_MQTT_TOPIC_PREFIX           "iotmqtttest"
+    #define IOT_TEST_MQTT_TOPIC_PREFIX                   "iotmqtttest"
+#endif
+#ifndef IOT_TEST_MQTT_CONNECT_INITIAL_RETRY_DELAY
+    #define IOT_TEST_MQTT_CONNECT_INITIAL_RETRY_DELAY    ( 1100 )
```
sarenameas, line 69: nitpick: I think we need to append a MS in this macro name. (Just to be consistent with having the units like line 63)

yourslab: okay agreed

PR #1899, opened 2020-04-13
Refactor mbedtls_error utilities to use switch-case instead of if-els…

aggarw13:
Optimize implementation of the custom mbedTLS stringification utilities with `switch-case`

Description
-----------
Use of switch-case is more space efficient than if-else-if checks. Thus, update implementation of `mbedtls_sterror_highlevel` and `mbedtls_strerror_lowlevel` utilities to use switch-case.

Performed sanity test for build outputs with GNU ARM GCC between if-else-if and swtich-case based logic on test files within [this folder](https://github.com/aws/amazon-freertos/files/4471792/test-build-files.zip). 
Test Results with and without `-Os` flag
```
/*  Binary sizes without -Os flag */
╰─➤  size test_if_else                                                                                                                                                  1 ↵
   text    data     bss     dec     hex filename
   1988    1088      64    3140     c44 test_if_else
╭─ubuntu@ip-172-31-24-209 ~/Experiments
╰─➤  size test_switch_case
   text    data     bss     dec     hex filename
   1936    1088      64    3088     c10 test_switch_case
```

```
/* Binary sizes with -Os flag */
╭─ubuntu@ip-172-31-24-209 ~/Experiments
╰─➤  size test_if_else
   text    data     bss     dec     hex filename
   1628    1088      64    2780     adc test_if_else
╭─ubuntu@ip-172-31-24-209 ~/Experiments
╰─➤  size test_switch_case
   text    data     bss     dec     hex filename
   1628    1088      64    2780     adc test_switch_case
```
Note: Build outputs are same with optimization as the `-Os` flag causes the compiler to pre-calculate both the if-else-if and switch-case logic results and replace the call-site with the results. However, this behavior is compiler dependent, and we want the utilities implementation to be platform-agnostic.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/3rdparty/mbedtls_utils/mbedtls_error.c
+++ b/libraries/3rdparty/mbedtls_utils/mbedtls_error.c
@@ -1323,4 +1323,14 @@ const char * mbedtls_strerror_lowlevel( int errnum )
 
         #if defined( MBEDTLS_XTEA_C )
             case -( MBEDTLS_ERR_XTEA_INVALID_INPUT_LENGTH ):
-                rc = "XTEA - The data input ha
+                rc = "XTEA - The data input has an invalid length";
+                break;
+
+            case -( MBEDTLS_ERR_XTEA_HW_ACCEL_FAILED ):
+                rc = "XTEA - XTEA hardware accelerator failed";
+                break;
+        #endif /* MBEDTLS_XTEA_C */
+    }
+
+    return rc;
+}
--- a/libraries/3rdparty/mbedtls_utils/mbedtls_error.c
+++ b/libraries/3rdparty/mbedtls_utils/mbedtls_error.c
@@ -223,631 +223,631 @@ const char * mbedtls_strerror_highlevel( int errnum )
         switch( use_ret )
         {
             #if defined( MBEDTLS_CIPHER_C )
-                case -( MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE ) )
+                case -( MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE )
                     rc = "CIPHER - The selected feature is not available";
                     break;
 
-                case -( MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA ) )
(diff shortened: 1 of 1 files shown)


libraries/3rdparty/mbedtls_utils/mbedtls_error.c
```diff
@@ -220,1364 +220,1117 @@ const char * mbedtls_strerror_highlevel( int errnum )
...
+
+                case -( MBEDTLS_ERR_RSA_VERIFY_FAILED )
+                    rc = "RSA - The PKCS#1 verification failed";
+                    break;
+
+                case -( MBEDTLS_ERR_RSA_OUTPUT_TOO_LARGE )
+                    rc = "RSA - The output buffer for decryption is not large enough";
+                    break;
+
+                case -( MBEDTLS_ERR_RSA_RNG_FAILED )
```
lundinc2, line 538: this doesn't look syntactically correct, there should be a colon after the case statement

aggarw13: Fixed

libraries/3rdparty/mbedtls_utils/mbedtls_error.c
```diff
@@ -220,1364 +220,1117 @@ const char * mbedtls_strerror_highlevel( int errnum )
...
+        #if defined( MBEDTLS_XTEA_C )
+            case -( MBEDTLS_ERR_XTEA_INVALID_INPUT_LENGTH ):
+                rc = "XTEA - The data input has an invalid length";
+                break;
+
+            case -( MBEDTLS_ERR_XTEA_HW_ACCEL_FAILED ):
+                rc = "XTEA - XTEA hardware accelerator failed";
+                break;
+        #endif /* MBEDTLS_XTEA_C */
+    }
```
muneebahmed10, line 1333: Should there be a `default:` for this `switch`?

aggarw13: Good question. The default value will continue to be the `NULL` initialized value of the return variable in this case.

PR #1908, opened 2020-04-15
common-io: Refactoring and updates to PWM test cases

osloapps-max:
Description
-----------
* Added "TEST_PROTECT" sections where needed
* Removed stale includes, variables and code
* Added in "AFQP_IotPwm_StartFuzzing" and "AFQP_IotPwmAccuracy" test cases in the group.
* Removed "AFQP_IotPwmOpenFailure" as it was testing the same as "AFQP_IotPwmOpenFuzzing" 
* Refactored "AFQP_IotPwmAccuracy" and "AFQP_IotPwmAccuracyAssisted", moving common code out to shared private function. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_pwm.c
+++ b/libraries/abstractions/common_io/test/test_iot_pwm.c
@@ -36,41 +36,28 @@
 /* Driver includes */
 #include "FreeRTOS.h"
 #include "semphr.h"
-#include "task.h"
-#include <time.h>
-#include <sys/time.h>
+
 #include "iot_pwm.h"
-#include "iot_board_gpio.h"
 #include "iot_gpio.h"
-#include "test_iot_config.h"
+#include "iot_test_common_io_internal.h"
 
+/*-----------------------------------------------------------*/
 
+#define PWM_TEST_DURATION_MSEC    5000
 
 /*-----------------------------------------------------------*/
 
 /* Globals values which can be overwritten by the test
  * framework invoking these tests */
 /*-----------------------------------------------------------*/
-uint32_t ultestIotPwmGpioOutputPin = 14;
 uint32_t ultestIotPwmGpioInputPin = 18;
-uint32_t ultestIotPwmGpioFunction = 0x4U; /*IOPCTL_PIO_FUNC4; */
-uint32_t ulAssistedTestIotPwmGpioInputPin = 23;
 
 uint32_t ultestIotPwmInstance = 2;            /* Use PWM instance 2 */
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_pwm.c
```diff
@@ -420,109 +400,42 @@ TEST( TEST_IOT_PWM, AFQP_IotPwmAccuracy )
...
-}
-
-/*-----------------------------------------------------------*/
-
-/**
- * @brief Assited Test Function to test pwm accuracy by running the
- * pwm signal for 5 seconds and also counting the number of pulses
- *
- */
-TEST( TEST_IOT_PWM, AFQP_IotPwmAccuracyAssisted )
```
qiutongs, line 477: Please keep "assisted" test cases. 

Assisted test cases are expected to not be called in TEST_GROUP_RUNNER. They are supposed to be called via other mechanism.

osloapps-max: Is it not an option to call AFQP_IotPwmAccuracy via the other mechanism right away? The two tests was identical to each other in every way except the name of the global intance/gpio variable.

vjjunnut: Please keep the test name as there are other scripts that depend on this test name.

osloapps-max: The assisted test case has now been added back. refactored that and the non-assisted version so that they share the same test code.

libraries/abstractions/common_io/test/test_iot_pwm.c
```diff
@@ -417,42 +397,45 @@ TEST( TEST_IOT_PWM, AFQP_IotPwmAccuracy )
...
-    /* Start the PWM signal */
-    lRetVal = iot_pwm_start( xPwmHandle );
-    TEST_ASSERT_EQUAL( IOT_PWM_SUCCESS, lRetVal );
+        /* Start the PWM signal */
+        lRetVal = iot_pwm_start( xPwmHandle );
+        TEST_ASSERT_EQUAL( IOT_PWM_SUCCESS, lRetVal );
 
-    /* Delay for 5 sec. */
-    vTaskDelay( PWM_TEST_DURATION_MSEC / portTICK_PERIOD_MS );
+        /* Delay for 5 sec. */
```
gkwicker, line 436: Note that this comment will no longer be accurate if someone changes the value of `PWM_TEST_DURATION_MSEC`.  There are several comments with this issue, e.g. line 455.

PR #1920, opened 2020-04-20
Ensure BLE_SUPPORTED is enabled correctly.

RaymondNgun:
Ensure BLE_SUPPORTED is enabled correctly.

Description
-----------
Remove from iot_config.h since it cannot be changed via compile time
anymore. Adjusted aws_test to reference CY_BLE_SUPPORTED instead of
BLE_SUPPORTED.

Checklist:
----------
- [ ] My code is Linted.


RaymondNgun:
@mingyue86010, thank you for correcting the BLE issues!  With that said, I believe it would be better to fix according to this patch here.  This way, I can still remove CY_BLE_SUPPORTED and still have it compile.

PR #1928, opened 2020-04-21
Fix Full_TCP test retry logic & disable DNS cache on all builds

gkwicker:
Description
-----------

We observed a test failure in continuous integration testing in which no retries to the AWS IoT endpoint succeeded.  There are two root causes:

1. The Full_TCP test suite was not re-resolving hostnames as part of the retry logic.
1. The FreeRTOS+TCP DNS cache stores only a single IP address, even though the DNS query
response returned a list of 6 of them.  The one that FreeRTOS+TCP cached (the first one in the response) was no longer valid.  Until the TTL expired, no connections to that endpoint could succeed.

This change addresses 1) by reworking the Full_TCP test suite retry logic, and 2) by disabling the FreeRTOS+TCP DNS cache in all demo and test builds.

See the following issue for more details; if this feature gets implemented in FreeRTOS+TCP, we would be able to re-enable the DNS cache in the demo and test builds.

https://github.com/FreeRTOS/FreeRTOS/issues/58

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -350,8 +350,8 @@ static BaseType_t prvAwsIotBrokerConnectHelper( Socket_t xSocket,
 }
 /*-----------------------------------------------------------*/
 
-static BaseType_t prvConnectHelper( Socket_t xSocket,
-                                    Server_t xConn )
+static BaseType_t prvConnect( Socket_t xSocket,
+                              Server_t xConn )
 {
     BaseType_t xResult = pdFAIL;
     SocketsSockaddr_t xAddress;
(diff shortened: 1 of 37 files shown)


htibosch:
Hi @gkwicker: in stead of disabling `ipconfigUSE_DNS_CACHE`, you can also clear the DNS cache (by calling `FreeRTOS_dnsclear()`) before calling `gethostbyname()`:
It would result in a much smaller patch.

In the early UDP-only library there was already:
~~~c
    uint32_t FreeRTOS_gethostbyname( const char *pcHostName )
~~~
As we always preserved backward compatibility, it was never changed to return a host-entry:
~~~c
    struct hostent *gethostbyname(const char *name);
~~~

Now that FreeRTOS+TCP will be extended with IPv6, it is a good moment to add the (POSIX-) compatible:
~~~C
int getaddrinfo( const char *node, const char *service,
                 const struct addrinfo *hints,
                 struct addrinfo **res);

void freeaddrinfo(struct addrinfo *res);
~~~
The code has been prepared for the IPv6 branch, but it can be adapted to work in the current master release.

Personally, I would rather see that `ipconfigUSE_DNS_CACHE` remains enabled by default.

gkwicker:
> Hi @gkwicker: in stead of disabling `ipconfigUSE_DNS_CACHE`, you can also clear the DNS cache (by calling `FreeRTOS_dnsclear()`) before calling `gethostbyname()`:
> It would result in a much smaller patch.
> 
While we can do that for this test group, we have many test groups which contact load-balanced services that do not use the same connection & retry logic and are also affected by this issue.  That patch would be much larger and more difficult.

Once we have a fix for issue 58 available in FreeRTOS+TCP, we will re-enable the DNS cache in all of the demos and tests in this repository.

PR #1932, opened 2020-04-23
common-io: Refactoring and updates to Temp Sensor test cases

osloapps-max:
* Added TEST_PROTECT sections
* Removed AFQP_IotTsensorCalibrationNullInputFuzz as the test was not valid according the API description stating that this request will not return invalid status on NULL buffer.
* Updated AFQP_IotTsensorCalibrationSuccess to test for status when passing in NULL buffer
* Reworked AFQP_IotTsensorTriggerMinThreshold and AFQP_IotTsensorTriggerMaxThreshold. 
* Updated expected return values based on what is defined by the API header file

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_tsensor.c
+++ b/libraries/abstractions/common_io/test/test_iot_tsensor.c
@@ -63,6 +63,17 @@ TEST_GROUP( TEST_IOT_TSENSOR );
  */
 TEST_SETUP( TEST_IOT_TSENSOR )
 {
+    if( xtestIotTsensorMinSemaphore == NULL )
+    {
+        xtestIotTsensorMinSemaphore = xSemaphoreCreateBinary();
+        TEST_ASSERT_NOT_EQUAL( NULL, xtestIotTsensorMinSemaphore );
+    }
+
+    if( xtestIotTsensorMaxSemaphore == NULL )
+    {
+        xtestIotTsensorMaxSemaphore = xSemaphoreCreateBinary();
+        TEST_ASSERT_NOT_EQUAL( NULL, xtestIotTsensorMaxSemaphore );
+    }
 }
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_tsensor.c
```diff
@@ -79,6 +81,16 @@ TEST_TEAR_DOWN( TEST_IOT_TSENSOR )
  */
 TEST_GROUP_RUNNER( TEST_IOT_TSENSOR )
 {
+    if( xtestIotTsensorMinSemaphore == NULL )
+    {
+        xtestIotTsensorMinSemaphore = xSemaphoreCreateBinary();
```
gkwicker, line 86: In general, you should always check return codes here since [`xSemaphoreCreateBinary()`](https://freertos.org/xSemaphoreCreateBinary.html) can return `NULL`.

osloapps-max: It makes sense. I moved the block out to the test setup phase and added asserts to check against NULL return.

libraries/abstractions/common_io/test/test_iot_tsensor.c
```diff
@@ -61,6 +63,17 @@ TEST_GROUP( TEST_IOT_TSENSOR );
  */
 TEST_SETUP( TEST_IOT_TSENSOR )
 {
+    if( xtestIotTsensorMinSemaphore == NULL )
```
cobusve, line 66: It think the test should check this and assert and fail if it already exists. What will happen if a prior test hands you a locked semaphore?

cobusve: There are a couple of these further down also, this applies to all of them.

qiutongs: I fixed them.

libraries/abstractions/common_io/test/test_iot_tsensor.c
```diff
@@ -39,6 +39,7 @@
 #include "FreeRTOS.h"
 #include "semphr.h"
 
+#define testIotTSENSOR_DELAY              pdMS_TO_TICKS( 5000 )
```
gkwicker, line 42: In general when a `#define` macro or variable has units, add the units to the end of the name.  This helps reviewers and maintainers spot misuse.  Here we would prefer the name `testIoTSENSOR_DELAY_TICKS`.

libraries/abstractions/common_io/test/test_iot_tsensor.c
```diff
@@ -605,28 +638,34 @@ TEST( TEST_IOT_TSENSOR, AFQP_IotTsensorTriggerMinThreshold )
...
-        TEST_ASSERT_EQUAL( pdTRUE, lRetVal );
-    }
+        /* Read out current temperature */
+        lRetVal = iot_tsensor_get_temp( xTsensorHandle, &setLowThreshold );
+        TEST_ASSERT_EQUAL( IOT_TSENSOR_SUCCESS, lRetVal );
 
-    lRetVal = iot_tsensor_ioctl( xTsensorHandle, eTsensorSetMinThreshold, &setLowThreshold );
+        /* Set min threshold to be 5 degrees C higher than current temperature.
+         * Expect it to fire immediately */
+        setLowThreshold += 5000;
```
gkwicker, line 656: These values are in _millidegrees Celsius_, consider a naming convention so that maintainers/readers can ensure the values are used & converted properly.

PR #1943, opened 2020-04-27
Export the compile commands in CMake

tgsong:
Description
-----------

Export the json compile commands file from CMake. This will generate a `compile_commands.json` file under the build directory, which contains the compiler flags for each file.

**Note: this only works for non-IDE generators like makefile and ninja. It doesn't work for IDE generators since they typically hide the compiler commands.**

Checklist:
----------

- [x] My code is Linted.


gkwicker:
Will the .json output be present in our CI build artifacts?

PR #1944, opened 2020-04-27
Fix breaking CBMC proof for DNSgetHostByName_a/_cancel & ReadNameField

AniruddhaKanhere:
Fix the breaking CBMC proof for DNSgetHostByName_a & DNSgetHostByName_cancel & ReadNameField.

Description
-----------
Added a #define in all three Makefile.json to declare an essential macro ipconfigDNS_CACHE_NAME_LENGTH used in CBMC proofs which is not defined if (ipconfigUSE_DNS_CACHE == 0).
Previous PR https://github.com/aws/amazon-freertos/pull/1928 disabled the use of DNS cache which broke the proofs.

Checklist:
----------
- [ ] My code is Linted.


gkwicker:
It looks like the `ReadNameField` proof is still failing.

PR #1949, opened 2020-04-28
Align compiler flags with IDE build

gkwicker:
Description
-----------

Incorrect compiler flags in the cmake build were causing tests in CI to fail due to stack overflow.  This change aligns the cmake build's flags with those of the IDE build and fixes the problem.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/include/aws_clientcredential_keys.h
+++ b/demos/include/aws_clientcredential_keys.h
@@ -37,7 +37,7 @@
  * "...base64 data...\n"\
  * "-----END CERTIFICATE-----\n"
  */
-#define keyCLIENT_CERTIFICATE_PEM                   ""
+#define keyCLIENT_CERTIFICATE_PEM                   NULL
 
 /*
  * @brief PEM-encoded issuer certificate for AWS IoT Just In Time Registration (JITR).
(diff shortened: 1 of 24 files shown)


gkwicker:
> Was the cmake build not defining the MBEDTLS_CONFIG_FILE macro before this change? Typically this should not be defined in the vendor board directory unless the board requires a config file that is different from our default. It is currently being defined in the main cmake file if the user has not already defined it.
> 
> https://github.com/aws/amazon-freertos/blob/24d5217a7e11903f44e815f3f3928ab4542addaa/CMakeLists.txt#L90

It was defining `MBEDTLS_CONFIG_FILE` in both IDE and cmake builds, however `MBEDTLS_USER_CONFIG_FILE` wasn't defined in the cmake build prior to the change, but it was defined in the IDE build.  Nuvoton has their specific version of aws_mbedtls_user_config.h [here](https://github.com/aws/amazon-freertos/blob/master/vendors/nuvoton/boards/numaker_iot_m487_wifi/aws_tests/application_code/nuvoton_code/mbedtls_user_config.h).

gkwicker:
Closing, will reopen after correct rebase

PR #1953, opened 2020-04-28
Fix msvc compiler warnings in iot_pkcs11_mbedtls.c

lundinc2:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -128,7 +128,7 @@ typedef int ( * pfnMbedTlsSign )( void * ctx,
  * @brief Indicates that no PKCS #11 operation is underway for given session.
  *
  */
-#define pkcs11NO_OPERATION     ( ( CK_MECHANISM_TYPE ) 0xFFFFul )
+#define pkcs11NO_OPERATION     ( ( CK_MECHANISM_TYPE ) -1 )
 
 /**
  * @ingroup pkcs11_macros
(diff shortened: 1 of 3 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -128,7 +128,7 @@ typedef int ( * pfnMbedTlsSign )( void * ctx,
  * @brief Indicates that no PKCS #11 operation is underway for given session.
  *
  */
-#define pkcs11NO_OPERATION     ( ( CK_MECHANISM_TYPE ) 0xFFFFFFFFF )
+#define pkcs11NO_OPERATION     ( ( CK_MECHANISM_TYPE ) 0xFFFFul )
```
cobusve, line 132: Richard pointed this one out to me, it has 9 "f"'s and now it has 4, that is 36 bits vs 16 bits. It this correct? Should the value be a ul with only the bottom 16 bits assigined? Seems odd to me. Most of the time something like this is best done doing  (type)(-1) or similar ...

lundinc2: Oops, this should be 8F's

lundinc2: Applied the -1 suggestion

cobusve: That makes more sense then !

cobusve: The issue I had here is that unsigned long is not necessarily 32 bits, if it is more this is not the same as -1, if you cast -1 to the type it will always be as many F's as you can fit into the variable ...

aggarw13: Wouldn't this cause compiler warnings of signedness conversion in cast?

aggarw13: I tested with gcc-7 and interestingly, it does not raise a warning even with ` -Wall -Wextra -Wpedantic`  flags passed.

Issue #1915, opened 2020-04-17
cbor decoder - getting last string in map with indefinite length has an offset of 1 byte


rvanputten:
**Describe the bug**
Some of the cbor responses from the aws cloud contain indefinite length maps.
If the last value of the map is a text string (probably also byte string) when getting the value using the find function of the cbor decoder, the value has an offset of 1.

Example: This is the response I get from an invalid create cert request:
`BF 6A 73 74 61 74 75 73 43 6F 64 65 19 01 90 69 65 72 72 6F 72 43 6F 64 65 6E 49 6E 76 61 6C 69 64 50 61 79 6C 6F 61 64 6C 65 72 72 6F 72 4D 65 73 73 61 67 65 78 18 4D 65 73 73 61 67 65 20 63 61 6E 6E 6F 74 20 62 65 20 70 61 72 73 65 64 FF`
I checked the validity of the cbor using http://cbor.me/. The JSON representation is this:
`{"statusCode": 400, "errorCode": "InvalidPayload", "errorMessage": "Message cannot be parsed"}`

When I print the errorMessage from my code, I get this: `essage cannot be parsed�`

If a map doesn't specify the length upfront but uses an indefinite length map https://tools.ietf.org/html/rfc7049#section-2.2, a special break character is inserted after the values of the map: https://tools.ietf.org/html/rfc7049#section-2.2.1

I think I've narrowed down the problem to this line: https://github.com/aws/amazon-freertos/blob/master/libraries/c_sdk/standard/serializer/src/cbor/iot_serializer_tinycbor_decoder.c#L251
which doesn't take the extra break character into account.

**System information**
- Which hardware board or part numbers?
`esp32`
- IDE used
`CLion`
- Operating System [Windows|Linux|MacOS]
`Ubuntu`
- Version of FreeRTOS (run ``git describe --tags`` to find it)
`201912.00-86-g94797e54a`
- Project [Custom Application|MQTT|Shadow|Greengrass|OTA|Device Defender|TCP Echo|Sub pub Demo|Simple TCP Server]
`Custom Application`
- If your project is a Custom Application, please add the relevant code snippet in the section `Code to reproduce the bug`. 

**Expected behavior**
The strings is returned complete and without offset:
`Message cannot be parsed`

**Screenshots or console output**
Actual result:
`essage cannot be parsed�`
As you can see, the first byte is missing and the break character is printed

**Code to reproduce the bug**
I've removed error checking for readability but none of the calls give an error.
[...]

PR #1955, opened 2020-04-29
Handle indefinite length case of map/array when decoding string

qiutongs:
Description
-----------

This is to fix this bug reported by customer. https://github.com/aws/amazon-freertos/issues/1915

When decoding definite-length text/binary string, our current code has an optimization by returning the pointer in the decoded buffer directly, thus avoiding copying data around.

But it didn't handle the case of indefinite length case of map/array. In that case, a special break character 0xFF is appended in the end of map/array. Therefore our calculation of the pointer offsets by 1. 

Checklist:
----------

- [x] My code is Linted.


qiutongs:
Fix should be applied to C-SDK first and then sync to Amazon FreeRTOS. Temporarily close this PR.

PR #1959, opened 2020-04-30
Test: simplify adding new unit tests

alfred2g:
Simplify Unit test cmake for a better user experience adding new unit tests

Description
-----------
Simplify the cmake file for unit tests

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/utest/CMakeLists.txt
+++ b/libraries/abstractions/pkcs11/utest/CMakeLists.txt
@@ -1,142 +1,135 @@
 project ("abstractions pkcs11 cmock unit test")
 cmake_minimum_required (VERSION 3.13)
-set(3rdparty_dir "${AFR_ROOT_DIR}/libraries/3rdparty")
-set(abstractions_dir "${AFR_ROOT_DIR}/libraries/abstractions")
-set(standard_dir "${AFR_ROOT_DIR}/libraries/freertos_plus/standard")
-set(kernel_dir "${AFR_ROOT_DIR}/freertos_kernel")
-set(common_dir "${AFR_ROOT_DIR}/libraries/c_sdk/standard/common")
 
-# Mock library
-list(APPEND mock_list
-    "${abstractions_dir}/pkcs11/include/iot_pkcs11_pal.h"
-    "${kernel_dir}/include/portable.h"
-    "${kernel_dir}/include/queue.h"
-    "${common_dir}/include/iot_logging_task.h"
-    "${standard_dir}/crypto/include/iot_crypto.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/platform_util.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/ctr_drbg.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/sha256.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/base64.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/bignum.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/entropy.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/md.h"
-    "${3rdparty_dir}/mbedtls/include/mbedtls/rsa.h"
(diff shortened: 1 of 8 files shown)


libraries/abstractions/pkcs11/utest/CMakeLists.txt
```diff
@@ -1,142 +1,135 @@
...
-       "${3rdparty_dir}/mbedtls/include"
-       "${3rdparty_dir}/mbedtls_utils"
-       "${standard_dir}/crypto/include"
-       "${AFR_ROOT_DIR}/tests/include"
-       "${AFR_ROOT_DIR}/libraries/c_sdk/standard/common/include/private"
-       "${kernel_dir}/include"
-       "${CMAKE_CURRENT_BINARY_DIR}/mocks"
+# list the directories the module under test includes
+    list(APPEND real_include_directories
+                "${3rdparty_dir}/pkcs11" # should come first
```
lundinc2, line 61: This is getting included twice now, is that necessary?

libraries/freertos_plus/standard/pkcs11/utest/CMakeLists.txt
```diff
@@ -1,93 +1,116 @@
...
-preprocess_mock_list(pkcs11_mock ${preprocessed_mock_list} "${unix_flags}")
-
-
-target_compile_definitions(pkcs11_mock PUBLIC
-        portHAS_STACK_OVERFLOW_CHECKING=1
-        portUSING_MPU_WRAPPERS=1
-        MPU_WRAPPERS_INCLUDED_FROM_API_FILE
+    project ("freertos_plus pkcs11 cmock unit test")
+    cmake_minimum_required (VERSION 3.13)
+    set(3rdparty_dir "${AFR_ROOT_DIR}/libraries/3rdparty")
```
lundinc2, line 3: looks like we can remove these to align with the other mock suites

alfred2g: oups, i forgot about this one
i will remove it

tools/cmock/create_test.cmake
```diff
@@ -109,10 +122,42 @@ function(create_mock_list mock_name mock_list cmock_config)
...
+                             real_include_list
+                             mock_name)
+    add_library(${target} STATIC
+            ${src_file}
+        )
+    target_include_directories(${target} PUBLIC
+            ${real_include_list}
+            )
+    set_target_properties(${target} PROPERTIES
+                COMPILE_FLAGS " \
```
lundinc2, line 149: It might be desirable to also add -Wpedantic and -Wextra for the real code. What are your thoughts on this?

alfred2g: sounds good, i will add them, there is a failing check I will fix both

vendors/pc/boards/linux/CMakeLists.txt
```diff
@@ -2,10 +2,18 @@ if (AFR_ENABLE_UNIT_TESTS)
 
     set(CMAKE_C_FLAGS_DEBUG "-ggdb3 -Og")
     set(CMAKE_C_FLAGS_RELEASE "-O3")
-    set(CMAKE_C_FLAGS "-Wall")
+    set(CMAKE_C_FLAGS "-Wall -Wextra -Wpedantic")
```
lundinc2, line 5: Does this add these warnings to the mocks as well? That might create a lot of noise when compiling

alfred2g: Yes it does actually, I will fix that

vendors/pc/boards/linux/CMakeLists.txt
```diff
@@ -6,6 +6,14 @@ if (AFR_ENABLE_UNIT_TESTS)
 
     set(CMAKE_SHARED_LINKER_FLAGS "-fPIC")
 
+    set(3rdparty_dir "${AFR_ROOT_DIR}/libraries/3rdparty")
```
yourslab, line 9: We could also use AFR_3RDPARTY_DIR, right? Then there's AFR_MODULES_*_DIR. Any reason why we're not using these variables declared in afr.cmake instead?

PR #1967, opened 2020-05-04
iot_pkcs11_mbedtls.c Unit Tests

lundinc2:
* Slight refactor to existing tests
* C_GetSlotList 100% line and branch coverage
* C_GetTokenInfo Happy Path
* C_GetMechanismInfo 100% line and branch coverage
* C_InitToken 100% line and branch coverage
* C_OpenSession Happy Path
* C_Login 100% line and branch coverage
* C_CreateObject Happy Path
	* RSA private key
	* EC private key

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
+++ b/libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
@@ -753,8 +753,6 @@ void test_pkcs11_C_CreateObjectECPrivKey( void )
     mbedtls_ecp_keypair_init_CMockIgnore();
     mbedtls_ecp_group_init_CMockIgnore();
     mbedtls_ecp_group_load_IgnoreAndReturn( 0 );
-
-    /* current callstack */
     mbedtls_mpi_read_binary_IgnoreAndReturn( 0 );
     pvPortMalloc_Stub( pvPkcs11MallocCb );
     mbedtls_pk_write_key_der_IgnoreAndReturn( 1 );
--- a/libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
+++ b/libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
@@ -175,11 +175,11 @@ int suiteTearDown( int numFailures )
  * @brief Helper function to initialize PKCS #11.
  *
  */
-static CK_RV prvInitializePkcs11( SemaphoreHandle_t pxPKCS11Mutex )
+static CK_RV prvInitializePkcs11()
 {
     CK_RV xResult = CKR_OK;
 
-    xQueueCreateMutex_IgnoreAndReturn( pxPKCS11Mutex );
+    xQueueCreateMutex_IgnoreAndReturn( 1 );
     CRYPTO_Init_Ignore();
     mbedtls_entropy_init_Ignore();
     mbedtls_ctr_drbg_init_Ignore();
@@ -219,8 +219,6 @@ static CK_RV prvOpenSession( CK_SESSION_HANDLE_PTR pxSession )
     xQueueCreateMutex_IgnoreAndReturn( ( SemaphoreHandle_t ) &xResult );
     xResult = C_OpenSession( 0, xFlags, NULL, 0, pxSession );
 
-    TEST_ASSERT_EQUAL( CKR_OK, xResult );
-
     return xResult;
 }
 
@@ -238,8 +236,6 @@ static CK_RV prvCloseSession( CK_SESSION_HANDLE_PTR pxSession )
     mbedtls_sha256_free_CMockIgnore();
     xResult = C_CloseSession( *pxSession );
 
-    TEST_ASSERT_EQUAL( CKR_OK, xResult );
-
     return xResult;
 }
 
@@ -253,7 +249,7 @@ void test_pkcs11_C_Initialize( void )
 {
     CK_RV xResult = CKR_OK;
 
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
 
     xResult = prvUninitializePkcs11();
@@ -269,7 +265,7 @@ void test_pkcs11_C_InitializeMemFail( void )
     CK_RV xResult = CKR_OK;
 
     xQueueCreateMutex_IgnoreAndReturn( NULL );
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
 
     TEST_ASSERT_EQUAL( CKR_HOST_MEMORY, xResult );
 }
@@ -286,7 +282,7 @@ void test_pkcs11_C_InitializeSeedFail( void )
 
     /* Hack: Giving out pointer to xResult when creating a mutex. This unit testcase
      * will never use the actual mutex. */
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
 
     TEST_ASSERT_EQUAL( CKR_FUNCTION_FAILED, xResult );
 }
@@ -299,7 +295,7 @@ void test_pkcs11_C_InitializeInitTwice( void )
 {
     CK_RV xResult = CKR_OK;
 
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
 
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
 
@@ -378,27 +374,22 @@ void test_pkcs11_C_GetFunctionListBadArgs( void )
 void test_pkcs11_C_GetSlotList( void )
 {
     CK_RV xResult = CKR_OK;
-    CK_SLOT_ID_PTR pxSlotId = NULL;
-    CK_ULONG xSlotCount = 0;
+    CK_SLOT_ID xSlotId = 0;
+    CK_ULONG xSlotCount = 1;
     CK_BBOOL xTokenPresent = CK_TRUE;
 
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
 
-    xResult = C_GetSlotList( xTokenPresent, pxSlotId, &xSlotCount );
+    xResult = C_GetSlotList( xTokenPresent, &xSlotId, &xSlotCount );
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
     TEST_ASSERT_EQUAL( 1, xSlotCount );
 
-    pxSlotId = pvPkcs11MallocCb( sizeof( CK_SLOT_ID ) * xSlotCount, 1 );
-    memset( pxSlotId, 0, sizeof( CK_SLOT_ID ) * 1 );
-
-    xResult = C_GetSlotList( xTokenPresent, pxSlotId, &xSlotCount );
+    xResult = C_GetSlotList( xTokenPresent, &xSlotId, &xSlotCount );
 
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
     TEST_ASSERT_EQUAL( 1, xSlotCount );
-    TEST_ASSERT_EQUAL( 1, pxSlotId[ 0 ] );
-
-    vPkcs11FreeCb( pxSlotId, 1 );
+    TEST_ASSERT_EQUAL( 1, xSlotId );
 
     xResult = prvUninitializePkcs11();
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
@@ -414,7 +405,7 @@ void test_pkcs11_C_GetSlotListNullSlot( void )
     CK_ULONG xSlotCount = 0;
     CK_BBOOL xTokenPresent = CK_TRUE;
 
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
 
     xResult = C_GetSlotList( xTokenPresent, NULL, &xSlotCount );
@@ -436,7 +427,7 @@ void test_pkcs11_C_GetSlotListNullCount( void )
     CK_RV xResult = CKR_OK;
     CK_BBOOL xTokenPresent = CK_TRUE;
 
-    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    xResult = prvInitializePkcs11();
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
 
     xResult = C_GetSlotList( xTokenPresent, NULL, NULL );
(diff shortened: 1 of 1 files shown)


libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -36,18 +37,90 @@
...
+/* ============================  GLOBAL VARIABLES =========================== */
+/* Length parameters for importing EC private keys. */
+#define EC_PARAMS_LENGTH    10
+#define EC_D_LENGTH         32
+
+#define EC_PRIV_KEY_INITIALIZER                                                    \
+    {                                                                              \
+        { CKA_CLASS, &xPrivateKeyClass, sizeof( CK_OBJECT_CLASS ) },               \
+        { CKA_KEY_TYPE, &xPrivateKeyType, sizeof( CK_KEY_TYPE ) },                 \
+        { CKA_LABEL, pucLabel, ( CK_ULONG ) strlen( ( const char * ) pucLabel ) }, \
```
dcgaws, line 69: If this is used with a variable declaration, then I think that the strlen causes a compiler warning. If so, can you please refactor?

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -36,18 +37,90 @@
...
+#define PRIME_2_LENGTH        128
+#define EXPONENT_1_LENGTH     128
+#define EXPONENT_2_LENGTH     128
+#define COEFFICIENT_LENGTH    128
+
+#define RSA_PRIV_KEY_INITIALIZER                                                   \
+    {                                                                              \
+        { CKA_CLASS, &xPrivateKeyClass, sizeof( CK_OBJECT_CLASS ) },               \
+        { CKA_KEY_TYPE, &xPrivateKeyType, sizeof( CK_KEY_TYPE ) },                 \
+        { CKA_LABEL, pucLabel, ( CK_ULONG ) strlen( ( const char * ) pucLabel ) }, \
```
dcgaws, line 91: Ditto.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -36,18 +37,90 @@
...
+        { CKA_TOKEN, &xTrue, sizeof( CK_BBOOL ) },                                 \
+        { CKA_SIGN, &xTrue, sizeof( CK_BBOOL ) },                                  \
+        { CKA_MODULUS, pxRsaParams->modulus + 1, MODULUS_LENGTH },                 \
```
dcgaws, line 94: This initialization is based on the assumption that the mbedTLS MPI encoding always starts with a length byte, correct? But isn't the MPI encoding stated to be opaque? I.e. it could, in theory, change.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -96,32 +169,94 @@ int suiteTearDown( int numFailures )
...
  */
-void test_pkcs11_C_Initialize( void )
+static CK_RV prvInitializePkcs11( SemaphoreHandle_t pxPKCS11Mutex )
 {
     CK_RV xResult = CKR_OK;
 
-    /* Hack: Giving out pointer to xResult when creating a mutex. This unit testcase
-     * will never use the actual mutex. */
-    xQueueCreateMutex_IgnoreAndReturn( ( SemaphoreHandle_t ) &xResult );
+    xQueueCreateMutex_IgnoreAndReturn( pxPKCS11Mutex );
```
dcgaws, line 182: If the test paths will never use the mutex, why create it?

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -96,32 +169,94 @@ int suiteTearDown( int numFailures )
...
     CRYPTO_Init_Ignore();
     mbedtls_entropy_init_Ignore();
     mbedtls_ctr_drbg_init_Ignore();
     mbedtls_ctr_drbg_seed_IgnoreAndReturn( 0 );
     xResult = C_Initialize( NULL );
 
-    TEST_ASSERT_EQUAL( CKR_OK, xResult );
```
dcgaws, line 118: Shouldn't the test check every API call?

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+    xResult = C_GetSlotList( xTokenPresent, pxSlotId, &xSlotCount );
     TEST_ASSERT_EQUAL( CKR_OK, xResult );
+    TEST_ASSERT_EQUAL( 1, xSlotCount );
 
-    mbedtls_entropy_free_CMockIgnore();
-    mbedtls_ctr_drbg_free_CMockIgnore();
-    vQueueDelete_CMockIgnore();
-    xResult = C_Finalize( NULL );
+    pxSlotId = pvPkcs11MallocCb( sizeof( CK_SLOT_ID ) * xSlotCount, 1 );
+    memset( pxSlotId, 0, sizeof( CK_SLOT_ID ) * 1 );
```
dcgaws, line 393: Why is the malloc done in such a way that there could be more than one slot, but the memset isn't?

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+{
+    CK_RV xResult = CKR_OK;
+    CK_SLOT_ID_PTR pxSlotId = NULL;
+    CK_ULONG xSlotCount = 0;
+    CK_BBOOL xTokenPresent = CK_TRUE;
+
+    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
+    TEST_ASSERT_EQUAL( CKR_OK, xResult );
+
+    pxSlotId = pvPkcs11MallocCb( sizeof( CK_SLOT_ID ) * 1, 1 );
```
dcgaws, line 463: If we're going to assume that there's only one slot, why use malloc?

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+    vPkcs11FreeCb( pxSlotId, 1 );
 }
 
+/* ======================  TESTING C_GetTokenInfo  ============================ */
+
 /*!
- * @brief C_Finalize was called with a non-null pointer.
+ * @brief C_GetTokenInfo Happy path.
+ *
+ * @note this will need to be updated if tokens are implemented in this PKCS #11
```
dcgaws, line 505: More accurate might be, if this port needs to start returning token information at some point - which it probably should in order to be compatible with hardware behavior - then this test will need to be updated.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+    TEST_ASSERT_EQUAL( CKR_MECHANISM_INVALID, xResult );
+}
+
+/* ======================  TESTING C_InitToken  ============================ */
 
 /*!
- * @brief C_GetFunctionList happy path.
+ * @brief C_InitToken Happy path.
+ *
+ * @note this will need to be updated if tokens are supported
```
dcgaws, line 606: I think it would be more accurate to say that our port behaves like a fixed token and that it doesn't do anything when C_InitToken is called. The former is unlikely to change, but the latter might.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+
+    xResult = prvUninitializePkcs11();
+    TEST_ASSERT_EQUAL( CKR_OK, xResult );
+}
+/* ======================  TESTING C_Login  ============================ */
+
+/*!
+ * @brief C_Login happy path.
+ *
+ * @note This test will need to be updated if support is added for C_Login
```
dcgaws, line 704: Missing a period.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -171,112 +299,540 @@ void test_pkcs11_C_InitializeInitTwice( void )
...
+    PKCS11_PAL_FindObject_IgnoreAndReturn( 1 );
+    PKCS11_PAL_GetObjectValue_IgnoreAndReturn( CKR_OK );
+    mbedtls_pk_parse_key_IgnoreAndReturn( 0 );
+    PKCS11_PAL_GetObjectValueCleanup_CMockIgnore();
+    pvPortMalloc_Stub( pvPkcs11MallocCb );
+    mbedtls_ecp_keypair_init_CMockIgnore();
+    mbedtls_ecp_group_init_CMockIgnore();
+    mbedtls_ecp_group_load_IgnoreAndReturn( 0 );
+
+    /* current callstack */
```
dcgaws: What does this comment mean?

lundinc2: Sorry this was a tracker for me, forgot to remove it.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Hmm I did it this way so I could pass in a pointer but returning anything other than NULL should be sufficient, I will update this based on your suggestion.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: I compiled this on GCC with -Wall -Wpedantic -Wextra and confirmed no warnings  👍

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Good point, I will re-add this assert

lundinc2: I am leaving it as is. I forgot I was using this helper function for different branches, so this is not guaranteed to be OK. So the unit test case must always check the return of this helper function.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: I aligned this code with what we are using in the provisioning code. Let me make a ticket to investigate this, I'm unsure if a refactor this is within the scope of this PR as we'll likely want to make the same change across the whole code base at once.

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Fixed this

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -96,32 +169,94 @@ int suiteTearDown( int numFailures )
...
+ *
+ */
+static CK_RV prvUninitializePkcs11()
+{
+    CK_RV xResult = CKR_OK;
 
     mbedtls_entropy_free_CMockIgnore();
     mbedtls_ctr_drbg_free_CMockIgnore();
     vQueueDelete_CMockIgnore();
     xResult = C_Finalize( NULL );
```
alfred2g, line 204: I think it looks cleaner is the expectations, function call, and asserts are grouped together in a single block of code

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -96,32 +169,94 @@ int suiteTearDown( int numFailures )
...
+
+/*!
+ * @brief C_Initialize happy path.
+ *
+ */
+void test_pkcs11_C_Initialize( void )
+{
+    CK_RV xResult = CKR_OK;
+
+    xResult = prvInitializePkcs11( ( SemaphoreHandle_t ) &xResult );
```
alfred2g: Since the main point is to test the initialization stuff, it would be neater to see the logic here (expectations, function calls, and asserts)

The uninitialize below is ok to have the logic in a function not distracting us from the main point of this test function (the test_assert_equal included)

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: I assert after the return from this helper so I can reuse some expectations for different branch paths

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
```diff
@@ -96,32 +169,90 @@ int suiteTearDown( int numFailures )
...
+
+/*!
+ * @brief C_Initialize happy path.
+ *
+ */
+void test_pkcs11_C_Initialize( void )
+{
+    CK_RV xResult = CKR_OK;
+
+    xResult = prvInitializePkcs11();
```
alfred2g, line 252: it would be much clearer what this function is trying to do if the contents of the initialize were right here instead of in a function, since this is the purpose of this function testing initialize

It is ok for the content of the deinitialize  to be in a function since it is only a helper including its assert

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

libraries/abstractions/pkcs11/utest/iot_pkcs11_mbedtls_utest.c
lundinc2: Done

PR #722, opened 2019-06-06
Test update: Removed one wifi test. Moved TCP thread safety tests to the end of the group

xuelix:
Checklist:
----------

- [x ] I have tested my changes. No regression in existing tests.
- [ ] My code is Linted.

PR #1977, opened 2020-05-05
common-io: Refactoring and updates to RTC test cases and top-level header file

osloapps-max:
Description
-----------
* Added TEST_PROTECT sections to test cases
* Removed global handle as the usage of the RTC is always open -> close which means the handle would be closed after the first test had run (often without setting it to NULL again).
* Added "Week day" to the date/time structure for each test so that the whole structure is in a known state when performing the tests.
* Changed semaphore timeouts to not be "eternal".
* Added global variable that can be used to configure which RTC instance to open (test was hard-coded to 0)
* Updated/added expected return value to the top-level API header file based on how the test cases was setup (please feedback on this as the test and API description was not aligned).

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_rtc.c
+++ b/libraries/abstractions/common_io/test/test_iot_rtc.c
@@ -41,26 +41,32 @@
 #include "task.h"
 
 /*-----------------------------------------------------------*/
-#define testIotRTC_DEFAULT_INVALID_DATE      ( 0xFFFF )
-#define testIotRTC_DEFAULT_CURRENT_YEAR      ( 119 )    /* 2019 */
-#define testIotRTC_DEFAULT_CURRENT_MONTH     ( 0 )      /* January*/
-#define testIotRTC_DEFAULT_CURRENT_DAY       ( 1 )      /* 1st of the month */
-#define testIotRTC_DEFAULT_CURRENT_HOUR      ( 0 )      /* 0th hour */
-#define testIotRTC_DEFAULT_CURRENT_MINUTE    ( 0 )      /* 0th minute */
-#define testIotRTC_DEFAULT_CURRENT_SECOND    ( 0 )      /* 0th second */
+#define testIotRTC_DEFAULT_INVALID_DATE        ( 0xFFFF )
+#define testIotRTC_DEFAULT_CURRENT_WEEK_DAY    ( 2 )    /* 1st of January 2019 is a tuesday */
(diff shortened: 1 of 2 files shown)


libraries/abstractions/common_io/test/test_iot_rtc.c
```diff
@@ -712,79 +686,83 @@ TEST( TEST_IOT_RTC, AFQP_IotRtcSetAlarmInvalid )
...
+    TEST_ASSERT_NOT_EQUAL( NULL, xRtcHandle );
 
+    xSetDateTime.ucWday = testIotRTC_DEFAULT_CURRENT_WEEK_DAY;
     xSetDateTime.usYear = testIotRTC_DEFAULT_CURRENT_YEAR;
     xSetDateTime.ucMonth = testIotRTC_DEFAULT_CURRENT_MONTH;
     xSetDateTime.ucDay = testIotRTC_DEFAULT_CURRENT_DAY;
     xSetDateTime.ucHour = testIotRTC_DEFAULT_CURRENT_HOUR;
     xSetDateTime.ucMinute = testIotRTC_DEFAULT_CURRENT_MINUTE;
-    xSetDateTime.ucSecond = testIotRTC_DEFAULT_CURRENT_SECOND;
+    xSetDateTime.ucSecond = testIotRTC_DEFAULT_CURRENT_SECOND + testIotRTC_DEFAULT_ALARM_TIME;
```
vjjunnut, line 698: Why is the DEFAULT_ALARM_TIME added here to the default time being set? #722 may not return error if its the current second.

osloapps-max: The reason is due to the fact that testIotRTC_DEFAULT_CURRENT_SECOND is zero. The original setup would subtract 5 from 0 in a uint8 causing a underflow. That said, it could be a good idea to bump it so that xSetDateTime.ucSecond > testIotRTC_DEFAULT_ALARM_TIME.

You agree with bumping it to say 2 *  testIotRTC_DEFAULT_ALARM_TIME? Alternatively, set testIotRTC_DEFAULT_CURRENT_SECOND > testIotRTC_DEFAULT_ALARM_TIME.

vjjunnut: Yes. Didnt realize that. I agree to either of the approach

osloapps-max: Updated.

PR #1980, opened 2020-05-06
Hygiene/remove gcc warnings from mqtt ggd

aggarw13:
Resolve (ARM) GCC warnings from MQTT and GGD test files

Description
-----------

Checklist:
----------

- [ ] My code is Linted.


libraries/c_sdk/aws/defender/test/system/aws_iot_tests_defender_system.c
```diff
@@ -74,7 +74,7 @@
 #define MAX_ADDRESS_LENGTH                   25
 
 /* Use a big number to represent no event happened in defender. */
-#define NO_EVENT                             10000
+#define NO_EVENT                             ( 1u << ( 8 * sizeof( AwsIotDefenderEventType_t ) - 1 ) )
```
muneebahmed10, line 77: I don't see why this is any better really. Seeing a very specific expression like this makes me think there's a special reason for it being written this way, but the comment implies that any number will do as long as it's more than the number of enums.

aggarw13: That change was unrelated to the scope of the PR (Was present in the diff as I had branched off of PR: #1975). That value has been updated to the `255` literal instead

PR #1984, opened 2020-05-06
PKCS11 Refactor

lundinc2:
Fix MISRA violations in pkcs11_mbetls, impossible paths, dead code. Fix a memory leak when a RSA ctx is not freed. Fix some error logging. Reduce logic complexity.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -807,6 +807,11 @@ CK_RV prvAddObjectToList( CK_OBJECT_HANDLE xPalHandle,
                 }
             }
 
+            if( xResult != CKR_OK )
+            {
+                PKCS11_PRINT( ( "Error cleaning up PAL Handle 2. \r\n, %d", xResult ) );
+            }
+
             xResult = prvDeleteObjectFromList( xAppHandle );
         }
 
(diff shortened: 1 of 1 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1293,6 +1297,12 @@ CK_DECLARE_FUNCTION( CK_RV, C_CloseSession )( CK_SESSION_HANDLE xSession )
             vSemaphoreDelete( pxSession->xVerifyMutex );
         }
 
+        if( NULL != pxSession->pxFindObjectLabel )
+        {
+            vPortFree( pxSession->pxFindObjectLabel );
+            pxSession->pxFindObjectLabel = NULL;
+        }
+
```
alfred2g, line 1305: the free libc free allows freeing of null variables (with no effect of course), not sure about vPortFree implementation 
if it is a requirements, then we don't need to check for != NULL

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -2019,6 +2030,11 @@ CK_RV prvCreatePrivateKey( CK_ATTRIBUTE_PTR pxTemplate,
         vPortFree( pxDerKey );
     }
 
+    if( pxRsaCtx != NULL )
```
alfred2g, line 2033: same

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: Oh interesting, I'll try to remove this

alfred2g: don't forget to create another commit to remove the if (ptr == null) before freeing

PR #1995, opened 2020-05-11
Fix compile warnings in common io and wifi tests

qiutongs:
Description
-----------

Fix compile warnings in common io and wifi tests

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_i2c.c
+++ b/libraries/abstractions/common_io/test/test_iot_i2c.c
@@ -48,7 +48,7 @@
 #define testIotI2C_BAUDRATE_LOW_SPEED     IOT_I2C_STANDARD_MODE_BPS
 #define testIotI2C_DEFAULT_TIMEOUT        500 /**< 500 msec */
 #define testIotI2C_FAST_TIMEOUT           100 /**< 100 msec */
-#define testIotI2C_INVALID_IOCTL_INDEX    UINT32_MAX
+#define testIotI2C_INVALID_IOCTL_INDEX    UINT8_MAX
 #define testIotI2C_HANDLE_NUM             4
 #define testIotI2C_MAX_TIMEOUT            pdMS_TO_TICKS( 10000 )
 #define testIotI2C_MESSAGE_LENGTH         50
@@ -126,7 +126,7 @@ TEST_TEAR_DOWN( TEST_IOT_I2C )
 void prvI2CCallback( IotI2COperationStatus_t xOpStatus,
                      void * pvParam )
 {
-    /* Silence the compiler. */
+    /* Disable unused parameter warning. */
     ( void ) pvParam;
 
     if( xOpStatus == eI2CCompleted )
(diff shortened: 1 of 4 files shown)


libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -126,6 +126,9 @@ TEST_TEAR_DOWN( TEST_IOT_I2C )
 void prvI2CCallback( IotI2COperationStatus_t xOpStatus,
                      void * pvParam )
 {
+    /* Silence the compiler. */
```
aggarw13, line 129: Can you mention something like `Disable unused parameter warning.` instead to clearly mention the warning being silenced

libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -357,7 +360,7 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CSetGetConfigurationFail )
         TEST_ASSERT_EQUAL( IOT_I2C_INVALID_VALUE, lRetVal );
 
         /* i2c ioctl with unsupported request */
-        lRetVal = iot_i2c_ioctl( xI2CHandle, testIotI2C_INVALID_IOCTL_INDEX, &xI2CConfig );
+        lRetVal = iot_i2c_ioctl( xI2CHandle, ( IotI2CIoctlRequest_t ) testIotI2C_INVALID_IOCTL_INDEX, &xI2CConfig );
```
aggarw13, line 363: I notice that `testIotI2C_INVALID_IOCTL_INDEX` is defined as `UINT32_MAX`. Instead of explicitly downcasting it, should it be defined to be a lower value as `UINT8_MAX` instead or any other unsupported value?

libraries/abstractions/common_io/test/test_iot_i2c.c
qiutongs: Sure.

libraries/abstractions/common_io/test/test_iot_i2c.c
```diff
@@ -484,7 +487,7 @@ TEST( TEST_IOT_I2C, AFQP_IotI2CReadAsyncAssisted )
         TEST_ASSERT_EQUAL( IOT_I2C_SUCCESS, lRetVal );
 
         /* Read from i2c device */
-        lRetVal = iot_i2c_read_async( xI2CHandle, &ucReadBuf, sizeof( ucReadBuf ) );
+        lRetVal = iot_i2c_read_async( xI2CHandle, ucReadBuf, sizeof( ucReadBuf ) );
```
aggarw13, line 490: Were these tests passing earlier?

libraries/abstractions/common_io/test/test_iot_i2c.c
qiutongs: I think this is a better idea. Will do.

libraries/abstractions/common_io/test/test_iot_i2c.c
qiutongs: They weren't running. See the test case name is suffixed with "assisted". We haven't support assisted tests yet.

aggarw13: Is it possible to locally verify that these tests are passing?

qiutongs: Yes, but it takes extra work. Since assisted test is not supported for now, let's worry about this later.

cobusve: Since ucReadBuf is an array this would work either way. I agree with Qiutong, this should not hold us up here. The warning this removes is simply that the adress off operator is redundant.

libraries/abstractions/common_io/test/test_iot_i2c.c
cobusve: Yes please "Silence the compiler" is not a good comment

PR #2006, opened 2020-05-13
Adds OTA Support to CY8CKIT_064S0S2_4343W

RaymondNgun:
Adds OTA Support to CY8CKIT_064S0S2_4343W

Description
-----------
Patch set introduces OTA support to the kit. However, the "phone home" feature is absent.

Checklist:
----------

- [ ] My code is Linted.


vendors/cypress/psoc6/psoc6make/make/scripts/sign_script.bash
```diff
@@ -0,0 +1,147 @@
...
+# This file is used by make to create the build commands to sign an OTA image
+#
+# Modify at your peril !
+#
+# The output files will be in the same directory as the .elf, .hex, etc files
+#
+(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required
+set -e
+
+# Arguments
```
gkwicker, line 12: Ideally you should be using command-line options, verifying arguments, and providing help if they aren't correct.  Here's a [good stack overflow post](https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash) on how to do this in `bash`.

vendors/cypress/psoc6/psoc6make/make/scripts/sign_tar.bash
```diff
@@ -0,0 +1,190 @@
...
+#
+# The output files will be in the same directory as the .elf, .hex, etc files
+#
+(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required
+set -e
+
+# Arguments
+# We have a lot
+#
+CY_OUTPUT_PATH=$1
```
gkwicker, line 15: See previous comment

RaymondNgun:
> Why does this PR have changes in `vendors/espressif/esp-idf`?

That's a good question. Removed. Apologies.

vendors/cypress/psoc6/psoc6make/make/scripts/sign_script.bash
RaymondNgun: Thanks for the feedback.  I will submit an internal ticket to enhance this. It won't be done and verified within the time frame needed here however.

PR #2008, opened 2020-05-13
PKCS #11 Refactor

lundinc2:
PKCS #11 Refactor

Description
-----------
* Fixed various MISRA violations.
* Refactored last two remaining functions with a too high GNU Complexity.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -631,7 +631,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
                                 mbedtls_pk_context * pxMbedContext )
 {
     CK_RV xResult = CKR_OK;
-    int lMbedReturn = 0;
+    int32_t lMbedReturn = 0;
     CK_BBOOL xBool;
     mbedtls_rsa_context * pxRsaContext = ( mbedtls_rsa_context * ) pxMbedContext->pk_ctx;
 
@@ -656,7 +656,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_MODULUS ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
                                                    pxAttribute->pValue, pxAttribute->ulValueLen, /* N */
                                                    NULL, 0,                                      /* P */
                                                    NULL, 0,                                      /* Q */
@@ -665,7 +665,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_PUBLIC_EXPONENT ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
                                                    NULL, 0,                                        /* N */
                                                    NULL, 0,                                        /* P */
                                                    NULL, 0,                                        /* Q */
@@ -674,7 +674,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_PRIME_1 ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
                                                    NULL, 0,                                      /* N */
                                                    pxAttribute->pValue, pxAttribute->ulValueLen, /* P */
                                                    NULL, 0,                                      /* Q */
@@ -683,7 +683,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_PRIME_2 ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
                                                    NULL, 0,                                      /* N */
                                                    NULL, 0,                                      /* P */
                                                    pxAttribute->pValue, pxAttribute->ulValueLen, /* Q */
@@ -692,7 +692,7 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_PRIVATE_EXPONENT ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
                                                    NULL, 0,                                      /* N */
                                                    NULL, 0,                                      /* P */
                                                    NULL, 0,                                      /* Q */
@@ -701,15 +701,15 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_EXPONENT_1 ):
-            lMbedReturn |= mbedtls_mpi_read_binary( &pxRsaContext->DP, pxAttribute->pValue, pxAttribute->ulValueLen );
+            lMbedReturn = mbedtls_mpi_read_binary( &pxRsaContext->DP, pxAttribute->pValue, pxAttribute->ulValueLen );
             break;
 
         case ( CKA_EXPONENT_2 ):
-            lMbedReturn |= mbedtls_mpi_read_binary( &pxRsaContext->DQ, pxAttribute->pValue, pxAttribute->ulValueLen );
+            lMbedReturn = mbedtls_mpi_read_binary( &pxRsaContext->DQ, pxAttribute->pValue, pxAttribute->ulValueLen );
             break;
 
         case ( CKA_COEFFICIENT ):
-            lMbedReturn |= mbedtls_mpi_read_binary( &pxRsaContext->QP, pxAttribute->pValue, pxAttribute->ulValueLen );
+            lMbedReturn = mbedtls_mpi_read_binary( &pxRsaContext->QP, pxAttribute->pValue, pxAttribute->ulValueLen );
             break;
 
         default:
@@ -737,7 +737,7 @@ static CK_RV prvEcPrivKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
                                    mbedtls_pk_context * pxMbedContext )
 {
     CK_BBOOL xBool = CK_FALSE;
-    int lMbedReturn = 0;
+    int32_t lMbedReturn = 0;
     CK_RV xResult = CKR_ATTRIBUTE_VALUE_INVALID;
     mbedtls_ecp_keypair * pxKeyPair = ( mbedtls_ecp_keypair * ) pxMbedContext->pk_ctx;
 
@@ -793,7 +793,7 @@ static CK_RV prvEcPubKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
                                   mbedtls_pk_context * pxMbedContext )
 {
     CK_BBOOL xBool = CK_FALSE;
-    int lMbedReturn = 0;
+    int32_t lMbedReturn = 0;
     CK_RV xResult = CKR_ATTRIBUTE_VALUE_INVALID;
     mbedtls_ecp_keypair * pxKeyPair = ( mbedtls_ecp_keypair * ) pxMbedContext->pk_ctx;
 
@@ -899,13 +899,13 @@ static CK_RV prvEcKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
     }
 
     /* private EC key attributes. */
-    if( ( xResult != CKR_OK ) && ( xIsPrivate == CK_TRUE ) )
+    if( ( xResult == CKR_ATTRIBUTE_VALUE_INVALID ) && ( xIsPrivate == CK_TRUE ) )
     {
         xResult = prvEcPrivKeyAttParse( pxAttribute, pxMbedContext );
     }
 
     /* public EC key attributes. */
-    if( ( xResult != CKR_OK ) && ( xIsPrivate == CK_FALSE ) )
+    else if( ( xResult == CKR_ATTRIBUTE_VALUE_INVALID  ) && ( xIsPrivate == CK_FALSE ) )
     {
         xResult = prvEcPubKeyAttParse( pxAttribute, pxMbedContext );
     }
@@ -963,12 +963,12 @@ void prvFindObjectInListByLabel( uint8_t * pcLabel,
  * @param[out] pxLabelLength     Pointer to label length (includes a string null terminator).
  *                               0 if no object found.
  */
-void prvFindObjectInListByHandle( CK_OBJECT_HANDLE xAppHandle,
+static void prvFindObjectInListByHandle( CK_OBJECT_HANDLE xAppHandle,
                                   CK_OBJECT_HANDLE_PTR pxPalHandle,
                                   uint8_t ** ppcLabel,
                                   size_t * pxLabelLength )
 {
-    int lIndex = xAppHandle - 1;
+    uint32_t lIndex = xAppHandle - 1;
 
     *ppcLabel = NULL;
     *pxLabelLength = 0;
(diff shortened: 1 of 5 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -899,13 +894,13 @@ static CK_RV prvEcKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
     }
 
     /* private EC key attributes. */
-    if( ( xResult != CKR_OK ) && ( xIsPrivate == CK_TRUE ) )
+    if( ( xResult == CKR_ATTRIBUTE_VALUE_INVALID ) && ( xIsPrivate == CK_TRUE ) )
```
qiutongs, line 897: Are you intentionally making this logic change?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -899,13 +894,13 @@ static CK_RV prvEcKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
...
     {
         xResult = prvEcPrivKeyAttParse( pxAttribute, pxMbedContext );
     }
 
     /* public EC key attributes. */
-    if( ( xResult != CKR_OK ) && ( xIsPrivate == CK_FALSE ) )
+    else if( ( xResult == CKR_ATTRIBUTE_VALUE_INVALID ) && ( xIsPrivate == CK_FALSE ) )
```
qiutongs, line 903: similar here.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1159,19 +1166,19 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
         uint8_t emptyPubKey[ 6 ] = { 0xa1, 0x04, 0x03, 0x02, 0x00, 0x00 };
         lCompare = memcmp( &pxDerKey[ MAX_LENGTH_KEY - 6 ], emptyPubKey, 6 );
 
-        if( lCompare == 0 )
+        if( ( lCompare == 0 ) && ( ulActualKeyLength >= 6 ) )
         {
             /* Do not write the last 6 bytes to key storage. */
             pxDerKey[ MAX_LENGTH_KEY - lDerKeyLength + 1 ] -= 6;
-            lActualKeyLength -= 6;
```
qiutongs, line 1166: Is it a bug in the old code? There seems no update to lActualKeyLength after initialization.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1242,40 +1247,36 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
...
         }
+        else
+        {
+            xResult = CKR_ATTRIBUTE_VALUE_INVALID;
+        }
 
         if( xResult == CKR_OK )
         {
-            if( 0 == memcmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabelLength ) )
+            if( 0 == strcmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS ) )
```
qiutongs, line 1257: Is it less safe if change to strcmp? Is it better to explicitly specify the length?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1242,40 +1247,36 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
...
             }
 
             xResult = prvDeleteObjectFromList( xAppHandle );
         }
 
-        if( xFreeMemory == CK_TRUE )
-        {
-            PKCS11_PAL_GetObjectValueCleanup( pxObject, ulObjectLength );
-        }
+        PKCS11_PAL_GetObjectValueCleanup( pxObject, ulObjectLength );
```
qiutongs, line 1279: So it is safe to always call cleanup even PKCS11_PAL_GetObjectValue returns error in line 1219?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: yes, the issue was it would override error messages from an earlier failure, this default case indicates that nothing was caught by the above switch statement

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: this logic needs to be examined thoroughly, i didn't get a regression anywhere when i removed it, but we will need to dig deep into the mbedTLS APIs and how the EC keys are created / used.

lundinc2: Also i kept the logic here as it was originally, as i need to deep dive

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: it should be OK, the second operator is a constant that is always NULL t terminated. MISRA complains when you use mem API with NULL terminated strings

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: It should always be safe, as the handle will be invalid if the other calls were unsuccessful.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
qiutongs: ok

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
qiutongs: I concern the case when the length of pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS is larger than xLabelLength. You might access invalid memory.

Let's use strncmp:
```
strcmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabelLength )
```

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
qiutongs: ok

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -656,60 +651,60 @@ static CK_RV prvRsaKeyAttParse( CK_ATTRIBUTE_PTR pxAttribute,
             break;
 
         case ( CKA_MODULUS ):
-            lMbedReturn |= mbedtls_rsa_import_raw( pxRsaContext,
-                                                   pxAttribute->pValue, pxAttribute->ulValueLen, /* N */
-                                                   NULL, 0,                                      /* P */
-                                                   NULL, 0,                                      /* Q */
-                                                   NULL, 0,                                      /* D */
-                                                   NULL, 0 );                                    /* E */
+            lMbedReturn = mbedtls_rsa_import_raw( pxRsaContext,
```
dan4thewin, line 654: Can you describe the impact of these changes ( `|=` vs `=` )?

lundinc2: Sure! Originally all of these functions were called in a loop without checking their return value. The intention of the code was if one call failed, it wouldn't get overwritten from a call after it.

Now it is refactored to check the return value in each loop iteration. Also it seems a bit misleading to have the return value's get OR'd together, as it makes it seems the return value are related, but they are independent of each other.

Lastly, MISRA does not allow for bitwise operations on signed integers.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
dcgaws: This approach seems risky from the start, though, because we're attempting to parse the data as a certain format without explicitly checking the attribute type. Safer would be to move these blocks into the switch/case above. That would also make this function easier to review and maintain, since it makes for a more typical structure.

As part of that, xResult could for example be initialized as CKR_OK. Then if you fall into the default switch/case, set it to ATTR_VALUE_INVALID.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -948,7 +943,7 @@ void prvFindObjectInListByLabel( uint8_t * pcLabel,
         if( 0 == memcmp( pcLabel, xP11Context.xObjectList.xObjects[ ucIndex ].xLabel, xLabelLength ) )
         {
             *pxPalHandle = xP11Context.xObjectList.xObjects[ ucIndex ].xHandle;
-            *pxAppHandle = ucIndex + 1; /* Zero is not a valid handle, so let's offset by 1. */
+            *pxAppHandle = ucIndex + 1UL; /* Zero is not a valid handle, so let's offset by 1. */
```
dcgaws, line 946: Converting our internal storage pointer to and from the app handle is something that I've typically seen handled by static helper functions. That makes it easier to maintain and modify down the road, if necessary.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
dcgaws: In other words, by adding an empty public key to a "naked" private key, mbed is making it so that the resulting private key format is inconsistent with the PKCS #11 standard. We're therefore omitting the last six bytes in that case (i.e. because we're using mbed to encode the key, but we need our storage and I/O values to be conformant with P#11). However, this situation probably only arises when we're doing round-trip testing between keys generated by different modules.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1214,7 +1220,6 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
 
             if( xResult == CKR_OK )
             {
-                xFreeMemory = CK_TRUE;
```
dcgaws, line 1217: We're still unconditionally calling vPortFree below, though, even in the case where the malloc failed.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1242,40 +1247,36 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
...
         }
+        else
+        {
+            xResult = CKR_ATTRIBUTE_VALUE_INVALID;
+        }
 
         if( xResult == CKR_OK )
         {
-            if( 0 == memcmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabelLength ) )
+            if( 0 == strncmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabel.ulValueLen ) )
```
dcgaws, line 1257: Isn't strncmp less efficient than memcmp in this case?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1926,27 +1928,31 @@ void prvGetLabel( CK_ATTRIBUTE_PTR * ppxLabel,
...
     }
-    else if( 0 == memcmp( pxLabel->pValue, pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS, pxLabel->ulValueLen ) )
+    else if( 0 == strncmp( pxLabel->pValue, pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS, pxLabel->ulValueLen ) )
     {
         *pxPalHandle = PKCS11_PAL_FindObject( ( uint8_t * ) pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, ( uint8_t ) pxLabel->ulValueLen );
         xIsPrivate = CK_FALSE;
     }
+    else
+    {
+        /* Unknown label passed to function */
```
dcgaws, line 1954: Shouldn't an error condition be set here then?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1242,40 +1247,36 @@ static CK_RV prvSaveDerKeyToPal( mbedtls_pk_context * pxMbedContext,
...
+            xResult = CKR_ATTRIBUTE_VALUE_INVALID;
+        }
 
         if( xResult == CKR_OK )
         {
-            if( 0 == memcmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabelLength ) )
+            if( 0 == strncmp( xLabel.pValue, pkcs11configLABEL_DEVICE_PRIVATE_KEY_FOR_TLS, xLabel.ulValueLen ) )
             {
                 prvFindObjectInListByLabel( ( uint8_t * ) pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS, strlen( ( char * ) pkcs11configLABEL_DEVICE_PUBLIC_KEY_FOR_TLS ), &xPalHandle, &xAppHandle2 );
-
```
dcgaws, line 1251: This is hard to review in GitHub with the wide columns. Is it over 80?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: Alfred pointed out that free handles the case when it is passed  a NULL pointer, so there is no reason to check if malloc failed when freeing

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: MISRA complains about NULL terminated strings being operated on by the mem family of functions

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: Okay will follow up with a PR to move this to static helper function. Cutting a ticket to track this effort

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: Yes, I'll make a follow up PR to condense all lines to 80 chars

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: The pxPalHandle is set to INVALID when initialized, so we will return an error

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3718,8 +3728,13 @@ CK_DECLARE_FUNCTION( CK_RV, C_Verify )( CK_SESSION_HANDLE xSession,
             mbedtls_mpi xS;
             mbedtls_mpi_init( &xR );
             mbedtls_mpi_init( &xS );
+
             lMbedTLSResult = mbedtls_mpi_read_binary( &xR, &pucSignature[ 0 ], 32 );
-            lMbedTLSResult |= mbedtls_mpi_read_binary( &xS, &pucSignature[ 32 ], 32 );
+
+            if( lMbedTLSResult == 0 )
+            {
+                lMbedTLSResult = mbedtls_mpi_read_binary( &xS, &pucSignature[ 32 ], 32 );
```
dcgaws, line 3736: It would be great to stop using the hard-coded key length values for ECDSA as part of this refactoring.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: I had to split this function up to due to the GNU complexity requirement. The switch statement contains too much logic if its all in one. I'm more than happy to combine all the switches into one, as i agree it is easier to read, but i will need to get an exception for the GNU complexity requirement.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: Will update this

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: I have refactored this switch statement to be more clear

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c

PR #2034, opened 2020-05-23
Go Echo Server TLS 

lundinc2:
Only use TLS cipher suite's supported by AWS IoT Core

Description
-----------
Explicitly only allow the use of the cipher suite's AWS IoT Core uses.

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/tools/echo_server/echo_server.go
+++ b/tools/echo_server/echo_server.go
@@ -72,6 +72,17 @@ func secureEcho(certPath string, keyPath string, port string, verbose bool) {
 		MinVersion: tls.VersionTLS12,
 		RootCAs:    serverCAPool,
 		ClientAuth: tls.RequireAnyClientCert,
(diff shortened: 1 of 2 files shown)


tools/echo_server/echo_server.go
```diff
@@ -72,6 +72,17 @@ func secureEcho(certPath string, keyPath string, port string, verbose bool) {
 		MinVersion: tls.VersionTLS12,
 		RootCAs:    serverCAPool,
 		ClientAuth: tls.RequireAnyClientCert,
+		// Cipher suites supported by AWS IoT Core. Note this is the intersection of the set
+		// of cipher suites supported by GO and by AWS IoT Core.
```
yuhui-zheng, line 76: Would this be more accurate? -- "cipher suites supported by **OpenSSL** and by AWS IoT Core"

ps: https://www.openssl.org/docs/man1.1.0/man1/ciphers.html this is what I'm referring to.

PR #2035, opened 2020-05-26
Update s3 bucket example pre-signed link in HTTP demo docs

sarenameas:
The s3 bucket and endpoint are automatically generated from the python script "presigned_url_gen.py".
The example URL in the documentation shows what could be generated. What is generated will be copied by users to their iot_config.h so it is safe to change to "aws-s3-endpoint".

Checklist:
----------

- [ ] My code is Linted.


demos/https/README.md
```diff
@@ -25,17 +25,17 @@
 ### Usage
 1. Run the presigned_url_gen.py with your s3 **bucket-name** and s3 object key **object-key**.  
    ```
-   python presigned_urls_gen.py --bucket my-bucket --key object-key
+   python presigned_urls_gen.py --bucket my_bucket --key object-key
```
gkwicker, line 28: This needs to be something syntactically incorrect.  Anytime you provide shell commands, assume that someone will copy&paste them into a shell, e.g.:

```sh
    python presigned_urls_gen.py --bucket <YOUR BUCKET NAME> --key <YOUR OBJECT KEY> 
```

sarenameas: Thanks

AniruddhaKanhere:
@sarenameas, A check (`test_pkcs11_C_CloseSessionUninit`) is failing. Is it of any concern?
Just asking since you have 2 approvals and this might get merged 'as is' since the failing check is not necessary :)

PR #2042, opened 2020-05-27
Update BLE HAL doxygen

doujiahu:
Update BLE HAL doxygen

Description
-----------
Update BLE HAL doxygen
Remove unnecessary extern in front of function prototype

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/ble_hal/include/bt_hal_manager.h
+++ b/libraries/abstractions/ble_hal/include/bt_hal_manager.h
@@ -861,7 +861,7 @@ typedef struct
     BTStatus_t ( * pxGetStackFeaturesSupport )( uint32_t * pulFeatureMask );
 } BTInterface_t;
 
-extern const BTInterface_t * BTGetBluetoothInterface();
+const BTInterface_t * BTGetBluetoothInterface();
 
 #endif /* _BT_HAL_MANAGER_H_ */
 /** @} */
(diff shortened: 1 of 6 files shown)


libraries/abstractions/ble_hal/include/bt_hal_manager.h
```diff
@@ -59,16 +59,10 @@
...
- * @brief  Help functions to convert version to string.
- */
-#define STR_HELPER( x, y, z )    # x "." # y "." # z
-#define STR( x, y, z )           STR_HELPER( x, y, z )
-
 /**
  * @brief  Stringified version number.
  */
-#define btSTRING_VERSION                        STR( btMAJOR_VERSION, btMINOR_VERSION, btPATCH_VERSION )
+#define btSTRING_VERSION                        # btMAJOR_VERSION "." # btMINOR_VERSION "." # btPATCH_VERSION
```
ravibhagavandas, line 65: Can you confirm this works ? The stringification of macro requires it to be passed as an argument and requires one level of indirection as mentioned here:
https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

doujiahu: Added back macros and appended "bt" in front of macro to address ACS team concerns

PR #2046, opened 2020-05-28
Ota/Add support for allowing downgrade

pvyawaha:
This change adds support for allowing downgrade and same version update using a new OTA configuration.

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -1825,20 +1825,11 @@ static OTA_Err_t prvValidateUpdateVersion( OTA_FileContext_t * C )
         /* Check if version reported is the same as the running version. */
         else if( C->ulUpdaterVersion == xAppFirmwareVersion.u.ulVersion32 )
         {
-            if( ( xOTA_Agent.pcClientTokenFromJob == NULL ) ||
-                ( strtoul( ( const char * ) xOTA_Agent.pcClientTokenFromJob, NULL, 0 ) == 0U ) )
-            /*lint !e9007 We don't provide a modifiable variable to strtoul.*/
-            {
-                /* The version is the same so either we're not actually the new firmware or
-                 * someone messed up and sent firmware with the same version. In either case,
-                 * this is a failure of the OTA update so reject the job.
-                 */
-                OTA_LOG_L1( "[%s] We rebooted and the version is still the same.\r\n", OTA_METHOD_NAME );
-            }
-            else
-            {
-                OTA_LOG_L1( "[%s] Invalid client token. Device must be rebooted first.\r\n", OTA_METHOD_NAME );
-            }
+            /* The version is the same so either we're not actually the new firmware or
+             * someone messed up and sent firmware with the same version. In either case,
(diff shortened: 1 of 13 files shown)


libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1796,9 +1796,65 @@ static DocParseErr_t prvInitDocModel( JSON_DocModel_t * pxDocModel,
...
+                /* The version is the same so either we're not actually the new firmware or
+                 * someone messed up and sent firmware with the same version. In either case,
+                 * this is a failure of the OTA update so reject the job.
+                 */
+                OTA_LOG_L1( "[%s] We rebooted and the version is still the same.\r\n", OTA_METHOD_NAME );
+            }
+            else
+            {
+                OTA_LOG_L1( "[%s] Invalid client token. Device must be rebooted first.\r\n", OTA_METHOD_NAME );
+            }
```
dan4thewin, line 1841: This doesn't make sense to me.  Why is the validity of the client token checked after checking that the OTA version  is the same as the running version?  Also, strtoul returns 0 when it can't find a number, not just when the number is zero.

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1905,71 +1962,43 @@ static OTA_FileContext_t * prvParseJobDoc( const char * pcJSON,
...
             if( C->xIsInSelfTest == ( bool_t ) pdTRUE )
             {
                 OTA_LOG_L1( "[%s] In self test mode.\r\n", OTA_METHOD_NAME );
 
-                /* Only check for versions if the target is self */
-                if( xOTA_Agent.ulServerFileID == 0 )
+                /* Validate version of the update received.*/
+                xErrVersionCheck = prvValidateUpdateVersion( C );
+
+                if( otaconfigAllowDowngrade || ( xErrVersionCheck == kOTA_Err_None ) )
```
ravibhagavandas, line 1983: Does this bypass all checks (including invalid image) if otaconfigAllowDowngrade  is set to true? Shouldn't it still check for invalid client token?

pvyawaha: We are only bypassing the version check if the config option is enabled ( we will keep it disabled by default) . This does not bypass any other checks and still logs if downgrade and same version update is performed.

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
pvyawaha: Removed redundant check in https://github.com/aws/amazon-freertos/commit/f8bd5081136e1c5e72f9f55f7f185cb52da305ee

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
ravibhagavandas: Prasad will add a test that a job document with invalid client token will be rejected in both modes (downgrade allowed and not allowed).

PR #2015, opened 2020-05-15
FreeRTOS+TCP after merging with the multi branch v2

htibosch:
FreeRTOS+TCP after merging with the multi branch v2
-----------

Description
-----------
In the last few weeks, we have been preparing a big [PR #1918](https://github.com/aws/amazon-freertos/pull/1918), with the same name as this PR.

It is now ready for reviewing and to get merged with master.
We created a new PR in order to avoid merging conflicts.

There are two groups of files in this PR:

- Sources in libraries/freertos_plus/standard/freertos_plus_tcp :

In some aspects, the /multi version of FreeRTOS+TCP was running ahead of the /multi version.
We merged the current FreeRTOS+TCP /single with the /multi release as much as possible.
In a following stage, /multi will replace /single, and add the possibility to use multiple interfaces. And optionally, IPv6 can be enabled, running in parallel with IPv4.

The sources have passed the checks for MISRA by both Coverity and PC-lint. I places a file called freertos_plus_tcp/coverity_misra.config that was used during the checks.

Beside that, there are many cosmetic changed, e.g. use the ux prefix for all unsigned numeric variables.

- Sources in tools/cbmc :

Thanks to Mark Tuttle and Aniruddha Kanhere, the CBMC proofs are adapted to the new release and they are also extended.

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [x] My code is Linted.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2048, opened 2020-05-29
DNS lookup: warn if ipconfigDNS_CACHE_NAME_LENGTH is too small

htibosch:
Description
-----------
Until recently, when the size of a DNS cache entry was too small, the name would be silently truncated. This could lead to confusion if similar URL's are being used.

Since PR #2015,  we changed this: the name must fit entirely, if not it will be dropped.

In this PR, the length of the URL will be checked if DNS caching is used. And if the URL is too long to be stored, a message will be printed, and the lookup is cancelled.
This will be the message:
~~~
    "prvPrepareLookup: name is too long ( 24 > 16 )"
~~~
There are 9 projects in which `ipconfigDNS_CACHE_NAME_LENGTH` is defined as 16. The aws servers have a name that is much longer, close to 50 characters. I propose to increase the length to 81:
~~~c
    #define ipconfigDNS_CACHE_NAME_LENGTH              ( 81 )
~~~
and to decrease the number of entries, from 4 to 2
~~~c
    #define ipconfigDNS_CACHE_ENTRIES                  ( 2 )
~~~
in order to save some RAM.

Checklist:
----------

- [x] My code is Linted.


vendors/infineon/boards/xmc4800_iotkit/aws_demos/config_files/FreeRTOSIPConfig.h
```diff
@@ -93,8 +93,8 @@ extern void vLoggingPrintf( const char * pcFormatString,
  */
 #define ipconfigUSE_DNS_CACHE                      ( 1 )
 #define ipconfigDNS_CACHE_ADDRESSES_PER_ENTRY      ( 6 )
-#define ipconfigDNS_CACHE_NAME_LENGTH              ( 16 )
-#define ipconfigDNS_CACHE_ENTRIES                  ( 4 )
+#define ipconfigDNS_CACHE_NAME_LENGTH              ( 81 )
```
AniruddhaKanhere, line 96: Why was 81 chosen for the name length?

htibosch: That is completely arbitrary. Add a comment like what for example?

vendors/infineon/boards/xmc4800_plus_optiga_trust_x/aws_demos/config_files/FreeRTOSIPConfig.h
```diff
@@ -93,8 +93,8 @@ extern void vLoggingPrintf( const char * pcFormatString,
  */
 #define ipconfigUSE_DNS_CACHE                      ( 1 )
 #define ipconfigDNS_CACHE_ADDRESSES_PER_ENTRY      ( 6 )
-#define ipconfigDNS_CACHE_NAME_LENGTH              ( 16 )
-#define ipconfigDNS_CACHE_ENTRIES                  ( 4 )
+#define ipconfigDNS_CACHE_NAME_LENGTH              ( 81 )
+#define ipconfigDNS_CACHE_ENTRIES                  ( 2 )
```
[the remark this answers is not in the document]
htibosch: I grep'd the word `ipconfigDNS_CACHE_NAME_LENGTH` in all occurrences of `FreeRTOSIPConfig.h`. A few of them use the maximum  value. All others have a value of 16. I have changed them to 81.

vendors/infineon/boards/xmc4800_iotkit/aws_demos/config_files/FreeRTOSIPConfig.h
AniruddhaKanhere: A number which is an exponential power of 2 (2,4,8,16,32 etc) or power of 2-1 (Something like 1,3,7,15,31 etc) seems okay when I see it as a constant somewhere.
But `81` seems something which is well thought out and must have an explanation for being there since it is so odd of a number. Therefore, I was looking for an explanation. 
But that just might be how I think :)

vendors/infineon/boards/xmc4800_plus_optiga_trust_x/aws_demos/config_files/FreeRTOSIPConfig.h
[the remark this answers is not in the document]
AniruddhaKanhere: Why not make it same across all the boards? Either the maximum value or `81`.
Thinking from a consumer's perspective, one might wonder "why we are using different values across different boards. Is there a reason for it?"

vendors/infineon/boards/xmc4800_iotkit/aws_demos/config_files/FreeRTOSIPConfig.h
htibosch: Haha, 80 was the width in characters of the earliest text terminals. Plus a zero-byte makes 81. Later on, the lines became 128 bytes long.
The AWS IoT server that I use in my programs has a name of 46 characters. Making `ipconfigDNS_CACHE_NAME_LENGTH` equal to 64+1 would work for me as well, for now.

With this PR at least there will be a printed warning if the name doesn't fit.

vendors/infineon/boards/xmc4800_plus_optiga_trust_x/aws_demos/config_files/FreeRTOSIPConfig.h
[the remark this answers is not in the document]
htibosch: If we put all occurrences of `ipconfigDNS_CACHE_NAME_LENGTH` to the maximum of 254, some project maintainers will complain about a high memory usage. Hence I already lowered the number of DNS cache entries ( `ipconfigDNS_CACHE_ENTRIES` ) from 4 to 2.

> why we are using different values across different boards.
> Is there a reason for it?"

There are many minor differences among the `FreeRTOSIPConfig.h` files. I'd rather not touch them too much in order not to break any projects.__

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -532,8 +532,38 @@ TickType_t uxReadTimeOut_ticks = ipconfigDNS_RECEIVE_BLOCK_TIME_TICKS;
  as gethostbyname() may be called from different threads */
 BaseType_t xHasRandom = pdFALSE;
 TickType_t uxIdentifier = 0U;
+#if( ipconfigUSE_DNS_CACHE != 0 )
+	BaseType_t xLengthOk;
```
gkwicker, line 536: Defensive programming opportunity: initialize status variables to error values, e.g.:

```c
BaseType_t xLengthOk = pdFALSE;
```

...as a bonus, this improves readability as lines 555 & 558-561 can be deleted.

Krall-12: Hey it’s clear that you use that and I understand, this would be a great investment for you!

htibosch: In current code, `xLengthOk` will be assigned only once. But you are right that it looks much better with less statements. Thanks

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -552,13 +552,8 @@ TickType_t uxIdentifier = 0U;
 				FreeRTOS_printf( ( "prvPrepareLookup: name is too long ( %lu > %lu )\n",
 								   ( unsigned long ) xLength,
 								   ( unsigned long ) ipconfigDNS_CACHE_NAME_LENGTH ) );
-				xLengthOk = pdFALSE;
 			}
 		}
-		else
-		{
-			xLengthOk = pdFALSE;
-		}
```
AniruddhaKanhere: I think we should keep the `else` statement. It can be empty, but still we should keep it. (Due to MISRA requirement of having and `else` for every `if`).
```
else
{
    /* Do Nothing. */
}
```

htibosch: I haven't seen that MISRA rule yet. What I did see is the following, both from PC-Lint and Coverity:
~~~c
if( count > max_count )
{
	count  = max_count;
}
else if( count < min_count )
{
	count  = min_count;
}
/* an else clause is expected here. */
~~~
error 9013: (Note -- no 'else' at end of 'if ... else if' chain [MISRA 2012 Rule 15.7, required])

What I did see in the FreeRTOS kernel is this:
~~~c
    if( pcName[ x ] == ( char ) 0x00 )
    {
        break;
    }
    else
    {
        mtCOVERAGE_TEST_MARKER();
    }
~~~

htibosch:
This PR will be replaced by a new clean: #2207 
Therefore closing this one.
Hein

PR #2050, opened 2020-05-29
PKCS #11 Refactor

lundinc2:
PKCS #11 Refactor
Description
-----------
This fix refactors session handles to be managed by a static array, instead of casting pointer to 32 bit integers. This resolves MISRA
violation 11.4.

Along with this, the semaphore's used by this file are converted to being allocated statically, further reducing the heap impact of this library.

Checklist:
----------

- [x ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -287,7 +287,7 @@ static P11Session_t * prvSessionPointerFromHandle( CK_SESSION_HANDLE xSession )
 {
     P11Session_t * pxSession = NULL;
 
-    if( ( xSession >= 1 ) && (xSession <= pkcs11configMAX_SESSIONS ) )
+    if( ( xSession >= 1 ) && ( xSession <= pkcs11configMAX_SESSIONS ) )
     {
         /* Decrement by 1, invalid handles in PKCS #11 are defined to be 0. */
         pxSession = &pxP11Sessions[ xSession - 1 ];
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -283,15 +283,14 @@ static CK_RV prvCheckValidSessionAndModule( const P11Session_t * pxSession )
 /**
  * @brief Maps an opaque caller session handle into its internal state structure.
  */
-static P11Session_t * prvSessionPointerFromHandle( const CK_SESSION_HANDLE xSession )
+static P11Session_t * prvSessionPointerFromHandle( CK_SESSION_HANDLE xSession )
 {
     P11Session_t * pxSession = NULL;
-    CK_SESSION_HANDLE xSessionHandle = xSession - 1;
 
-    if( xSessionHandle < pkcs11configMAX_SESSIONS )
+    if( ( xSession >= 1 ) && (xSession <= pkcs11configMAX_SESSIONS ) )
     {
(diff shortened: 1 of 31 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -271,11 +285,16 @@ static CK_RV prvCheckValidSessionAndModule( const P11Session_t * pxSession )
  */
 static P11Session_t * prvSessionPointerFromHandle( const CK_SESSION_HANDLE xSession )
 {
-    /* PKCS #11 uses opaque integer handles to manage sessions.
-     * It is necessary to cast this to a pointer in order to
-     * retrieve the session. */
-    /* coverity[misra_c_2012_rule_11_4_violation] */
-    return ( P11Session_t * ) xSession;
+    P11Session_t * pxSession = NULL;
+    CK_SESSION_HANDLE xSessionHandle = xSession - 1;
```
dan4thewin, line 289: What if xSession is 0?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -1794,9 +1836,19 @@ CK_DECLARE_FUNCTION( CK_RV, C_OpenSession )( CK_SLOT_ID slotID,
 CK_DECLARE_FUNCTION( CK_RV, C_CloseSession )( CK_SESSION_HANDLE hSession )
 {
     P11Session_t * pxSession = prvSessionPointerFromHandle( hSession );
-    CK_RV xResult = prvCheckValidSessionAndModule( pxSession );
+    CK_RV xResult = CKR_OK;
```
dan4thewin, line 1839: Does removing this check change the behavior?

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: The underlying type is an unsigned long int, so it will overflow and be too big of a number to pass the if( xSessionHandle < pkcs11configMAX_SESSIONS ) statment

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
lundinc2: it's slightly different as I added a check to see if the session was opened or not, this wasn't necessary when it was added dynamically, as if the memory was valid we knew it was opened. The function has almost the same logic, except it checks to see if the session was opened before clearing the memory

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
dan4thewin: I wondered if that was the case.  Playing devil's advocate, the type could change to an unsigned char, and pkcs11configMAX_SESSIONS could be set similarly...  I'd really like to see the code make the intention explicit.

lundinc2: The type for CK_SESSION_HANDLE is defined in the PKCS #11 specification. I'm not sure I understand the second part, would you like to see a check that xSession is > 1?

dan4thewin: I guess I think the careful way to proceed is not to rely upon wrap-around, i.e., explicitly check before subtracting.  Sorry that my contrived scenario does not evoke the desired dread...

lundinc2: I have updated this logic based on your feedback, now it uses a more explicit range of valid session handles

PR #2055, opened 2020-05-29
Revert removal of sending second eI2CSendNoStopFlag iot_i2c_ioctl call

wsheet:
* The there are two eI2CSendNoStopFlag calls, one for the write and one for the read
  both are needed.

Description
-----------

Checklist:
----------

- [ x ] My code is Linted.


alfred2g:
I am afraid you will need to run uncrustify locally then push again

```
FAIL: libraries/abstractions/common_io/test/test_iot_i2c.c (File size changed from 52551 to 52637)
```

PR #2059, opened 2020-05-31
Fixes to test_iot_gpio.c

wsheet:
Moved global semaphore creation to TEST_SETUP().
Added code to delete semaphore in TEST_TEAR_DOWN().
Fixed TEST_ASSERT_GREATER_THAN() in AFQP_IotGpioSpeed.

Description
-----------

Checklist:
----------

- [ x ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_gpio.c
+++ b/libraries/abstractions/common_io/test/test_iot_gpio.c
@@ -97,6 +97,8 @@ TEST_GROUP( TEST_IOT_GPIO );
  */
 TEST_SETUP( TEST_IOT_GPIO )
 {
+    xtestIotGpioSemaphore = xSemaphoreCreateBinary();
+    TEST_ASSERT_NOT_EQUAL( NULL, xtestIotGpioSemaphore );
 }
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_gpio.c
```diff
@@ -97,6 +97,8 @@ TEST_GROUP( TEST_IOT_GPIO );
  */
 TEST_SETUP( TEST_IOT_GPIO )
 {
+    xtestIotGpioSemaphore = xSemaphoreCreateBinary();
```
qiutongs, line 100: TEST_SETUP will be called before each test case. So you should delete the semaphore in TEST_TEAR_DOWN as well.

libraries/abstractions/common_io/test/test_iot_gpio.c
```diff
@@ -109,6 +111,12 @@ TEST_TEAR_DOWN( TEST_IOT_GPIO )
     /* Make sure resources are freed up */
     iot_gpio_close( xtestIotGpioHandleA );
     iot_gpio_close( xtestIotGpioHandleB );
+
+    if ( xtestIotGpioSemaphore != NULL )
```
qiutongs, line 115: No need to check NULL because it is already asserted not NULL in TEST_SETUP.

PR #2060, opened 2020-05-31
Fix/test iot tsensor

wsheet:
…_INVALID_VALUE.

* In AFQP_IotTsensorCalibrationSuccess, there is call to iot_tsensor_ioctl() for
  sensor calibration.  The param is NULL but expect to return success.  It should
  expect invalid value as return code instead of success.

Description
-----------

Checklist:
----------

- [ x ] My code is Linted.


qiutongs:
eTsensorPerformCalibration doesn’t need the input parameter of buffer. 

https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/include/iot_tsensor.h#L190

Issue #2067, opened 2020-06-02
ESP32 Port SSID Max Length only 31 bytes due to strlcpy


danahern:
On an ESP32 board trying to connect to a 32 byte SSID fails.  

We have created an access point named “iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiK”31 ‘i' characters and a final 'K’ characters.

In this function, the function esp_wifi_scan_get_ap_records is used to fetch the scan access points, and then a loop is performed to copy records between structures.

When we print the SSID just before the strlcpy, we have the following output (in raw hex format):

I (20941) WIFI: 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69 
I (20947) WIFI: 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69 **4B**

When we print the output of the value copied in the structure right after the strlcpy, we have the following output:

W (20954) WIFI: 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69
W (20961) WIFI: 69 69 69 69 69 69 69 69 69 69 69 69 69 69 69 **00**

https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c#L565

Followed the code everywhere else and it should support it.  It looks like the iot_wifi.c#L565 is the only thing stopping it because it's setting the last byte to null due to strlcpy. 

Thanks!
Dan

shubhamkulkarni97:
Hi @danahern,

Have created a [PR](https://github.com/aws/amazon-freertos/pull/2073) to fix the issue.

While debugging I also observed that `sizeof` operator is used to calculate SSID length. Due to use of sizeof operator, NULL termination is also counted in the length. However, the [documentation](https://github.com/aws/amazon-freertos/blob/f264005442069bc7c5ba78ce2fa6405b144a6cd7/libraries/abstractions/wifi/include/iot_wifi.h#L110) states that length should not include NULL termination.

So, this should also be fixed in all the places.

danahern:
Fix looks good to me!

aggarw13:
The PR #2073 for the fix has been merged. 
@danahern, please let us know if you have any further questions about the issue or fix.

PR #2073, opened 2020-06-03
Fix issue related to WiFi scan and configure AP with SSID of maximum length on ESP32

shubhamkulkarni97:
Description
-----------
strlcpy sets last byte to NULL. Hence, length of SSID to be copied should be (max length + 1).

Using `sizeof` operator to calculate string length also counts NULL termination character. Hence, use strlen to calculate length of SSID and Passphrase.

Fixes issue https://github.com/aws/amazon-freertos/issues/2067

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/aws_demos/config_files/aws_wifi_config.h
+++ b/vendors/espressif/boards/esp32/aws_demos/config_files/aws_wifi_config.h
@@ -58,8 +58,8 @@
 
 /**
  * @brief Max passphrase length
+ * Maximum allowed WPA2 passphrase length (per specification) is 63
  */
-// Maximum allowed WPA2 passphrase length (per specification) is 63
 #define wificonfigMAX_PASSPHRASE_LEN          ( 63 )
 
 /**
--- a/vendors/espressif/boards/esp32/aws_demos/config_files/aws_wifi_config.h
+++ b/vendors/espressif/boards/esp32/aws_demos/config_files/aws_wifi_config.h
@@ -59,7 +59,8 @@
 /**
  * @brief Max passphrase length
  */
-#define wificonfigMAX_PASSPHRASE_LEN          ( 64 )
+// Maximum allowed WPA2 passphrase length (per specification) is 63
+#define wificonfigMAX_PASSPHRASE_LEN          ( 63 )
 
 /**
  * @brief Soft Access point SSID
--- a/vendors/espressif/boards/esp32/aws_tests/config_files/aws_wifi_config.h
+++ b/vendors/espressif/boards/esp32/aws_tests/config_files/aws_wifi_config.h
@@ -58,8 +58,8 @@
 
 /**
  * @brief Max passphrase length
+ * Maximum allowed WPA2 passphrase length (per specification) is 63
  */
-// Maximum allowed WPA2 passphrase length (per specification) is 63
 #define wificonfigMAX_PASSPHRASE_LEN          ( 63 )
 
 /**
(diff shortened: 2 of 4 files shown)


vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c
```diff
@@ -1291,9 +1291,11 @@ WIFIReturnCode_t WIFI_ConfigureAP( const WIFINetworkParams_t * const pxNetworkPa
         }
 
         /* ssid/password is required */
-        strlcpy((char *) &wifi_config.ap.ssid, pxNetworkParams->pcSSID, pxNetworkParams->ucSSIDLength);
+        // SSID can be a non NULL terminated string if ssid_len is specified. Hence memcpy is used.
+        memcpy((char *) &wifi_config.ap.ssid, pxNetworkParams->pcSSID, pxNetworkParams->ucSSIDLength);
+        wifi_config.ap.ssid_len = pxNetworkParams->ucSSIDLength;
         if (pxNetworkParams->xSecurity != eWiFiSecurityOpen) {
-            strlcpy((char *) &wifi_config.ap.password, pxNetworkParams->pcPassword, pxNetworkParams->ucPasswordLength);
+            strlcpy((char *) &wifi_config.ap.password, pxNetworkParams->pcPassword, pxNetworkParams->ucPasswordLength + 1);
```
aggarw13, line 1298: The copy operations of SSID ( using `memcpy` ) and password (using `strlcpy`) are inconsistent. Can they both use either `memcpy` or `strlcpy`?

shubhamkulkarni97: There is an option to have SSID as a non NULL terminated string and set the length in ssid_len (as specified in the [documentation](https://github.com/espressif/esp-afr-sdk/blob/ee3a985073c2474defb20ab353dd4be0dfd59956/components/esp32/include/esp_wifi_types.h#L213)).

In case of password, it is mandatory to have NULL terminated string. Hence, it is necessary to use `strlcpy` in case of password.

If `memcpy` is not used for SSID, the maximum length will be reduced to 31 bytes.

aggarw13: Okay, thank you for the clarification. A comment to explicitly mentioning the reason for using `memcpy` to support 32 characters long SSID name could be helpful.

vendors/espressif/boards/esp32/aws_demos/config_files/aws_wifi_config.h
```diff
@@ -59,7 +59,8 @@
 /**
  * @brief Max passphrase length
  */
-#define wificonfigMAX_PASSPHRASE_LEN          ( 64 )
+// Maximum allowed WPA2 passphrase length (per specification) is 63
```
dachalco, line 62: Good note. Could you please reformat into existing comment Line#59 ?
```suggestion
/**
 * @brief Max passphrase length
 * Maximum allowed WPA2 passphrase length (per specification) is 63
 */
```

vendors/espressif/boards/esp32/aws_tests/config_files/aws_wifi_config.h
```diff
@@ -59,7 +59,8 @@
 /**
  * @brief Max passphrase length
  */
-#define wificonfigMAX_PASSPHRASE_LEN          ( 64 )
+// Maximum allowed WPA2 passphrase length (per specification) is 63
```
dachalco, line 62: Ditto for this file. Please reformat this info into comment starting at Line#59 as suggested in my other comment.

aggarw13:
Re-approving the PR as style and build check failures reported by CI are from unrelated files.

PR #2084, opened 2020-06-03
Suppress -Wunused-but-set-variable warnings

aggarw13:
Fix `-Wunused-but-set-variable` warnings when building with (ARM) GCC using `-Wall -Wextra` flags

Description
---------------
The variables flagged are used only for logging; the warnings are raised when logging is disabled.

### DIFF
--- a/libraries/c_sdk/standard/https/src/iot_https_client.c
+++ b/libraries/c_sdk/standard/https/src/iot_https_client.c
@@ -1792,6 +1792,10 @@ static IotHttpsReturnCode_t _parseHttpsMessage( _httpParserInfo_t * pHttpParserI
     const char * pHttpParserErrorDescription = NULL;
     http_parser * pHttpParser = &( pHttpParserInfo->responseParser );
 
+    /* Disable -Wunused-but-set-variable for local variables used for logging. */
+    ( void ) parsedBytes;
+    ( void ) pHttpParserErrorDescription;
+
     IotLogDebug( "Now parsing HTTP message buffer to process a response." );
     parsedBytes = pHttpParserInfo->parseFunc( pHttpParser, &_httpParserSettings, pBuf, len );
     IotLogDebug( "http-parser parsed %d bytes out of %d specified.", parsedBytes, len );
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -1792,6 +1792,10 @@ static IotHttpsReturnCode_t _parseHttpsMessage( _httpParserInfo_t * pHttpParserI
     const char * pHttpParserErrorDescription = NULL;
     http_parser * pHttpParser = &( pHttpParserInfo->responseParser );
 
+    /* Disable -Wunused-but-set-variable for local variables used for logging. */
+    ( void ) parsedBytes;
+    ( void ) pHttpParserErrorDescription;
+
```
cobusve, line 1798: Why do we have these variables if they are not being used. The appropriate fix seems to be to delete the dead code?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2029,6 +2033,9 @@ static IotHttpsReturnCode_t _flushHttpsNetworkData( _httpsConnection_t * pHttpsC
     IotHttpsReturnCode_t networkStatus = IOT_HTTPS_OK;
     size_t numBytesRecv = 0;
 
+    /* Disable -Wunused-but-set-variable for local variables used for logging. */
+    ( void ) pHttpParserErrorDescription;
+
```
cobusve, line 2038: Again this seems to be dead code that needs to be removed.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -3267,6 +3274,10 @@ IotHttpsReturnCode_t IotHttpsClient_ReadHeader( IotHttpsResponseHandle_t respHan
     IotHttpsResponseParserState_t savedParserState = PARSER_STATE_NONE;
     size_t numParsed = 0;
 
+    /* Disable -Wunused-but-set-variable for local variables used for logging. */
+    ( void ) numParsed;
+    ( void ) pHttpParserErrorDescription;
+
```
cobusve, line 3280: Same here. General guidance is that using this technique to ignore unused variables should only be done for unused function parameters, and when that is the case there must be strong motivation why the function can meet it's interface contract without using all the parameters.

libraries/c_sdk/standard/https/src/iot_https_client.c
aggarw13: These variables are used in the logging macros

libraries/c_sdk/standard/https/src/iot_https_client.c
aggarw13: Same for this as well. This variable is used for logging.

libraries/c_sdk/standard/https/src/iot_https_client.c
aggarw13: These variables are used in the logging macros, but warnings for them are generated only when logging is disabled. This PR suppresses those warnings for builds with logging disabled.

PR #2087, opened 2020-06-04
Fix Cypress makefiles include paths and defines (#2070)

yanjos-dev:
Fix Cypress makefile include paths and defines

Description
-----------
The AFQP_Verify test was failing on both of the cypress boards. The root issue was that it was using the default MbedTLS config file instead of the "aws_mbedtls_config.h" file.

Changes I made to fix this:
- Fixed the way that the include paths and defines are added in aws.mk.
- Added the path to the aws_mbedtls_config.h file to the aws_demo.mk and aws_test.mk files for consistency.

Checklist:
----------
- [ ] My code is Linted.


aggarw13:
The commit is the same as the merge commit on master.

PR #2089, opened 2020-06-04
Add 1.4x MQTT API deprecation warning message for all supported toolc…

aggarw13:
Add a deprecation warning message for the 1.4x MQTT API for all supported toolchains

Description
-----------
As part of the deprecation path of 1.4x series of MQTT APIs, we will issue a build-time warning of deprecation. 
Note that `#warning` directive is not supported by all toolchains (as it is not part of the ISO C standard), thus, the logic uses toolchain-specific directives to support displaying the message on their respective build outputs.

Checklist:
----------

- [ ] My code is Linted.


aggarw13:
We decided to take a different strategy for communication about deprecation path of the `1.4.x` series instead of showing it as a compile-time warning.

PR #2090, opened 2020-06-04
OTA/ Start self test timer before job doc is requested.

pvyawaha:
This fix starts the self-test timer before it receives the OTA job document.

Description
-----------

The fix helps in handling conditions where the OTA job is forced cancelled after it is started to allow device to reset and rollback.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -725,6 +725,8 @@ static OTA_Err_t prvStartHandler( OTA_EventData_t * pxEventData )
     OTA_Err_t xReturn = kOTA_Err_None;
     OTA_EventMsg_t xEventMsg = { 0 };
 
+    /* Start self-test timer, if platform is in self-test. */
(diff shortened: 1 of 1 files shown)


libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -725,6 +725,8 @@ static OTA_Err_t prvStartHandler( OTA_EventData_t * pxEventData )
     OTA_Err_t xReturn = kOTA_Err_None;
     OTA_EventMsg_t xEventMsg = { 0 };
 
+    /* Start self-test timer, if platform is in self-test. */
+    prvStartSelfTestTimer();
```
dan4thewin, line 729: Can you explain why we moved the timer start?

pvyawaha: The timer should be started even in case device can not get a job document or OTA job is forced cancelled. This helps to ensure that in all cases timer is started and device can roll back if the self-test is not completed.

PR #2102, opened 2020-06-06
Fix: Mediatek wifi_ping fail on zero

alfred2g:
Mediatek Wifi_ping fail on zero

Description
-----------
if no pings are received fail wifi ping, as other errors could lie behind which are not reported due to driver code

Checklist:
----------

- [ X ] My code is Linted.


mingyue86010:
The branch need to rebase to top of master to get the build check pass.

PR #2113, opened 2020-06-08
Remove CBC from mbed TLS

lundinc2:
Remove CBC from mbedTLS

Description
-----------

Disable some algorithms unsupported by IOT Core. Namely ones using CBC. Alongside this, other features that only work with CBC have been disabled, as well as some features that are currently not being used, such as the self-test and version APIs.

Checklist:
----------

- [x ] My code is Linted.

### DIFF
--- a/libraries/3rdparty/mbedtls_config/aws_mbedtls_config.h
+++ b/libraries/3rdparty/mbedtls_config/aws_mbedtls_config.h
@@ -600,7 +600,7 @@
  *
  * Enable Cipher Block Chaining mode (CBC) for symmetric ciphers.
  */
-#define MBEDTLS_CIPHER_MODE_CBC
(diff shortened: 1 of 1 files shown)


libraries/3rdparty/mbedtls_config/aws_mbedtls_config.h
```diff
@@ -1343,7 +1343,7 @@
  *
  * Comment this macro to disable support for Encrypt-then-MAC
  */
-#define MBEDTLS_SSL_ENCRYPT_THEN_MAC
+//#define MBEDTLS_SSL_ENCRYPT_THEN_MAC
```
lundinc2, line 1346: This feature is only for CBC

lundinc2:
Added bonus, on microchip this saves 5,732 bytes of ROM.

PR #2117, opened 2020-06-09
Move test_dns_multiple_addresses to a new test group

gkwicker:
Description
-----------

Move the multiple-address DNS test to a separate test group.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -3262,7 +3262,7 @@ TEST( Full_TCP, AFQP_SOCKETS_htons_HappyCase )
 }
 /*-----------------------------------------------------------*/
 
-TEST( Full_TCP_Extended SOCKETS_dns_multiple_addresses )
+TEST( Full_TCP_Extended, SOCKETS_dns_multiple_addresses )
 {
     BaseType_t xResult = pdFAIL;
     uint32_t i;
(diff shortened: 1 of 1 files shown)


libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -440,6 +440,11 @@ size_t xHeapA;
 /* Group declaration required by Unity framework. */
 TEST_GROUP( Full_TCP );
 
+/* Extended test group for tests which aren't required for qualification
+ * but demonstrate additional features or preferred behavior.
+ */
+TEST_GROUP( Full_TCP_Extended );
```
AniruddhaKanhere, line 446: Should this be added to [aws_test_runner.c](https://github.com/aws/amazon-freertos/blob/master/tests/common/aws_test_runner.c#L82) as well? I am not sure what is required for release and thus I might be wrong.
If so, we need to add #define in [aws_test_runner_config.h](https://github.com/aws/amazon-freertos/blob/master/vendors/pc/boards/windows/aws_tests/config_files/aws_test_runner_config.h#L56) in vendor directory of all respective vendors.

gkwicker: Thanks @AniruddhaKanhere, we will not add the extended tests here.  These are optional tests for now.

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -3262,7 +3262,7 @@ TEST( Full_TCP, AFQP_SOCKETS_htons_HappyCase )
 }
 /*-----------------------------------------------------------*/
 
-TEST( Full_TCP_Extended, test_dns_multiple_addresses )
+TEST( Full_TCP_Extended SOCKETS_dns_multiple_addresses )
```
AniruddhaKanhere, line 3265: Oh, you missed a comma here :)
Causing tests to fail

gkwicker: fixed

PR #2132, opened 2020-06-11
Make each config of a CBMC proof have unique name

karkhaz:
Some CBMC proofs are built in several different 'configurations'. Prior
to this commit, the name of each of these configurations was the same as
the name of the original proof. This meant that CI would sometimes write
the proof artifacts of each of these configurations on top of each
other, since the artifacts are saved under a folder whose name is based
on the name of the proof.

This commit fixes this so that each configuration of each proof has a
unique name, formed by joining the name of the configuration to the name
of the proof with a double underscore---for example,
TaskIncrementTick__config_default.

### DIFF
--- a/tools/cbmc/proofs/make_configuration_directories.py
+++ b/tools/cbmc/proofs/make_configuration_directories.py
@@ -125,13 +125,18 @@ def process(folder, files):
                 """))
             LOGGER.error("The offending entry is %s", config)
             return
-        new_config_folder = os.path.join(folder, "config_" + configname)
+
+        new_entry = "%s__config_%s" % (json_content["ENTRY"], configname)
+
+        new_config_folder = os.path.join(folder, new_entry)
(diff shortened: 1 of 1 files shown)


tools/cbmc/proofs/make_configuration_directories.py
```diff
@@ -125,13 +125,18 @@ def process(folder, files):
                 """))
             LOGGER.error("The offending entry is %s", config)
             return
+
+        new_entry = "%s__config_%s" % (json_content["ENTRY"], configname)
```
markrtuttle, line 129: The problem is clashing batch job names resulting in clashing S3 folder names.  Don't the job names come from the name of the directory holding the proof files?  Don't we need to change the directory name to "proofname_config_configname" and not just the harness name?

karkhaz: Done (see line 131).

mingyue86010:
The branch need to rebase to top of master to get build test passed.

PR #2142, opened 2020-06-15
Use configPRINTF instead of directly using logging API

lundinc2:
Use configPRINTF instead of directly using logging API

Description
-----------
Use configPRINTF instead of directly using logging API. This allows using a logging mechanism that is configurable, as well as removing the dependency on vLoggingPrintf.

Checklist:
----------

- [x] My code is Linted.


Krall-12:
The attack around the sex zone musst be change, to one way ticket 🧬🙀

PR #2151, opened 2020-06-18
Fix SPI baud change and default config on ST

qiutongs:
Description
-----------

Baud fix: The low level LL_SPI_SetBaudRatePrescaler requires a prescaler parameter instead of the raw frequency. So we need to calculate the prescaler based on the system clock and desired frequency.

Default config fix: when open, the frequency should be calculated to the default one which is system clock divided by the default prescaler - 2.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
+++ b/vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
@@ -57,6 +57,8 @@ typedef struct IotSPIDescriptor
 } IotSPIDescriptor_t;
 /*-----------------------------------------------------------*/
 
+extern uint32_t SystemCoreClock;
+
 static SPI_HandleTypeDef xSpiHandleMap[] =
 {
     {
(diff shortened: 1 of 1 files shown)


vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c
```diff
@@ -279,7 +301,26 @@ int32_t iot_spi_ioctl( IotSPIHandle_t const pxSPIPeripheral,
                             break;
                     }
 
-                    LL_SPI_SetBaudRatePrescaler( pxSpi->Instance, ( ( IotSPIMasterConfig_t * ) pvBuffer )->ulFreq );
+                    uint32_t sysClkTmp = SystemCoreClock;
+                    uint32_t preScaler = 0;
+                    uint8_t divisor = 0;
+
+                    while( sysClkTmp > ( ( IotSPIMasterConfig_t * ) pvBuffer )->ulFreq )
```
gkwicker, line 308: This would be clearer if written as:

```c
while( ( sysClkTmp > ( ( IotSPIMasterConfig_t * ) pvBuffer )->ulFreq ) && ( divisor < 7 ) )
{
    divisor++;
    sysClkTmp = ( sysClkTmp >> 1 );
}	
```

qiutongs: That sounds better. Will do.

PR #2160, opened 2020-06-19
Common io/nrf52/spi

dachalco:
Nordic NRF52840-DK CommonIO SPI

Description
-----------
Based off our [ST implementation](https://github.com/aws/amazon-freertos/blob/master/vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c). Ioctls for querying number of bytes sent/read for current transaction have been deferred since there's no official way to query this from nRF_SDK driver. Though, I left code comments in their place which explain a potential way to achieve them. 

In separate PR, may request change for user configurable pin assignments in future, as this is lacking from current iot_spi.h

Checklist:
----------

- [ ] My code is Linted.
- [ ] Just as ST impl, pin assignments need to be decided for SPIM instances. Currently only assigned for SPIM_0 


dachalco:
Moved to PR#2175

PR #2175, opened 2020-06-24
Nordic NRF52840-DK CommonIO SPI

dachalco:
Nordic NRF52840-DK CommonIO SPI
\+ support to run existing CommonIO tests (except for _assisted_ tests. Read below)
\+ new HW loopback tests

Description
-----------
Based off our [ST implementation](https://github.com/aws/amazon-freertos/blob/master/vendors/st/boards/stm32l475_discovery/ports/common_io/iot_spi.c). 
Ioctls for querying number of bytes sent/read for current transaction have been deferred since there's no official way to query this from nRF_SDK driver. Though, I left code comments in their place which explain a potential way to achieve them. 

Pin assignments for SPI instances are set in ports/common_io/iot_spi_config.h. Currently, only SPI_1 has a default pin mapping and the testing uses it.

**Additional Improvements:**
Also made an improvement to existing `AFQP_IotSPI_CancelSuccess` by removing a pointless semaphore take. The semaphore is supposed to be given in the callback, but at its current state, the callback was not installed for that test in the first place. So the sem-take would always timeout after `testIotSPI_DEFAULT_SEMAPHORE_DELAY`. The test also had a race-condition wherein it could fail if the SPI transaction finished async, before reaching the call to`iot_spi_cancel`.

**Testing:**

- Add support to run existing CommonIO tests **_except_** for the _assisted_ testing (These involve an interface with RPi).
   To run the now supported tests, edit `aws_test_runner_config.h` to define `testrunnerFULL_COMMON_IO_ENABLED = 1`
- Added new, non-afqp, loopback tests which entail tying MOSI --> MISO for a single SPI instance. With the current configuration that means tying pins 1.13 and 1.14. These loopback tests can be enabled by defining `IOT_TEST_COMMON_IO_SPI_LOOPBACK = 1 ` in  `iot_test_common_io_config.h`

Run typical test flow after implementing either of the above notes.

Checklist:
----------

- [ X] My code is Linted.


libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -1119,10 +1130,6 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_CancelSuccess )
             TEST_ASSERT_EQUAL( IOT_SPI_SUCCESS, lRetVal );
         }
 
-        /* Wait to make sure operation was really canceled. */
-        xCallbackReturn = xSemaphoreTake( xtestIotSPISemaphore, testIotSPI_DEFAULT_SEMAPHORE_DELAY );
```
qiutongs, line 1123: Thanks for catching this bug. But should we fix it by setting the callback instead?

```
iot_spi_set_callback( xSPIHandle, prvSpiAsyncCallback, NULL );
```

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -1479,7 +1486,221 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_CancelFuzzing )
...
+        /* The callback should not be called to give sem back. Read, verify callback not called */
+        TEST_ASSERT( IOT_SPI_SUCCESS == iot_spi_read_sync( xSPIHandle, pucRxBuf, ulNBytes ) );
+        TEST_ASSERT( pdFAIL == xSemaphoreTake( xtestIotSPISemaphore, testIotSPI_DEFAULT_SEMAPHORE_DELAY / 2 ) );
+
+        /* Verify expected bytes. Because it's loopback, the configured dummy value
+         * should be transmitted, and thus read, for all N queried bytes.*/
+        TEST_ASSERT_EACH_EQUAL_UINT8( xConfig.ucDummyValue, pucRxBuf, ulNBytes );
+    }
+
+    lRetVal = iot_spi_close( xSPIHandle );
```
qiutongs, line 1526: I think the implementation should make sure the dummy value is reset to 0 in the close function.

libraries/abstractions/common_io/test/test_iot_spi.c
```diff
@@ -1479,7 +1486,221 @@ TEST( TEST_IOT_SPI, AFQP_IotSPI_CancelFuzzing )
...
+
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Assumes HW loopback where MOSI --> MISO.
+ *        Verify the happy-path of a syncronous write
+ *
+ */
+TEST( TEST_IOT_SPI, IotSPI_LoopBack_WriteSync )
```
qiutongs, line 1580: I see there is no verification in these two "write" tests, compared to read or transfer loopback tests. If it is difficult to add verification code, I would say remove these two test cases to not provide a wrong impression.

libraries/abstractions/common_io/test/test_iot_spi.c
dachalco: I did consider that, but removed it because there is a race condition. If the async spi transaction finishes before the [subsequent call to `iot_spi_cancel`](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/test/test_iot_spi.c#L1115), then the installed callback would have given the semaphore and [the following assertion](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/test/test_iot_spi.c#L1124) would be violated since taking the given sem would be `pdPASS`.

Concluded it'd be better to remove it for now, until more determinism can be applied to avoid the race condition. 
For example, if there was a portable CommonIO way to stop the SPI driver source clock, doing so could likely halt the pending transmission, at least for nrf52.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,598 @@
...
+    nrf_drv_spi_config_t config;
+} SpiContext_t;
+
+typedef struct IotSPIDescriptor
+{
+    SpiContext_t const * pxSpiContext; /* Nordic Handle */
+    IotSPIMasterConfig_t xConfig;      /* Master Configuration */
+    IotSPICallback_t xSpiCallback;     /* Callback function */
+    void * pvUserContext;              /* User context passed in callback */
+    uint8_t sOpened;                   /* Bit flags to track different states. */
```
qiutongs, line 55: Nitpick: rename this variable.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,598 @@
...
+        }
+    }
+};
+
+/*
+ * Forward declare event handler passed to NRF SPI controller.
+ * Invokes user callbacks for async modes.
+ * Sync modes poll atomic flags to track state
+ */
+static volatile bool bTransferDone = true;
```
qiutongs, line 121: Instead of global, this state should be associated with each SPI instance.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
```diff
@@ -0,0 +1,598 @@
...
+            CRITICAL_REGION_EXIT();
+
+            lError = IOT_SPI_TRANSFER_ERROR;
+        }
+        else
+        {
+            if( bSync )
+            {
+                /* For sync transfers, we block until they complete */
+                while( !bTransferDone )
```
qiutongs, line 583: If you want it non-blocking, you can use a semaphore as an alternative. You take the semaphore here and you give the semaphore in the event handler.

If you want it blocking, it seems you can pass a NULL event handler to nrf_drv_spi_transfer, in which case it is blocking mode.

qiutongs:
@vjjunnut Please take a look the changes of SPI tests.

libraries/abstractions/common_io/test/test_iot_spi.c
dachalco: Could you elaborate on this? Do you mean restoring the configuration to default values so the next open will also have initial values? Or just the dummy value?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
dachalco: I did consider using semaphore, but was basing of ST CommonIO where only vendor HAL is used. Also implemented as such to mimic Nordic [driver blocking/non-blocking implementation](https://github.com/aws/amazon-freertos/blob/master/vendors/nordic/nRF5_SDK_15.2.0/modules/nrfx/drivers/src/nrfx_spim.c#L523) which busy-polls an event register.

`nrf_drv_spi_transfer` doesn't take an event handler as an argument. At initialization, `nrf_drv_spi_init` takes an event handler and if it's NULL, then subsequent calls to `nrf_drv_spi_transfer` for that SPI instance will be blocking and it it's non-NULL then subsequent transfers will be async with the input callback. Given that, it's more work to constantly init/unit with and without callbacks just to have the HAL drivers internally implement blocking/non-blocking. Hence the permanent installation of `prvSpiEventCallback` which handles sync/async behaviour without having to init/uninit the SPI instance so frequently, and does so in a similar manner that [the HAL driver does it](https://github.com/aws/amazon-freertos/blob/master/vendors/nordic/nRF5_SDK_15.2.0/modules/nrfx/drivers/src/nrfx_spim.c#L523).

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
dachalco: Very true! Thank you for highlighting this.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
dachalco: Could you clarify which one?

libraries/abstractions/common_io/test/test_iot_spi.c
dachalco: The main check in these tests is to make sure that after sync the user callback is NOT called, and for async the user callback IS called. Exercised by:
`TEST_ASSERT( pdFAIL == xSemaphoreTake( xtestIotSPISemaphore, testIotSPI_DEFAULT_SEMAPHORE_DELAY / 2 ) );`

Can't verify the MISO data as it's a write as there's no RxBuffer and MISO is ignored/undefined.

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: Yeah, the test for cancel is not easy to be deterministic. I am OK with removing it for now.

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: Sorry, I confused you. I mean restoring the configuration to default values in next open. Dummy value is just one of the them.

The restore can happen in close or open.

libraries/abstractions/common_io/test/test_iot_spi.c
qiutongs: ok, maybe you can call read_async before calling the write? Then you can verify the the data returned by read_async.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_spi.c
qiutongs: I mean sOpened -> sStates

libraries/abstractions/common_io/test/test_iot_spi.c
dachalco: I like this. Adding this now.

libraries/abstractions/common_io/test/test_iot_spi.c
dachalco: This could likely work with separate Master and Slave instances, but currently the CommonIO only supports Master perspective. 

Wouldn't be possible using the same Master instance since async/sync transfers can't occur at the same time. A Master read is still a transfer, just one that only transmits dummy bytes while it clocks in MISO. Any read transaction that occurs with MOSI --> MISO tied will only read the dummy byte.

Couldn't do it with two masters connected to each other since data synchronization is needed and both will attempt to drive SCLK which is prohibited.

Issue #2172, opened 2020-06-23
Race condition when running the TCP echo client demo


miteshdedhia7:
**Describe the bug**
There is a race condition in the TCP echo client demo that causes "Demo Runner" thread to finish before the "Echo" thread has finished. `DEMO_RUNNER_RunDemos` creates a thread that runs the `runDemoTask` function. `runDemoTask` then runs the `DEMO_entryFUNCTION` which in the case of the TCP echo demo is `vStartTCPEchoClientTasks_SingleTasks` which creates "Echo" thread that runs the demo and returns immediately. The "Demo Runner" thread and the "Echo" thread are at the same priority and the `runDemoTask` does not wait for the "Echo" thread to finish before calling `_cleanup` causing the network to be disabled before the "Echo" thread runs. 

Up till the 201910.00 branch the example runs fine in spite of the race condition because the `_cleanup` function did not disable the network but after https://github.com/aws/amazon-freertos/commit/4b7ca8864f8e8772d959f364c4915bfed31a3f71 the example stopped working because the network is disabled before the "Echo" thread starts communication.

**System information**
- Which hardware board or part numbers?
**CY8CPROTO_062_4343W**
- IDE used
**ModusToolbox**
- Operating System 
**MacOS**
- Version of FreeRTOS (run ``git describe --tags`` to find it)
**202002.00**
- Project 
**TCP Echo**

**Expected behavior**
The demo runs successfully with sending and receiving the message.

**Screenshots or console output**
![image](https://user-images.githubusercontent.com/53099677/85470345-bcac0000-b563-11ea-8c48-bdb09de5f063.png)
Note that this screenshot was taken from the 201910.00 branch and hence runs the demo fine but notice that the "----DEMO FINISHED----" string printed in the `runDemoTask` appears even before the echo thread has started communication.

**To reproduce**
1. Run the TCP echo demo example on 202002.00 branch
2. Observe output.

Thank you!

aggarw13:
Hi @miteshdedhia7, thank you for informing us about the issue along with your analysis of the cause. We are looking into the issue, and will provide you an update soon.

AniruddhaKanhere:
Hello @miteshdedhia7, thanks for pointing out this issue.
I have created a PR https://github.com/aws/amazon-freertos/pull/2181 to resolve this race-condition. You might want to look at the PR and if you would like, try it out as well to see if your issue is resolved.
If the issue persists, tell us about it in comments below.

[...]

PR #2190, opened 2020-06-26
Sync Single Echo Client demo tasks

AniruddhaKanhere:
Description
-----------
Thanks to the issue: https://github.com/aws/amazon-freertos/issues/2172, a race condition in tasks (parent and child) was found in TCP_Echo_Single demo.
As the issue points out, the parent task finishes before the child (echo) task(s) and therefore the demo finishes even before the echo-client has a chance to connect to the echo server.

This PR rectifies the aforementioned race condition by making the parent task wait using Semaphores.
+ The child (echo) tasks were designed as a `for( ; ; )` and were not getting cleaned up after the demo. This PR addresses that as well. The iterations are limited by a certain amount and the parent task cleans up the child tasks after their completion.

Checklist:
----------
- [ ] My code is Linted.

### DIFF
--- a/demos/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/tcp/aws_tcp_echo_client_single_task.c
@@ -424,9 +424,14 @@ static void prvEchoClientTask( void * pvParameters )
         vTaskDelay( echoLOOP_DELAY );
     }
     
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere:
see https://github.com/aws/amazon-freertos/pull/2191

PR #2197, opened 2020-06-28
Shim Layer Implementation Using MQTT Lightweight API

gshvang:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -44,6 +44,11 @@
 #include "platform/iot_clock.h"
 #include "platform/iot_threads.h"
 
+/* MQTT v4_beta 2 ligghtweight library includes. */
```
leegeth, line 47: ligghtweight -> lightweight

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -164,8 +233,20 @@ static bool _mqttSubscription_setUnsubscribe( const IotLink_t * pSubscriptionLin
 
     return true;
 }
+/*Generate Id for packet*/
```
leegeth, line 236: please add a line and separator(/*-----------------------------------------------------------*/) between 2 functions

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+    connectInfo.cleanSession = pConnectInfo->cleanSession;
+    connectInfo.keepAliveSeconds = pConnectInfo->keepAliveSeconds;
+    connectInfo.pClientIdentifier = pConnectInfo->pClientIdentifier;
+    connectInfo.clientIdentifierLength = pConnectInfo->clientIdentifierLength;
+    connectInfo.pUserName = pConnectInfo->pUserName;
+    connectInfo.userNameLength = pConnectInfo->userNameLength;
+    connectInfo.pPassword = pConnectInfo->pPassword;
+    connectInfo.passwordLength = pConnectInfo->passwordLength;
+    MQTTFixedBuffer_t networkBuffer;
+    MQTTStatus_t mqttStatus = MQTTSuccess;
```
leegeth, line 859: please move the variable declarations to the beginning of the function

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+/* Connect Serialize Wrapper*/
+static IotMqttError_t _connectSerializeWrapper( const IotMqttConnectInfo_t * pConnectInfo,
+                                                uint8_t ** pConnectPacket,
+                                                size_t * pPacketSize )
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    size_t remainingLength = 0UL;
+    MQTTConnectInfo_t connectInfo;
+
+    connectInfo.cleanSession = pConnectInfo->cleanSession;
```
leegeth, line 850: Don't we need to assert for validating parameters?

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+    connectInfo.userNameLength = pConnectInfo->userNameLength;
+    connectInfo.pPassword = pConnectInfo->pPassword;
+    connectInfo.passwordLength = pConnectInfo->passwordLength;
+    MQTTFixedBuffer_t networkBuffer;
+    MQTTStatus_t mqttStatus = MQTTSuccess;
+
+    /*NULL Check willInfo */
+    if( pConnectInfo->pWillInfo != NULL )
+    {
+        MQTTPublishInfo_t willInfo;
```
leegeth, line 864: please move variable declarations to the beginning of the function

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+
+        /* Getting Connect packet size using MQTT V4_beta2 API*/
+
+        mqttStatus = MQTT_GetConnectPacketSize( &connectInfo,
+                                                &willInfo,
+                                                &remainingLength,
+                                                pPacketSize );
+
+        /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+        status = convertReturnCode( mqttStatus );
```
leegeth, line 881: I think we would need to proceed only if the status is success

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+        networkBuffer.size = *pPacketSize;
+
+        /*Serializing the connect packet using MQTT V4_beta2 API*/
+        mqttStatus = MQTT_SerializeConnect( &connectInfo,
+                                            NULL,
+                                            remainingLength,
+                                            &( networkBuffer ) );
+
+        /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+        status = convertReturnCode( mqttStatus );
```
leegeth, line 918: I think the code is repeated for getting packet size and serializing connect. Please refactor to avoid redundant code.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+        mqttStatus = MQTT_GetConnectPacketSize( &connectInfo,
+                                                NULL,
+                                                &remainingLength,
+                                                pPacketSize );
+
+        /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+        status = convertReturnCode( mqttStatus );
+
+        /*Allocating memory for Connect packet*/
+        networkBuffer.pBuffer = IotMqtt_MallocMessage( *pPacketSize );
```
leegeth, line 908: When will this memory be freed?

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+                                            &( networkBuffer ) );
+
+        /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+        status = convertReturnCode( mqttStatus );
+    }
+
+    *pConnectPacket = networkBuffer.pBuffer;
+    return status;
+}
+/*-----------------------------------------------------------*/
```
leegeth, line 924: a space required between the function and the separator.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    MQTTFixedBuffer_t networkBuffer;
+
+    MQTTStatus_t mqttStatus = MQTTSuccess;
+
+    /* Getting Disconnect packet size using MQTT V4_beta2 API*/
+    mqttStatus = MQTT_GetDisconnectPacketSize( pPacketSize );
+
+    /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+    status = convertReturnCode( mqttStatus );
```
leegeth, line 939: I don't think we need to proceed if the MQTT_GetDisconnectPacketSize call returns an error.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+
+    /* Allocate memory to hold the Disconnect packet. */
+    networkBuffer.pBuffer = IotMqtt_MallocMessage( *pPacketSize );
+    networkBuffer.size = *pPacketSize;
+
+    /*Serializing the Disconnect packet using MQTT V4_beta2 API*/
+    mqttStatus = MQTT_SerializeDisconnect( &( networkBuffer ) );
+
+    /*Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status*/
+    status = convertReturnCode( mqttStatus );
```
leegeth, line 949: Also check for error before proceeding further.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -754,6 +837,342 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+    MQTTStatus_t mqttStatus = MQTTSuccess;
+    size_t remainingLength = 0UL;
+    MQTTSubscribeInfo_t * subscriptionList = IotMqtt_MallocMessage( sizeof( MQTTSubscribeInfo_t ) * subscriptionCount );
+    size_t i = 0;
+
+    for( i = 0; i < subscriptionCount; i++ )
+    {
+        if( ( pSubscriptionList + i )->qos == IOT_MQTT_QOS_1 )
+        {
+            subscriptionList[ i ].qos = MQTTQoS1;
```
leegeth, line 973: can this just be assign with a cast?
You are already doing it in line 871.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -2011,7 +2434,47 @@ const char * IotMqtt_OperationType( IotMqttOperationType_t operation )
...
+
+        case MQTTBadResponse:
+            status = IOT_MQTT_BAD_RESPONSE;
+            break;
+
+        case MQTTServerRefused:
+            status = IOT_MQTT_SERVER_REFUSED;
+            break;
+
+        default:
```
leegeth, line 2469: I think there are other error codes in the new library. It needs to be handled. If not handled those errors will be considered as success.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -64,9 +69,9 @@ static bool _incomingPacketValid( uint8_t packetType );
  *
  * @return #IOT_MQTT_SUCCESS, #IOT_MQTT_NO_MEMORY or #IOT_MQTT_BAD_RESPONSE.
  */
-static IotMqttError_t _getIncomingPacket( void * pNetworkConnection,
-                                          const _mqttConnection_t * pMqttConnection,
-                                          _mqttPacket_t * pIncomingPacket );
+static IotMqttError_t _getIncomingPacket(void* pNetworkConnection,
+    const _mqttConnection_t* pMqttConnection,
+    _mqttPacket_t* pIncomingPacket);
```
leegeth, line 74: is this the correct uncrustify used? It seems like the earlier formatting was correct.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
gshvang: They are getting validated in the IOT_mqttConnect().After validating only ,I am using them in this serialization function.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,645 @@
...
+
+/*-----------------------------------------------------------*/
+
+/* Suback Serializer Wrapper .*/
+IotMqttError_t _pubackSerializeWrapper( uint16_t packetIdentifier,
+                                        uint8_t ** pPubackPacket,
+                                        size_t * pPacketSize )
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    MQTTStatus_t mqttStatus = MQTTSuccess;
```
yourslab, line 572: It would be nice to have a better name to distinguish between these two statuses more easily.

libraries/c_sdk/standard/mqtt_v2/src/mqtt_lightweight.c
```diff
@@ -1785,7 +1785,7 @@ MQTTStatus_t MQTT_SerializeAck( const MQTTFixedBuffer_t * pBuffer,
                                 uint16_t packetId )
 {
     MQTTStatus_t status = MQTTSuccess;
-
+    //packetType = MQTT_PACKET_TYPE_PUBACK;
```
yourslab, line 1788: Please remove comment

libraries/c_sdk/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -914,3 +914,146 @@ void _IotMqtt_CloseNetworkConnection( IotMqttDisconnectReason_t disconnectReason
                                       _mqttConnection_t * pMqttConnection );
 
 #endif /* ifndef IOT_MQTT_INTERNAL_H_ */
+
+/*----------------- MQTT Serailization /Deserialization Wrapper functions for Shim------------------*/
```
yourslab, line 918: I believe these wrapper functions should not be below the header include guards unless you had a specific reason for doing so.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -666,10 +670,10 @@ static void _sendPuback( _mqttConnection_t * pMqttConnection,
...
 
     /* Default PUBACK serializer and free packet functions. */
-    IotMqttError_t ( * serializePuback )( uint16_t,
-                                          uint8_t **,
-                                          size_t * ) = _IotMqtt_SerializePuback;
-    void ( * freePacket )( uint8_t * ) = _IotMqtt_FreePacket;
+    IotMqttError_t (* serializePuback)( uint16_t,
+                                        uint8_t **,
+                                        size_t * ) = _pubackSerializeWrapper;
+    void (* freePacket)( uint8_t * ) = _IotMqtt_FreePacket;
```
abhidixi11, line 676: Another example of code style diff. Please run uncrustify, it should reduce diff as well.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -79,76 +79,12 @@
...
- * maximum length smaller than the max "Remaining Length" constant above.
- */
-#define MQTT_PACKET_CONNECT_MAX_SIZE                ( 327700UL )
-
-/*
- * Constants relating to CONNACK packets, defined by MQTT 3.1.1 spec.
- */
-#define MQTT_PACKET_CONNACK_REMAINING_LENGTH        ( ( uint8_t ) 2 )    /**< @brief A CONNACK packet always has a "Remaining length" of 2. */
-#define MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK    ( ( uint8_t ) 0x01 ) /**< @brief The "Session Present" bit is always the lowest bit. */
+#define MQTT_PUBLISH_FLAG_DUP    ( 3 )                     /**< @brief Duplicate message. */
```
abhidixi11, line 86: Should this file, iot_mqtt_serialize.c completely disappear and instead get replaced by mqtt_light_weight.c ?

libraries/c_sdk/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -913,4 +735,146 @@ bool _IotMqtt_GetNextByte( void * pNetworkConnection,
...
 
+/*----------------- MQTT Serailization /Deserialization Wrapper functions for Shim------------------*/
+
+/**
+ * @brief The wrapper for  @ref _IotMqtt_SerializeConnect
+ *
+ * See @ref _IotMqtt_SerializeConnect for a
+ * description of the parameters and return values.
+ */
+IotMqttError_t _connectSerializeWrapper( const IotMqttConnectInfo_t * pConnectInfo,
```
abhidixi11, line 746: For any new code we should avoid _ at the start.  You can rename them as connectSerializeWrapper()

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
...
+    {
+        encodedSize = 4;
+    }
+
+    return encodedSize;
+}
+
+/*-----------------------------------------------------------*/
+
+uint8_t _IotMqtt_GetPacketType( void * pNetworkConnection,
```
abhidixi11, line 129: Since it is a new file, let's keep same coding style for this file. Either make all these functions _IotMqtt_XYZ or xYZ() ( get rid of ' _ '  )

gshvang: Ok, will keep it as _IotMqtt_XYZ

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -44,6 +44,12 @@
 #include "platform/iot_clock.h"
 #include "platform/iot_threads.h"
 
+/*MQTT v4_beta_2 library includes.*/
+#include "mqtt_config.h"
```
leegeth, line 48: we may not need the mqtt_config here. This config is only needed by the new MQTT library.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
+/*
+ * FreeRTOS MQTT V2.1.1
```
leegeth, line 2: Please update all the headers to indicate this is the Shim, not MQTT library. This can be done as part of a different PR.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -302,1638 +146,37 @@ static uint16_t _nextPacketIdentifier( void )
...
-}
-
-/*-----------------------------------------------------------*/
-
-IotMqttError_t _IotMqtt_SerializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
-                                          uint8_t ** pConnectPacket,
-                                          size_t * pPacketSize )
+void _IotMqtt_PublishSetDup( uint8_t * pPublishPacket,
+                             uint8_t * pPacketIdentifierHigh,
+                             uint16_t * pNewPacketIdentifier )
```
leegeth, line 151: As I recall, the plan is to get rid of this file later as part of a clean up task right?

leegeth: Can we rename the file to indicate this is not serializer any more?

abhidixi11: I would get rid of this file if possible to avoid confusion.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
...
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    size_t remainingLength = 0UL;
+    MQTTConnectInfo_t connectInfo;
+    MQTTFixedBuffer_t networkBuffer;
+    MQTTStatus_t managedMqttStatus = MQTTSuccess;
+    MQTTPublishInfo_t willInfo;
+
+    /* Null Check for connectInfo  */
+    IotMqtt_Assert( pConnectInfo != NULL );
```
leegeth, line 214: Also assert for non-NULL pointers for output params

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
...
+
+/* Disconnect Serialize Wrapper*/
+IotMqttError_t _IotMqtt_disconnectSerializeWrapper( uint8_t ** pDisconnectPacket,
+                                                    size_t * pPacketSize )
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    MQTTFixedBuffer_t networkBuffer;
+
+    MQTTStatus_t managedMqttStatus = MQTTSuccess;
+
```
leegeth, line 280: assert the output params to ensure those are non NULL pointers.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
...
+    publishInfo.qos = ( MQTTQoS_t ) pPublishInfo->qos;
+
+    if( pPublishInfo->qos == IOT_MQTT_QOS_1 )
+    {
+        publishInfo.dup = true;
+    }
+    else
+    {
+        publishInfo.dup = false;
+    }
```
leegeth, line 469: This doesn't seem right. Are we always marking any QoS1 packet as duplicate?

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,772 @@
...
+    MQTTStatus_t managedMqttStatus = MQTTSuccess;
+    MQTTPacketInfo_t pIncomingPacket;
+
+    /* Null Check for suback packet  */
+    IotMqtt_Assert( pSuback != NULL );
+
+    pIncomingPacket.type = pSuback->type;
+    pIncomingPacket.pRemainingData = pSuback->pRemainingData;
+    pIncomingPacket.remainingLength = pSuback->remainingLength;
+    /*Deserializing SUBACK packet received from the netwok*/
```
leegeth, line 585: netwok -> network

libraries/c_sdk/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -913,4 +735,146 @@ bool _IotMqtt_GetNextByte( void * pNetworkConnection,
 void _IotMqtt_CloseNetworkConnection( IotMqttDisconnectReason_t disconnectReason,
                                       _mqttConnection_t * pMqttConnection );
 
+/*----------------- MQTT Serailization /Deserialization Wrapper functions for Shim------------------*/
+
+/**
+ * @brief The wrapper for  @ref _IotMqtt_SerializeConnect
```
leegeth, line 741: Please add a bit more details in comments similar to the old serializer functions

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -165,7 +168,7 @@ static bool _mqttSubscription_setUnsubscribe( const IotLink_t * pSubscriptionLin
     return true;
 }
 
-/*-----------------------------------------------------------*/
+/*------------------------------------------------------------------------------*/
```
yourslab, line 171: Nit: I'm not sure if you intentionally had this line to be longer than the others but it is really no biggie

libraries/c_sdk/standard/mqtt_v2/CMakeLists.txt
```diff
@@ -23,10 +23,8 @@ afr_module_sources(
 
 afr_module_include_dirs(
     ${AFR_CURRENT_MODULE}
-    PUBLIC
+    INTERFACE
```
yourslab, line 26: Any reason you had for changing these to interface? I think they were already good as is. Usually, you would only be using INTERFACE for tests. If you would like more insight into the differences among PUBLIC vs PRIVATE vs INTERFACE, see [here](https://cmake.org/pipermail/cmake/2016-May/063400.html):
>    - When A links in B as *PRIVATE*, it is saying that A uses B in its
>    implementation, but B is not used in any part of A's public API. Any code
>    that makes calls into A would not need to refer directly to anything from
>    B. An example of this could be a networking library A which can be built to
>    use one of a number of different SSL libraries internally (which B
>    represents). A presents a unified interface for client code which does not
>    reference any of the internal SSL data structures or functions. Client code
>    would have no idea what SSL implementation (B) is being used by A, nor does
>    that client code need to care.
>    - When A links in B as *INTERFACE*, it is saying that A does not use B
>    in its implementation, but B is used in A's public API. Code that calls
>    into A may need to refer to things from B in order to make such calls. One
>    example of this is an interface library which simply forwards calls along
>    to another library but doesn't actually reference the objects on the way
>    through other than by a pointer or reference. Another example is where A is
>    defined in CMake as an interface library, meaning it has no actual
>    implementation itself, it is effectively just a collection of other
>    libraries (I'm probably over-simplifying here, but you get the picture).
>    - When A links in B as *PUBLIC*, it is essentially a combination of
>    PRIVATE and INTERFACE. It says that A uses B in its implementation and B is
>    also used in A's public API.

gshvang: Yeah, I am using managed MQTT API as INTERFACE in my shim. This is the reason I changed it to interface.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_publish_duplicates.c
```diff
@@ -0,0 +1,182 @@
...
+/*
+ * Positions of each flag in the first byte of an MQTT PUBLISH packet's
+ * fixed header.
+ */
+#define MQTT_PUBLISH_FLAG_DUP    ( 3 )                     /**< @brief Duplicate message. */
+
+
+/* Username for metrics with AWS IoT. */
+#if AWS_IOT_MQTT_ENABLE_METRICS == 1 || DOXYGEN == 1
+    #ifndef AWS_IOT_METRICS_USERNAME
```
leegeth, line 91: Where is this used?

libraries/c_sdk/standard/mqtt/src/iot_mqtt_publish_duplicates.c
```diff
@@ -0,0 +1,182 @@
...
+ * @brief Specify C SDK and version.
+ */
+        #define AWS_IOT_METRICS_USERNAME           "?SDK=C&Version=4.0.0"
+
+        /**
+         * @brief The length of #AWS_IOT_METRICS_USERNAME.
+         */
+        #define AWS_IOT_METRICS_USERNAME_LENGTH    ( ( uint16_t ) sizeof( AWS_IOT_METRICS_USERNAME ) - 1 )
+    #endif /* ifndef AWS_IOT_METRICS_USERNAME */
+#endif /* if AWS_IOT_MQTT_ENABLE_METRICS == 1 || DOXYGEN == 1 */
```
leegeth, line 103: Same comment as above

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,788 @@
...
+/* Unsubscribe Serialize Wrapper*/
+IotMqttError_t _IotMqtt_unsubscribeSerializeWrapper( const IotMqttSubscription_t * pUnsubscriptionList,
+                                                     size_t unsubscriptionCount,
+                                                     uint8_t ** pUnsubscribePacket,
+                                                     size_t * pPacketSize,
+                                                     uint16_t * pPacketIdentifier )
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    MQTTStatus_t managedMqttStatus = MQTTSuccess;
+
```
leegeth, line 392: Nit: Remove the extra space

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,788 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+/* Pingreq Serialize Wrapper*/
+IotMqttError_t _IotMqtt_pingreqSerializeWrapper( uint8_t ** pPingreqPacket,
+                                                 size_t * pPacketSize )
+{
+    IotMqttError_t serializeStatus = IOT_MQTT_SUCCESS;
+
```
leegeth, line 526: Nit:Remove extra spaces

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,788 @@
...
+    MQTTFixedBuffer_t networkBuffer;
+
```
leegeth, line 528: Nit:Remove extra spaces.

libraries/c_sdk/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -913,4 +735,178 @@ bool _IotMqtt_GetNextByte( void * pNetworkConnection,
 void _IotMqtt_CloseNetworkConnection( IotMqttDisconnectReason_t disconnectReason,
                                       _mqttConnection_t * pMqttConnection );
 
+/*----------------- MQTT Serailization /Deserialization Wrapper functions for Shim------------------*/
+
+/**
+ * @brief Generate a CONNECT packet from the given parameters by using MQTT v4 beta 2 serializer.
```
leegeth, line 741: v4 beta 2 may not be the official name of the library, but we can clean it up later once we have a finalized name.

libraries/c_sdk/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -913,4 +735,178 @@ bool _IotMqtt_GetNextByte( void * pNetworkConnection,
...
+ * @return #IOT_MQTT_SUCCESS if CONNACK specifies that CONNECT was accepted;
+ * #IOT_MQTT_SERVER_REFUSED if CONNACK specifies that CONNECT was rejected;
+ * #IOT_MQTT_BAD_RESPONSE if the CONNACK packet doesn't follow MQTT spec.
+ */
+IotMqttError_t _IotMqtt_deserializeConnackWrapper( _mqttPacket_t * pConnack );
+
+
+/**
+ * @brief Deserialize a suback packet received from the network by using MQTT v4 beta 2 deserializer.
+* Converts the packet from a stream of bytes to an #IotMqttError_t and extracts
```
leegeth, line 842: please uncrustify this file to fix this formatting

libraries/c_sdk/standard/mqtt_v2/src/mqtt_lightweight.c
```diff
@@ -1998,7 +1998,7 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * pIncomingPacket,
         LogError( ( "pSessionPresent cannot be NULL for CONNACK packet." ) );
         status = MQTTBadParameter;
     }
-    else if( pIncomingPacket->pRemainingData == NULL )
+    else if( ( pIncomingPacket->pRemainingData == NULL ) && ( pIncomingPacket->type != MQTT_PACKET_TYPE_PINGRESP ) )
```
leegeth, line 2001: Is this from syncing the new library?

gshvang: I will raise a different PR for it.This I fixed earlier before syncing with the new library when I found this bug.

leegeth: No change should be directly made to the library, Ideally, it should be fixed in C-SDK which is the Source of truth and then synced. I think I understand the reason why you had to do it. Without this change it will fail some cases. I think when syncing the library code in the PR that you mentioned, It will be handled.

libraries/c_sdk/standard/mqtt_v2/CMakeLists.txt
leegeth: I think the reason for INTERFACE to avoid statically compile the MQTT library. mqtt_config.h is required for the library to be statically compiled. mqtt_config.h is provided by the application of the library and hence it cannot be statically compiled as a library.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,784 @@
...
+    status = convertReturnCode( managedMqttStatus );
+
+    if( status == IOT_MQTT_SUCCESS )
+    {
+        /* Generating the packet id for publish packe. */
+        packetId = _nextPacketIdentifier();
+
+        /* Allocating memory to hold publish packet. */
+        pBuffer = IotMqtt_MallocMessage( *pPacketSize );
+        networkBuffer.pBuffer = pBuffer;
```
gkwicker, line 493: See previous comment about checking this value; if the check is delegated to `MQTT_SerializePublish()`, add a comment to that effect.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,784 @@
...
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+/* Pingreq Serialize Wrapper. */
+IotMqttError_t _IotMqtt_pingreqSerializeWrapper( uint8_t ** pPingreqPacket,
+                                                 size_t * pPacketSize )
+{
+    IotMqttError_t serializeStatus = IOT_MQTT_SUCCESS;
```
gkwicker, line 521: Initialize this value and the one on line 523 to failure.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,784 @@
...
+    /* Getting pingrequest packet size  using MQTT V4_beta2 API. */
+    managedMqttStatus = MQTT_GetPingreqPacketSize( pPacketSize );
+
+    /* Converting status code from MQTT v4_beta2 status to MQTT v4_beta 1 status. */
+    serializeStatus = convertReturnCode( managedMqttStatus );
+
+    if( serializeStatus == IOT_MQTT_SUCCESS )
+    {
+        /* Allocate memory to hold the Pingrequest packet. */
+        networkBuffer.pBuffer = IotMqtt_MallocMessage( *pPacketSize );
```
gkwicker, line 537: See previous comments about validating `networkBuffer.pBuffer`

libraries/c_sdk/standard/mqtt/src/iot_mqtt_publish_duplicates.c
```diff
@@ -0,0 +1,164 @@
...
+
+/* Atomic operations. */
+#include "iot_atomic.h"
+
+/*-----------------------------------------------------------*/
+
+/*
+ * Macros for reading the high and low byte of a 2-byte unsigned int.
+ */
+#define UINT16_HIGH_BYTE( x )    ( ( uint8_t ) ( x >> 8 ) )            /**< @brief Get high byte. */
```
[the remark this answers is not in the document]
gshvang: This functionality will be removed in Phase 2 approach of Shim Implementation and this file will no longer exists in Phase 2. So that's why kept all the definitions required in this file here itself.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,784 @@
...
+ * @param[in] length The value of the "Remaining length" to encode.
+ *
+ * @return The size of the encoding of length. This is always `1`, `2`, `3`, or `4`.
+ */
+static size_t _remainingLengthEncodedSize( size_t length );
+
+/*-----------------------------------------------------------*/
+
+/* Generate Id for packet. */
+static uint16_t _nextPacketIdentifier( void )
```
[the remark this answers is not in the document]
gshvang: This functionality will be removed in Phase 2 approach of Shim Implementation and this file will no longer exists in Phase 2. So that's why kept all the definitions required in this file here itself.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_publish_duplicates.c
```diff
@@ -0,0 +1,164 @@
...
+
+/*-----------------------------------------------------------*/
+
+void _IotMqtt_PublishSetDup( uint8_t * pPublishPacket,
+                             uint8_t * pPacketIdentifierHigh,
+                             uint16_t * pNewPacketIdentifier )
+{
+    uint16_t newPacketIdentifier = 0;
+
+    /* For an AWS IoT MQTT server, change the packet identifier. */
```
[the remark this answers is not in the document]
gshvang: The Publish Duplicates functionality is no longer supported and this file will be completely removed in phase 2 of Shim Implementation. We are following the phased approach to have a workable shim layer in each phase.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -0,0 +1,784 @@
...
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    MQTTStatus_t managedMqttStatus = MQTTSuccess;
+    size_t remainingLength = 0UL;
+    MQTTSubscribeInfo_t * subscriptionList = IotMqtt_MallocMessage( sizeof( MQTTSubscribeInfo_t ) * subscriptionCount );
+    size_t i = 0;
+    MQTTFixedBuffer_t networkBuffer;
+    uint16_t packetId = 0;
+
+    IotMqtt_Assert( pSubscriptionList != NULL );
```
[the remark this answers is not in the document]
gshvang: This assert is for the function input parameter (pSubscrptionList). Also, I will update the code for handling the out-of-resource conditions emitting the error log and returning IOT_MQTT_NO_MEMORY error.

PR #2200, opened 2020-06-29
Address some PKCS #11 TODOs

lundinc2:
Address some PKCS #11 TODOs

Description
-----------
* Added check to see if the parsed buffer was a certificate.
* Added macro for magic number for the EC point used in EC public keys.
* Determined byte size of an EC signature on the prime256v1 curve.
Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -166,8 +166,11 @@ static const char * pNoLowLevelMbedTlsCodeStr = "<No-Low-Level-Code>";
  *
  * @note This length needs to be updated if using a different curve.
  *
+ * To summarize:
+ * 32 points of 2 bytes each + 1 point length byte, 1 length byte, and
+ * 1 type (uncompressed) byte
  */
-#define pkcs11EC_POINT_LENGTH                 67
+#define pkcs11EC_POINT_LENGTH                 ( ( 32 * 2 ) + 1 + 1 + 1 )
 
 /**
  * @ingroup pkcs11_macros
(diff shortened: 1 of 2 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -150,6 +150,25 @@ static const char * pNoLowLevelMbedTlsCodeStr = "<No-Low-Level-Code>";
...
+ * mbed TLS encodes the length of the point in the first byte of the buffer it
+ * receives, so an additional 1 byte in length is added to account for this.
+ *
+ * In addition to this, an additional 1 byte is added to store information
+ * indicating that the point is uncompressed.
+ *
+ * @note This length needs to be updated if using a different curve.
+ *
+ */
+#define pkcs11EC_POINT_LENGTH                 67
```
gkwicker, line 170: This would be more readable.  The compiler will compute this and there will be no runtime penalty:

```c
// 32 points of 2 bytes each + 1 point length byte, 1 length byte, 1 type (uncompressed) byte
 #define pkcs11EC_POINT_LENGTH                 ( ( 32 ) * ( 2 ) + 1 + 1 + 1 )
```

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -2673,12 +2696,14 @@ CK_DECLARE_FUNCTION( CK_RV, C_GetAttributeValue )( CK_SESSION_HANDLE hSession,
...
-        else
+        else if( 0 == mbedtls_x509_crt_parse( &xMbedX509Context, pxObjectValue, ulLength ) )
         {
-            /* TODO: Do we want to safety parse the cert?
-             * Assume certificate. */
             xClass = CKO_CERTIFICATE;
         }
+        else
+        {
+            /* Could not determine the object class." */
```
gkwicker, line 2705: 1. Is this an error case?  If so, should you set an error in `xReturn`, e.g. `            xResult = CKR_OBJECT_HANDLE_INVALID;`?
1. There is an extraneous `"` at the end of the comment on line 2705.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c

PR #2208, opened 2020-07-01
Update CHANGELOG.md, readme, and Cypress readme

yngki:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


yngki:
Small change... we are no longer putting the exact date, but just the month now.  This way if a delay changes a day or 2, we don't need to update it.

PR #2225, opened 2020-07-03
UDP: Add ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS compile time constant

htibosch:
UDP: Add ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS compile time constant
-----------

Description
-----------
In compliance with the UDP protocol specification the FreeRTOS+TCP stack currently processes UDP packets that have their checksum field set to 0.  This update introduces a new compile time constant `ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS` that defaults to 0.  The TCP/IP stack will drop UDP packets that have their checksum field set to zero unless that constant is set to 1.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
@@ -314,11 +314,7 @@ from the FreeRTOSIPConfig.h configuration header file. */
 #endif
 
 #ifndef ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND
-	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND	1
-#endif
-
-#ifndef ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS
-	#define ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS	0
+	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 1
 #endif
 
 #ifndef ipconfigUDP_TIME_TO_LIVE
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
@@ -314,7 +314,11 @@ from the FreeRTOSIPConfig.h configuration header file. */
 #endif
 
 #ifndef ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND
-	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 1
+	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND	1
+#endif
+
+#ifndef ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS
+	#define ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS	0
 #endif
 
 #ifndef ipconfigUDP_TIME_TO_LIVE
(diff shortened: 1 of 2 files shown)


libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -314,7 +314,11 @@ from the FreeRTOSIPConfig.h configuration header file. */
 #endif
 
 #ifndef ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND
-	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 1
+	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND	1
```
gkwicker, line 317: Is this documented anywhere?  Why would anyone want to set `ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND==0`?

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -314,7 +314,11 @@ from the FreeRTOSIPConfig.h configuration header file. */
...
+#endif
+
+#ifndef ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS
+	#define ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS	0
```
gkwicker, line 321: Same comment as above, where is this documented?  Why would I want to define `ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS==1`?

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
htibosch: This macro already existed in the UDP-only library, I only changed white-space.

If the macro is not defined, calling `sendto()` with an unbound socket will fail.

>  Why would anyone want to set `ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND==0`?

Good question. I don't see an advantage of forcing programmers to bind a socket explicitly.

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
htibosch: As you know: an UDP checksum has a value between `0x0001` and `0xFFFF`.
If the calculated checksum is 0, `0xFFFF` will be used in stead.
A value of `0x0000` means: no checksum. It is up to the receiver to accept it or not.
I guess that in the old days, it was a way to increase network performance.

htibosch:
I made a new commit: "Added same check in case the driver takes care of the RX checksum calculations".

When `ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM == 1`, it means that the function `usGenerateProtocolChecksum()` won't be called by `prvAllowIPPacket()`.

I added the following code ( simplified here ):
~~~c
#if( ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM == 0 )
{
    /* Library is responsible for checksum verifications. */
    if( usGenerateProtocolChecksum() != ipCORRECT_CRC )
    {
        /* Protocol checksum not accepted. */
        eReturn = eReleaseBuffer;
    }
}
#else
{
    #if( ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS == 0 )
    {
        if( pxIPPacket->xIPHeader.ucProtocol == ( uint8_t ) ipPROTOCOL_UDP )
        {
            if( *pusChecksum == ( uint16_t ) 0U )
            {
                /* Packet will be dropped, send some logging. */
                eReturn = eReleaseBuffer;
            }
        }
    }
}
#endif

I have tested it between 2 boards: Xilinx Zynq and an STM32Fx, both with and without checksum offloading enabled.
~~~

PR #2226, opened 2020-07-03
Add ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS compile time constant

htibosch:
Add ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS compile time constant
-----------
Description
-----------
The FreeRTOS+TCP does not support IP options, but does accept packets that contain IP options.  This update introduces a new compile time constant `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS` that defaults to 1.  When `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS` is set to 1 FreeRTOS+TCP will behave as per previous versions.  When `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS` is set to 0 FreeRTOS+TCP will drop packets that contain IP options.

Checklist:
----------

- [x] My code is Linted.


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1635,23 +1635,33 @@ uint8_t ucProtocol;
...
+				( void ) memmove( pucTarget, pucSource, xMoveLen );
+				pxNetworkBuffer->xDataLength -= optlen;
+
+				/* Fix-up new version/header length field in IP packet. */
+				pxIPHeader->ucVersionHeaderLength = ( pxIPHeader->ucVersionHeaderLength & 0xF0U ) | /* High nibble is the version. */
+													( ( ipSIZE_OF_IPv4_HEADER >> 2 ) & 0x0FU );
+			}
+			#else
+			{
+				/* 'ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS' is set, so packets carrying
```
ravibhagavandas, line 1660: nit pick: `is not set`

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -314,7 +314,11 @@ from the FreeRTOSIPConfig.h configuration header file. */
 #endif
 
 #ifndef ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND
-	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND 1
+	#define ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND	1
```
gkwicker, line 317: Where is this documented?  Why would someone set `ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND==0`?

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -314,7 +314,11 @@ from the FreeRTOSIPConfig.h configuration header file. */
...
+#endif
+
+#ifndef ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS
+	#define ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS	1
```
gkwicker, line 321: Where is this documented?  Why would someone set `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS==0`?

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1635,23 +1635,33 @@ uint8_t ucProtocol;
...
-												( ( ipSIZE_OF_IPv4_HEADER >> 2 ) & 0x0FU );
+			#if( ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS != 0 )
+			{
+				/* All structs of headers expect a IP header size of 20 bytes
+				 * IP header options were included, we'll ignore them and cut them out
+				 * Note: IP options are mostly use in Multi-cast protocols */
+				const size_t optlen = ( ( size_t ) uxHeaderLength ) - ipSIZE_OF_IPv4_HEADER;
+				/* From: the previous start of UDP/ICMP/TCP data */
+				const uint8_t *pucSource = ( const uint8_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ sizeof( EthernetHeader_t ) + uxHeaderLength ] );
+				/* To: the usual start of UDP/ICMP/TCP data at offset 20 from IP header */
```
gkwicker, line 1646: nit: would specify `20 (decimal)` in this comment since `0x20` would also be a valid header size when options are present.  Alternately, you could eliminate this value from the comment as a wary reader would only trust the value of the `ipSIZE_OF_IPv4_HEADER` constant anyway.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1635,23 +1635,33 @@ uint8_t ucProtocol;
...
-			pxNetworkBuffer->xDataLength -= optlen;
-
-			/* Fix-up new version/header length field in IP packet. */
-			pxIPHeader->ucVersionHeaderLength = ( pxIPHeader->ucVersionHeaderLength & 0xF0U ) | /* High nibble is the version. */
-												( ( ipSIZE_OF_IPv4_HEADER >> 2 ) & 0x0FU );
+			#if( ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS != 0 )
+			{
+				/* All structs of headers expect a IP header size of 20 bytes
+				 * IP header options were included, we'll ignore them and cut them out
+				 * Note: IP options are mostly use in Multi-cast protocols */
```
gkwicker, line 1642: Is this the case?  Multicast isn't mentioned in the [IP options wikipedia page](https://en.wikipedia.org/wiki/Internet_Protocol_Options), and the [IPv4 header wikipedia page](https://en.wikipedia.org/wiki/IPv4#Header) shows a variety of options which seem unrelated to multicast.  I would delete the comment on line 1642.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1635,23 +1635,33 @@ uint8_t ucProtocol;
...
+				uint8_t *pucTarget = ( uint8_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ sizeof( EthernetHeader_t ) + ipSIZE_OF_IPv4_HEADER ] );
+				/* How many: total length minus the options and the lower headers */
+				const size_t  xMoveLen = pxNetworkBuffer->xDataLength - ( optlen + ipSIZE_OF_IPv4_HEADER + ipSIZE_OF_ETH_HEADER );
+
+				( void ) memmove( pucTarget, pucSource, xMoveLen );
+				pxNetworkBuffer->xDataLength -= optlen;
+
+				/* Fix-up new version/header length field in IP packet. */
+				pxIPHeader->ucVersionHeaderLength = ( pxIPHeader->ucVersionHeaderLength & 0xF0U ) | /* High nibble is the version. */
+													( ( ipSIZE_OF_IPv4_HEADER >> 2 ) & 0x0FU );
```
gkwicker, line 1656: It looks like we are adjusting the length here as if there had never been any options.  If that's the case, I'd be more explicit in the comment on line 1654, e.g. `Rewrite the Version/IHL byte to indicate that this packet has no IP options.`

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1635,23 +1635,33 @@ uint8_t ucProtocol;
...
+			}
+			#else
+			{
+				/* 'ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS' is set, so packets carrying
+				IP-options will be dropped. */
```
gkwicker, line 1661: nit: `Any packets containing IP options will be dropped.`

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
htibosch: I answered this question in PR #2225 already.

libraries/freertos_plus/standard/freertos_plus_tcp/include/FreeRTOSIPConfigDefaults.h
htibosch: When the FreeRTOS+TCP stack see IP-options, it will call `memmove()` to get rid of the extra data.
Some developers may want to refuse any packet that contains IP-options. This option make this possible.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: Yes, you're right, the options are more about routing and so. I will remove the comment line. Thanks

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: Changed to `20 (decimal)`

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: Replaced the comment with the proposed one.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: The full comments is:
~~~c
	/* 'ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS' is not set, so packets carrying
	IP-options will be dropped. */
~~~
Is much the same, I think?

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: Not a nit pick: it could be very confusing for readers. Thanks

PR #2227, opened 2020-07-03
Add additional length check when checksum offloading is used

htibosch:
Add additional length check when checksum offloading is used
-----------
Description
-----------
The FreeRTOS+TCP stack already has a compile time constant named `ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM`.  When that constant is set to 0 the TCP/IP stack validates IP packet checksum and length fields in software.  When that constant is set to 1 those checks are offloaded to the networking hardware.  This update changes that behaviour so the length fields are checked in software even when `ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM` is set to 1.

Checklist:
----------

- [x] My code is Linted.


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1894,6 +1907,129 @@ uint8_t ucProtocol;
...
+				break;
+			}
+
+			/* Parse the packet length. */
+			pxIPPacket = ipPOINTER_CAST( const IPPacket_t *, pucEthernetBuffer );
+
+			/* Per https://tools.ietf.org/html/rfc791, the four-bit Internet Header
+			Length field contains the length of the internet header in 32-bit words. */
+			ucVersionHeaderLength = pxIPPacket->xIPHeader.ucVersionHeaderLength;
+			if( ( ucVersionHeaderLength < 0x45U ) || ( ucVersionHeaderLength > 0x4FU ) )
```
AniruddhaKanhere, line 1942: Can we make 0x45U and 0x4FU something more descriptive?

gkwicker: +1, the `0x40` upper nibble indicates IPv4 and the [`0x05-0x0f` lower nibble are the valid values for the IHL](https://en.wikipedia.org/wiki/IPv4#Header).  There should be constants or `#defines` for these somewhere.  Also, the comment on lines 1939-1940 only mention length but we are also validating version here as well.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1894,6 +1907,129 @@ uint8_t ucProtocol;
...
+	size_t uxLength;
+	const IPPacket_t * pxIPPacket;
+	UBaseType_t uxIPHeaderLength;
+	ProtocolPacket_t *pxProtPack;
+	uint8_t ucProtocol;
+	uint16_t usLength;
+	uint16_t ucVersionHeaderLength;
+	BaseType_t xLocation = 0;
+	size_t uxMinimumLength;
+	BaseType_t xResult = pdPASS;
```
gkwicker, line 1924: Defensive programming opportunity: prefer to initialize return values to failure, and only set them to success when all tests have passed.  As an added benefit, this sometimes allows you to write less code, which would be the case here.

If you initialized `BaseType_t xResult = pdFAIL;` on line 1924, that would allow you to delete lines 1931,1944,1954,1963,1996,2002, and 2017 and replace them with a single `xResult = pdPASS;` between lines 2020 and 2021.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
htibosch: I propose these two macro's for this purpose:
~~~c
/* The first byte in the IPv4 header combines the IP version (4) with
with the length of the IP header. */
#define	ipIPV4_VERSION_HEADER_LENGTH_MIN	0x45U
#define	ipIPV4_VERSION_HEADER_LENGTH_MAX	0x4FU
~~~

> only mention length but we are also validating version here as well.

What about the following?
~~~C
    /* Test if the length of the IP-header is between 20 and 60 bytes,
    and if the IP-version is 4. */
    else if( ( pxIPHeader->ucVersionHeaderLength < ipIPV4_VERSION_HEADER_LENGTH_MIN ) ||
             ( pxIPHeader->ucVersionHeaderLength > ipIPV4_VERSION_HEADER_LENGTH_MAX ) )
~~~

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1656,8 +1656,6 @@ uint8_t ucProtocol;
 		{
 			/* The size of the IP-header is larger than 20 bytes.
 			The extra space is used for IP-options. */
```
gkwicker, line 1658: 1. It looks like we've removed `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS`, so this is no longer configurable correct?
1. A `{` was removed from line 1660, but I don't see a corresponding `}` removed anywhere.  Did this code not compile previously?

htibosch: ad 1) `ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS` was not part of PR 2227, `ipconfigUDP_PASS_ZERO_CHECKSUM_PACKETS` neither.
Now I rebased this PR 2227 to master, so it includes both 2225, and 2226.

ad 2) when I introduced: `ipIPV4_VERSION_HEADER_LENGTH_MIN`/`MAX`, I painstakingly copied an `f( ipconfigIP_PASS_PACKETS_WITH_IP_OPTIONS != 0)`, which didn't have a matching `#endif`. Later I removed that from two PR's.

htibosch: I just updated `FreeRTOSIPConfigDefaults.h`, it still missed the changes from the 2 earlier PR's. Hopefully it is mergeable now.
Thanks, Hein

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1608,6 +1614,16 @@ eFrameProcessingResult_t eReturn = eProcessBuffer;
 	}
 	#else
 	{
+
+		if (eReturn == eProcessBuffer )
+		{
+			if( xCheckSizeFields( ( uint8_t * )( pxNetworkBuffer->pucEthernetBuffer ), pxNetworkBuffer->xDataLength ) != pdPASS )
+			{
+				/* Some of the lengths checks was not successful. */
```
gkwicker, line 1622: grammar: _Some of the length checks were not successful._

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -1954,6 +1970,124 @@ uint8_t ucProtocol;
...
+			}
+
+			uxLength = ( size_t ) usLength;
+			uxLength -= ( ( uint16_t ) uxIPHeaderLength ); /* normally minus 20 */
+
+			if( ( uxLength < ( ( size_t ) sizeof( pxProtPack->xUDPPacket.xUDPHeader ) ) ) ||
+				( uxLength > ( ( size_t ) ipconfigNETWORK_MTU - ( size_t ) uxIPHeaderLength ) ) )
+			{
+				/* Again, in a 16-bit return value there is no space to indicate an
+				error.  For incoming packets, 0x1234 will cause dropping of the packet.
```
gkwicker, line 2072: This comment doesn't seem to match the logic in this block.  This function returns `BaseType_t` with only `pdPASS` and `pdFAIL` values.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
AniruddhaKanhere: corrected, thanks

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
AniruddhaKanhere: Updated this comment as well

PR #2254, opened 2020-07-09
Update flash offset align check

jychoi-inoon:
Update flash offset align check

Description
-----------

There is flash start offset align check condition in unit test case of test_iot_flash.c 
I think it assumes the flash start offset is less than block size. 
If the flash start offset is set to the even multiple of the block size, the block size alignment check fails. 
For example, 

- pxFlashInfo->ulBlockSize = 0x10000 (64K) 

- ultestIotFlashStartOffset = 2 * pxFlashInfo->ulBlockSize

- Then the alignment check fails. 

This update extends the alignment check condition to support the flash start offset larger than the block size. 

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/test_iot_flash.c
+++ b/libraries/abstractions/common_io/test/test_iot_flash.c
@@ -488,7 +488,8 @@ TEST( TEST_IOT_FLASH, AFQP_IotFlashEraseFlashBlocks )
         TEST_ASSERT_NOT_EQUAL( NULL, pxFlashInfo );
 
         /* Make sure the offset is aligned to BlockSize */
-        if( ultestIotFlashStartOffset == ( ultestIotFlashStartOffset & pxFlashInfo->ulBlockSize ) )
+        if( ( ultestIotFlashStartOffset == ( ultestIotFlashStartOffset & pxFlashInfo->ulBlockSize ) ) || 
+            ( 0 == ( ultestIotFlashStartOffset % pxFlashInfo->ulBlockSize ) ) )
         {
             /* If Erase asyc is supported, register a callback */
             if( pxFlashInfo->ucAsyncSupported )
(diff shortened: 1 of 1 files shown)


libraries/abstractions/common_io/test/test_iot_flash.c
```diff
@@ -488,7 +488,8 @@ TEST( TEST_IOT_FLASH, AFQP_IotFlashEraseFlashBlocks )
         TEST_ASSERT_NOT_EQUAL( NULL, pxFlashInfo );
 
         /* Make sure the offset is aligned to BlockSize */
-        if( ultestIotFlashStartOffset == ( ultestIotFlashStartOffset & pxFlashInfo->ulBlockSize ) )
+        if( ( ultestIotFlashStartOffset == ( ultestIotFlashStartOffset & pxFlashInfo->ulBlockSize ) ) ||
```
ravibhagavandas, line 491: Assuming `pxFlashInfo->ulBlockSize` is a power of 2 could these two checks can be combined to a single check as `( ultestIotFlashStartOffset & ( pxFlashInfo->ulBlockSize - 1 ) ) ! = 0` ?

You can additionally have a check for block size is power of 2 by `( ( pxFlashInfo->ulBlockSize != 0 ) && ( ( pxFlashInfo->ulBlockSize & ( pxFlashInfo->ulBlockSize - 1 ) ) == 0)`

jychoi-inoon: Thank you. I agree with your opinion. Code is updated.

ravibhagavandas:
The code style checks are failing which is blocking the merge of the PR. Can you uncrustify the file using the config here:
https://github.com/aws/amazon-freertos/blob/master/tools/uncrustify.cfg

libraries/abstractions/common_io/test/test_iot_flash.c
ravibhagavandas: Thanks, It may makes sense to move the logic of checking if blocksize is a power of 2  into a separate `if` condition with a TEST assert with relevant message. Or atleast, could we leave a comment so its clear what is being done.

jychoi-inoon: Thanks. 
I think the better location of checking if block size is power of 2 is where validating the flash info. 
And I also checked the flash size, block size, sector size and page size.

PR #2256, opened 2020-07-09
MISRA related changes in FreeRTOS_DNS.c

AniruddhaKanhere:
Description
-----------
This PR adds changes related to MISRA correction in DNS{.c, .h} files.
Some corrections required code rework which has been done and tested Using Full_TCP test suite and Full_FreeRTOS_TCP test suite.

Remaining MISRA warnings are regarding:
- Pointer conversions which are essential to map data buffers to various `struct`s.  
- Use of `#includes` after some c-code. These `#includes` are used in "packing" the `structs`.

Checklist:
----------

- [ ] My code is Linted.


libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -112,6 +112,9 @@ type. */
 	#endif
 #endif
 
+/* Define the ASCII value of '.' (Period/Full-stop). */
+#define ASCII_BASELINE_DOT						46UL
```
htibosch, line 116: As `ASCII_BASELINE_DOT` will be cast to `uint8_t`, i think the 'L' suffix is not necessary, `46U` is enough.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,343 @@ uint16_t usType = 0U;
 	/* Ensure that the buffer is of at least minimal DNS message length. */
 	if( uxBufferLength < sizeof( DNSMessage_t ) )
 	{
-		return dnsPARSE_ERROR;
+		xReturn = pdFALSE;
 	}
```
htibosch, line 1113: I would suggest to begin an else clause here, like :
~~~c
    if( uxBufferLength < sizeof( DNSMessage_t ) )
    {
        xReturn = pdFALSE;
    }
    else
    {
        uxSourceBytesRemaining = uxBufferLength;

        /* Parse the DNS message header. Map the byte stream onto a structure 
        * for easier access. */
        pxDNSMessageHeader = ipPOINTER_CAST( DNSMessage_t *, pucUDPPayloadBuffer );

        /* Introduce a do {} while (0) to allow the use of breaks. */
        do
        {
        size_t uxBytesRead = 0U;
        // etc
~~~

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,343 @@ uint16_t usType = 0U;
...
-								/* This device has requested this DNS look-up.
-								The result may be stored in the DNS cache. */
-								xDoStore = pdTRUE;
-							}
-						}
-						#endif	/* ipconfigDNS_USE_CALLBACKS == 1 */
-						#if( ipconfigUSE_DNS_CACHE == 1 )
+						/* This is the required record type and is of sufficient size. */
+						/* Mapping pucByte to a DNSAnswerRecord allows easy access of the
+						 * feilds of the structure. */
```
htibosch, line 1267: feilds = fields

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,343 @@ uint16_t usType = 0U;
...
+							/* Malformed response. */
+							xReturn = pdFALSE;
+							break;
+						}
 					}
 					else
 					{
-						/* Malformed response. */
-						return dnsPARSE_ERROR;
+						/* Do nothing */
```
htibosch, line 1344: Or maybe:
~~~c
    /* The IP-address found will be returned. */
~~~

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1590,29 +1609,37 @@ static Socket_t prvCreateDNSSocket( void )
...
+BaseType_t xReturn, xResult = pdPASS;
 
 	/* This must be the first time this function has been called.  Create
 	the socket. */
 	xSocket = FreeRTOS_socket( FREERTOS_AF_INET, FREERTOS_SOCK_DGRAM, FREERTOS_IPPROTO_UDP );
 	if( ( xSocket == FREERTOS_INVALID_SOCKET ) || ( xSocket == NULL ) )
 	{
-		return NULL;
+		xResult = pdFAIL;
 	}
```
htibosch, line 1621: Or maybe like this:
~~~c
    if( ( xSocket == FREERTOS_INVALID_SOCKET ) || ( xSocket == NULL ) )
    {
        xResult = pdFAIL;
        /* There was an error, return NULL. */
        xSocket = NULL;
    }
    else
    {
        /* Auto bind the port. */
        xAddress.sin_port = 0U;
        xReturn = FreeRTOS_bind( xSocket, &xAddress, sizeof( xAddress ) );
~~~

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: Done, thanks!

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: Good catch. thanks!

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: Oops! Corrected.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: I kept it as is because the returning of IP-address depends on the value of `xExpected`. I added the explanation there.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: Done!

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,341 @@ uint16_t usType = 0U;
...
+				}
+				else
+				{
+					xReturn = pdFALSE;
+					break;
+				}
 			}
-			else
+
+			if( xReturn == pdFAIL )
```
yuhui-zheng, line 1203: pdFALSE maybe? (Mixing pdFALSE/pdTRUE with pdFAIL/pdPASS may be confusing.)

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,341 @@ uint16_t usType = 0U;
...
-	if( xExpected == pdFALSE )
+	if( xReturn == pdFALSE )
+	{
+		/* There was an error while parsing the DNS response. Return error code. */
+		ulIPAddress = dnsPARSE_ERROR;
+	}
+	else if( xExpected == pdFALSE )
 	{
 		/* Do not return a valid IP-address in case the reply was not expected. */
 		ulIPAddress = 0UL;
```
yuhui-zheng, line 1440: But...  isn't "#define dnsPARSE_ERROR							0UL"?  By reading the comments, I got the sense of the if/else-if branch should return something different...

AniruddhaKanhere: Yes, `dnsPARSE_ERROR` is defined as `0UL`.
I intentionally kept that in a different `if-else` clause since we may want to define the `dnsPARSE_ERROR` as something else at a later time. Also, I thought that using different `if-else` cases might help differentiate that if `xExpected == pdFALSE` then we are not returning a valid IP address because it was not expected and in case there was an error, we return `dnsPARSE_ERROR` regardless of the value of `dnsPARSE_ERROR`.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
```diff
@@ -1112,321 +1109,341 @@ uint16_t usType = 0U;
 	/* Ensure that the buffer is of at least minimal DNS message length. */
 	if( uxBufferLength < sizeof( DNSMessage_t ) )
 	{
-		return dnsPARSE_ERROR;
+		xReturn = pdFALSE;
 	}
+	else{
```
htibosch, line 1114: A new-line is missing between else and `{`

AniruddhaKanhere: Good catch! thanks!

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_DNS.c
AniruddhaKanhere: Oops! Thanks for spotting that :)

PR #2262, opened 2020-07-13
FreeRTOS+TCP using same uncrustify rules as AWS and FreeRTOS kernel

htibosch:
Description
-----------
Until now, the formatting of the source code was a bit different from other AWS libraries.
From now on the same uncrustify config file will be used, e.g. [uncrustify.cfg](https://github.com/FreeRTOS/FreeRTOS/blob/lts-development/tools/uncrustify.cfg).

I applied uncrustify with mentioned config file and pushed the changed source files.

Checklist:
----------

- [ ] My code is Linted.


htibosch:
B.t.w. I had to run uncrustify 2 or 3 time before the output would be the same as the input. These were minor white-space changes, like e.g.:
~~~
-#else  /* if ( ipconfigUSE_LINKED_RX_MESSAGES != 0 ) */
+#else /* if ( ipconfigUSE_LINKED_RX_MESSAGES != 0 ) */
~~~
In the second run it removed the extra space after `#else`.

htibosch:
Since I submitted this PR, many things have changed in the repo, and it will take a lot og rebasing/merging to get it OK.
One of the good things that will changes is that for CBMC, the patch files won't be used anymore for FreeRTOS+TCP.  These were used to remove certain keywords like "static".
Once these changes are ready, I will create a new PR that applies the uncrustify rules.
Now closing this PR.

PR #2268, opened 2020-07-13
PR: Sync managed MQTT code from csdk Development Branch and header Changes for Shim

gshvang:
Description
-----------

Checklist:
----------

- [ ] My code is Linted.


libraries/c_sdk/standard/mqtt/include/iot_mqtt.h
```diff
@@ -1,5 +1,5 @@
 /*
- * FreeRTOS MQTT V2.1.1
+ * FreeRTOS MQTT V2.1.1 SHIM
```
leegeth, line 2: Since the Shim is getting introduced for the first time, I think it will be of version 1.0.0. Also, since this PR is about syncing changes, I would have made it a different small PR to handle just that.

PR #2270, opened 2020-07-14
Correct error reporting in demo

AniruddhaKanhere:
Description
-----------
The TCP echo demo erroneously reported success even when the connection had failed. This PR rectifies the error with the demo returning success/failure depending on the actual statistics of the connections.
The user can choose the threshold for the number of failures before the demo will be deemed to have failed.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/demos/tcp/aws_tcp_echo_client_single_task.c
+++ b/demos/tcp/aws_tcp_echo_client_single_task.c
@@ -56,7 +56,7 @@
 #include "aws_demo_config.h"
 
 /* Dimensions the buffer used to generate the task name. */
-#define echoMAX_TASK_NAME_LENGTH    8
+#define echoMAX_TASK_NAME_LENGTH        8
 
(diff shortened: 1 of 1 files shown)


demos/tcp/aws_tcp_echo_client_single_task.c
```diff
@@ -56,7 +56,16 @@
...
+#define echoMAX_TASK_NAME_LENGTH        8
+
+/* Maximum connection count. */
+#define echoMAXIMUM_CONNECTION_COUNT    10
+
+/* Maximum Loop count. */
+#define echoMAX_LOOP_COUNT              10
+
+/* The threshold to declare this demo as successful. Range: 0 - 1. */
+#define echoSUCCESS_THRESHOLD           0.95
```
gkwicker, line 68: Why is this the right value?

PR #2280, opened 2020-07-15
MISRA changes in FreeRTOS_IP.c + Misc. changes

AniruddhaKanhere:
Description
-----------
This PR adds changes related to MISRA correction in IP.c files.
Some corrections required code rework which has been done and tested Using Full_TCP test suite, Full_TCP_Extended and Full_FreeRTOS_TCP test suite.

Remaining MISRA warnings are regarding:

- Pointer conversions which are essential to map data buffers to various structs.
- Use of #includes after some c-code. These #includes are used in "packing" the structs.
- Use of Unions.

Note to reviewers: GitHub has difficulty figuring out the block and indentation changes. While reviewing please enable the option `"Hide whitespace changes"` in settings or [click here](https://github.com/aws/amazon-freertos/pull/2280/files?diff=split&w=1).
OR
Please see the attached [minimal diff](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-codeminimalcode). Use `cat ip_c.txt` to view the color coded information.
[ip_c.txt](https://github.com/aws/amazon-freertos/files/4927889/ip_c.txt)

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
+++ b/libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
@@ -1288,6 +1288,7 @@ eFrameProcessingResult_t eConsiderFrameForProcessing( const uint8_t * const pucE
 eFrameProcessingResult_t eReturn;
 const EthernetHeader_t *pxEthernetHeader;
 
+	/* Map the buffer onto Ethernet Header struct for easy access to fields. */
 	pxEthernetHeader = ipPOINTER_CAST( const EthernetHeader_t *, pucEthernetBuffer );
 
 	if( memcmp( ipLOCAL_MAC_ADDRESS, pxEthernetHeader->xDestinationAddress.ucBytes, sizeof( MACAddress_t ) ) == 0 )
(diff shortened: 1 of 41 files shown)


htibosch:
@AniruddhaKanhere, thanks again for these comments and changes. I downloaded your FreeRTOS_IP.c, in order to have a clear view on the changes.
It looks all good to me, no remarks from my side.

libraries/freertos_plus/standard/freertos_plus_tcp/source/FreeRTOS_IP.c
```diff
@@ -2104,248 +2120,251 @@ uint16_t ucVersionHeaderLength;
...
-	 	usChecksum = ipINVALID_LENGTH;
-		location = 2;
-	 	goto error_exit;
-	}
-	usLength = pxIPPacket->xIPHeader.usLength;
-	usLength = FreeRTOS_ntohs( usLength );
-	if( uxBufferLength < ( size_t ) ( ipSIZE_OF_ETH_HEADER + ( size_t ) usLength ) )
+	/* Introduce a do-while loop to allow use of break statements.
+	 * Note: MISRA prohibits use of 'goto', thus replaced with breaks. */
+	do
```
yuhui-zheng, line 2125: It's interesting that we chose ```do {...} while( ipFALSE_BOOL );``` instead of something like ``` if ( continue == pdTrue ) {...}```. Both could work if done properly, the latter looks more "linear"...

Though with do-while loop, are we missing a break at the very end of the iteration? So that if nothing goes wrong with the first iteration, this function exits? Maybe I missed?

AniruddhaKanhere: It is `do{ ... } while ( 0 );` essentially. So no need of a `break`.

PR #2300, opened 2020-07-21
Minor PKCS #11 Refactoring

lundinc2:
Minor PKCS #11 Refactoring

Description
-----------
Removed unnecessary check now that prvMbedTLS_Initialize has a static scope. Added additional guard to prevent freeing the sha256 mbedtls context if an operation was never initialized.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -528,23 +528,17 @@ static CK_RV prvMbedTLS_Initialize( void )
 
     /* See explanation in prvCheckValidSessionAndModule for this exception. */
     /* coverity[misra_c_2012_rule_10_5_violation] */
-    if( xP11Context.xIsInitialized == ( CK_BBOOL ) CK_TRUE )
-    {
-        xResult = CKR_CRYPTOKI_ALREADY_INITIALIZED;
-    }
-    else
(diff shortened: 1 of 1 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -528,23 +528,17 @@ static CK_RV prvMbedTLS_Initialize( void )
...
-        xP11Context.xSessionMutex = xSemaphoreCreateMutexStatic(
-            &xP11Context.xSessionMutexBuffer );
+    xP11Context.xSessionMutex = xSemaphoreCreateMutexStatic(
+        &xP11Context.xSessionMutexBuffer );
 
-        if( ( xP11Context.xObjectList.xMutex == NULL ) || ( xP11Context.xSessionMutex == NULL ) )
-        {
-            xResult = CKR_HOST_MEMORY;
-        }
+    if( ( xP11Context.xObjectList.xMutex == NULL ) ||
```
ravibhagavandas, line 538: since the mutex is created from static buffer, will it ever fail ?

lundinc2: Yes it should never fail. I can remove this check, but I'd prefer to leave it in again, in case the threading API gets separated into an abstraction

PR #2305, opened 2020-07-23
ble-wifi-proxy/provisioning: Receive/Handle error codes from phone

dachalco:
**Description:**
New API call `IotBleDataTransfer_GetPeerError` which reports the current 7-bit error status reported by the phone. 
There's also a new event `IOT_BLE_DATA_TRANSFER_CHANNEL_ERROR_UPDATED` which invokes user callback installed from `IotBleDataTransfer_SetCallback`. This new event is used in ble-network layer to log errors that the phone (peer) encountered, such as incorrect mqtt credentials.

There is [a companion PR](https://github.com/aws/amazon-freertos-ble-android-sdk/pull/28) in `amazon-freertos-ble-android-sdk` which adds mechanism for phone to send over these error codes. This PR does not functionally require the companion PR and existing functionality is retained without it.

Checklist:
----------

- [X ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
+++ b/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
@@ -45,7 +45,7 @@ typedef enum IotBleDataTransferChannelEvent
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_RECEIVED, /**< Event invoked when the last chunk of a large object or a small packet is received on the channel. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_SENT,     /**< Event invoked after the last chunk of data stream is sent over the channel. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_CLOSED,        /**< Event invoked when the channel is closed. */
-    IOT_BLE_DATA_TRANSFER_CHANNEL_ERROR_UPDATED  /**< Event invoked when error status bits is overwritten */
+    IOT_BLE_DATA_TRANSFER_CHANNEL_PEER_ERROR     /**< Event invoked when error status bits is overwritten */
 } IotBleDataTransferChannelEvent_t;
 
 /**
--- a/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
+++ b/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
@@ -168,4 +168,3 @@ uint8_t IotBleDataTransfer_GetPeerError( IotBleDataTransferChannel_t * pChannel 
 
 
 #endif /* IOT_BLE_DATA_TRANSFER_H */
-
--- a/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
+++ b/libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
@@ -44,7 +44,8 @@ typedef enum IotBleDataTransferChannelEvent
     IOT_BLE_DATA_TRANSFER_CHANNEL_OPENED = 0,    /**< Indicates if the channel is opened and ready to read or write data. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_RECEIVED, /**< Event invoked when the last chunk of a large object or a small packet is received on the channel. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_SENT,     /**< Event invoked after the last chunk of data stream is sent over the channel. */
-    IOT_BLE_DATA_TRANSFER_CHANNEL_CLOSED         /**< Event invoked when the channel is closed. */
+    IOT_BLE_DATA_TRANSFER_CHANNEL_CLOSED,        /**< Event invoked when the channel is closed. */
+    IOT_BLE_DATA_TRANSFER_CHANNEL_ERROR_UPDATED  /**< Event invoked when error status bits is overwritten */
 } IotBleDataTransferChannelEvent_t;
 
 /**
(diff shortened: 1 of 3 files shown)


libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
```diff
@@ -44,7 +44,8 @@ typedef enum IotBleDataTransferChannelEvent
     IOT_BLE_DATA_TRANSFER_CHANNEL_OPENED = 0,    /**< Indicates if the channel is opened and ready to read or write data. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_RECEIVED, /**< Event invoked when the last chunk of a large object or a small packet is received on the channel. */
     IOT_BLE_DATA_TRANSFER_CHANNEL_DATA_SENT,     /**< Event invoked after the last chunk of data stream is sent over the channel. */
-    IOT_BLE_DATA_TRANSFER_CHANNEL_CLOSED         /**< Event invoked when the channel is closed. */
+    IOT_BLE_DATA_TRANSFER_CHANNEL_CLOSED,        /**< Event invoked when the channel is closed. */
+    IOT_BLE_DATA_TRANSFER_CHANNEL_ERROR_UPDATED  /**< Event invoked when error status bits is overwritten */
```
ravibhagavandas, line 48: Could this be renamed as `IOT_BLE_DATA_TRANSFER_CHANNEL_PEER_ERROR` to confirm with the API name?

libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
```diff
@@ -158,4 +159,12 @@ void IotBleDataTransfer_Reset( IotBleDataTransferChannel_t * pChannel );
  */
 bool IotBleDataTransfer_Cleanup( void );
 
+/**
+ * @brief Retrieve the current error code reported by Data Transfer peer.
+ *
```
ravibhagavandas, line 164: include `@param[in]` as well.

libraries/c_sdk/standard/ble/src/services/mqtt_ble/iot_ble_data_transfer.c
```diff
@@ -202,6 +202,8 @@ struct IotBleDataTransferChannel
 
     bool isUsed;                                  /**< Flag to indicate if the channel is used. */
     bool isOpen;                                  /**< Flag to indicate if the channel is ready to send/receive data. */
+
+    uint8_t ucPeerError : 7;                      /**< 7-bit error status reported by peer */
```
ravibhagavandas, line 206: I think declaring a bit field does not add additional value here. should we declare as uint8_t instead?

libraries/c_sdk/standard/ble/src/services/mqtt_ble/iot_network_ble.c
```diff
@@ -76,6 +84,34 @@ static void _callback( IotBleDataTransferChannelEvent_t event,
             pBleConnection->pCallback( pBleConnection, pBleConnection->pContext );
             break;
 
+        case IOT_BLE_DATA_TRANSFER_CHANNEL_ERROR_UPDATED:
+
+            /* Currently only used for logging, but could short-circuit failure here.
+             * At time of writing, such error events could occur but operations would only fail LATER,
+             * after timeout that resulted from this error */
+            ucPeerError = ( BleNetworkError_t ) IotBleDataTransfer_GetPeerError( pChannel );
```
ravibhagavandas, line 92: By making ucPeerError part of _bleNetworkConnection_t,  the `IotNetworkBle_Create()` code can check for the errors and log them appropriately by not failing with a timeout.

libraries/c_sdk/standard/ble/src/services/mqtt_ble/iot_ble_data_transfer.c
dachalco: The peer error is limited to 7-bits as it's using the previously unused upper 7-bits of CONTROL gatt char -- LSB being the service `ready` bit. I think it helps with readability

libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
dachalco: added in latest commit

libraries/c_sdk/standard/ble/include/iot_ble_data_transfer.h
dachalco: added in recent commit

libraries/c_sdk/standard/ble/src/services/mqtt_ble/iot_network_ble.c
dachalco: **Summary:** 
Due to interdependencies, covering all short-circuiting failures requires code changes at higher and lower level APIs with respect to `iot_network_ble.c`. Deferring short-circuiting to later PR.

**Details**
I have code that will short circuit fail IotNetworkBle_Create() if a `BLE_PEER_COGNITO_AUTH_FAIL` was received, but it doesn't cover all short-circuitable errors cases. 

While the behavior of `IotNetworkBle_Create()` is largely defined at the `iot_network_ble.c` level, thereby limiting required short-circuiting code changes to `iot_network_ble.c`, the `IotNetworkBle_Send/Receive()` calls are mostly defined by `IotBleDataTransfer_Send` from lower-level API `iot_ble_data_transfer.h`. At this lower level the error codes, similar to the tx/rx data, are arbitrary values that are handled by the higher level implementation like `iot_network_ble.c`. By itself it cannot determine whether a received error should short-circuit and cutoff a connection. 

Moreover, the callers _above_ `iot_network_ble.c` that use `IotNetworkBle_Send/Receive` have their own behavior and receive callbacks which likely need come into play too.

dachalco:
Newer features were requested and added in the last commit `31147f` that need more testing. Closing for now to revisit later.

PR #2308, opened 2020-07-24
Add all common-io peripherals to CMake build system

lundinc2:
Add all common-io peripherals to CMake build system

Description
-----------
Since the CMake now only tries to link supported peripheral symbols, all the tests and header files can be added to the build.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/iot_test_common_io.c
+++ b/libraries/abstractions/common_io/test/iot_test_common_io.c
@@ -57,6 +57,14 @@ TEST_GROUP_RUNNER( Common_IO )
     size_t i = 0;
 
     /* These already used loop back tests which require minimum of two pins */
+    #ifdef IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED
+        for( i = 0; i < IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED; i++ )
+        {
+            SET_TEST_IOT_PERFCOUNTER_CONFIG( i );
+            RUN_TEST_GROUP( TEST_IOT_PERFCOUNTER );
+        }
+    #endif
+
     #ifdef IOT_TEST_COMMON_IO_GPIO_SUPPORTED
         for( i = 1; i < IOT_TEST_COMMON_IO_GPIO_SUPPORTED; i++ )
         {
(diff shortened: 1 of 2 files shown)


lundinc2:
@osloapps-max tagging you in-case you want to weigh in on this PR before I merge it. Likely will merge it early Monday.

libraries/abstractions/common_io/test/iot_test_common_io.c
```diff
@@ -88,4 +88,116 @@ TEST_GROUP_RUNNER( Common_IO )
...
+
+    #ifdef IOT_TEST_COMMON_IO_FLASH_SUPPORTED
+        for( i = 0; i < IOT_TEST_COMMON_IO_FLASH_SUPPORTED; i++ )
+        {
+            SET_TEST_IOT_FLASH_CONFIG( i );
+            RUN_TEST_GROUP( TEST_IOT_FLASH );
+        }
+    #endif
+
+    #ifdef IOT_TEST_COMMON_IO_TIMER_SUPPORTED
```
NotCosyMax, line 132: Should the Timer"test case be located before the Flash tests? Reason for the question is that the Flash tests depend on the Timer APIs which means it make sense (for me)  to run these tests prior to the Flash ones (assuming you know about this dependency it might help potential debugging).

osloapps-max, line 132: Should the Timer"test case be located before the Flash tests? Reason for the question is that the Flash tests depend on the Timer APIs which means it make sense (for me) to run these tests prior to the Flash ones (assuming you know about this dependency it might help potential debugging).

libraries/abstractions/common_io/test/iot_test_common_io.c
```diff
@@ -88,4 +88,116 @@ TEST_GROUP_RUNNER( Common_IO )
...
+
+    #ifdef IOT_TEST_COMMON_IO_RESET_SUPPORTED
+        for( i = 0; i < IOT_TEST_COMMON_IO_RESET_SUPPORTED; i++ )
+        {
+            SET_TEST_IOT_RESET_CONFIG( i );
+            RUN_TEST_GROUP( TEST_IOT_RESET );
+        }
+    #endif
+
+    #ifdef IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED
```
osloapps-max, line 116: Same as for Timer and Flash. Should this test be located before the GPIO tests due to the potential dependency?

libraries/abstractions/common_io/test/iot_test_common_io.c
lundinc2: Good catch, I'll move this

libraries/abstractions/common_io/test/iot_test_common_io.c
lundinc2: Thanks for catching this as well, will apply this feedback

libraries/abstractions/common_io/test/iot_test_common_io.c
```diff
@@ -57,6 +57,14 @@ TEST_GROUP_RUNNER( Common_IO )
     size_t i = 0;
 
     /* These already used loop back tests which require minimum of two pins */
```
dachalco, line 59: This comment only applies to GPIO tests, Should go inside or above it's `#ifdef IOT_TEST_COMMON_IO_GPIO_SUPPORTED` block

PR #2311, opened 2020-07-24
Fix non-conformance to mbed TLS pk.h API when hashing a signature.

lundinc2:
Fix non-conformance to mbed TLS pk.h API when hashing a signature.

Description
-----------
The mbed TLS specifies that when using RSA, one should use no hashing algorithm and when using ECDSA one should use a hashing algorithm.

Checklist:
----------

- [ x] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -3652,7 +3652,6 @@ CK_DECLARE_FUNCTION( CK_RV, C_Sign )( CK_SESSION_HANDLE hSession,
         {
             xSignatureLength = pkcs11RSA_2048_SIGNATURE_LENGTH;
             xExpectedInputLength = pkcs11RSA_SIGNATURE_INPUT_LENGTH;
-            xHashType = MBEDTLS_MD_NONE;
         }
         else if( pxSessionObj->xOperationSignMechanism == CKM_ECDSA )
         {
--- a/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
+++ b/libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
@@ -3637,6 +3637,7 @@ CK_DECLARE_FUNCTION( CK_RV, C_Sign )( CK_SESSION_HANDLE hSession,
     /* 8 bytes added to hold ASN.1 encoding information. */
     uint8_t ecSignature[ pkcs11ECDSA_P256_SIGNATURE_LENGTH + 8 ];
     int32_t lMbedTLSResult;
+    mbedtls_md_type_t xHashType;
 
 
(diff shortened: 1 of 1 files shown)


libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3689,14 +3689,40 @@ CK_DECLARE_FUNCTION( CK_RV, C_Sign )( CK_SESSION_HANDLE hSession,
...
-                                                      pxSignatureBuffer,
-                                                      &xExpectedInputLength,
-                                                      mbedtls_ctr_drbg_random,
-                                                      &xP11Context.xMbedDrbgCtx );
+                    /* Per mbed TLS documentation, if using RSA, md_alg should
+                     * be MBEDTLS_MD_NONE. If ECDSA, md_alg should never be
+                     * MBEDTLS_MD_NONE. SHA-256 will be used for ECDSA for
+                     * consistency with the rest of the port.
+                     */
+                    if( pxSessionObj->xOperationSignMechanism == CKM_RSA_PKCS )
```
gkwicker, line 3697: The logic in lines 3698-3707 and 3709-3718 is identical except for the value of the second parameter.  Also, the code as written falls through at line 3721 rather than asserting as it probably should.  This would be much easier to read as:

```c
/*
* NOTE: assert() here rather than allowing a silent fall-through; this code
* cannot handle any other signing type.
*/
assert( pxSessionObj->xOperationSignMechanism == CKM_ECDSA ||
        pxSessionObj->xOperationSignMechanism == CKM_RSA_PKCS );

uint32_t ulMessageDigestType = 
    pxSessionObj->xOperationSignMechanism == CKM_ECDSA ?
        MBEDTLS_MD_SHA256 : MBEDTLS_MD_NONE;

lMbedTLSResult = mbedtls_pk_sign( &pxSessionObj->xSignKey,
        ulMessageDigestType,
        .
        .
        .
```

lundinc2: Good callout, I condensed the code to more efficient form based on your feedback

lundinc2:
Sorry @gkwicker the make project for ESP32 did not like the uninitialized variable.

libraries/abstractions/pkcs11/mbedtls/iot_pkcs11_mbedtls.c
```diff
@@ -3651,12 +3652,14 @@ CK_DECLARE_FUNCTION( CK_RV, C_Sign )( CK_SESSION_HANDLE hSession,
         {
             xSignatureLength = pkcs11RSA_2048_SIGNATURE_LENGTH;
             xExpectedInputLength = pkcs11RSA_SIGNATURE_INPUT_LENGTH;
+            xHashType = MBEDTLS_MD_NONE;
```
gkwicker, line 3655: Already set at line 3640, looks like this can be deleted.

PR #2320, opened 2020-07-28
CommonIO Flash: nrf52840-dk

dachalco:
**Description:**
Implementation of iot_flash.h

Async APIs are not supported in this PR. The memory mapped NVMC peripheral does not provide a mechanism for generating interrupts. NVMC is shared with CPU. CPU is halted if it's waiting to use NVMC. The _appearance_ of async can be done by launching a sync API in an RTOS thread.

Flash access protection is not supported in this PR, but I do have a working implementation in a separate branch. Excluded for now because testing is trickier and incomplete. For this board, flash protection violations cause hardfaults. There's no way to test via injected failure without overriding `HardFault_Handler` and restoring execution for test to catch failure. Need further discussion with team.

**Notes:**
- For this board, hardfaults are generated when certain flash violations or flash failures occur. These tests require a working flash page that is within life-span. 
 - Some of the CommonIO Flash tests erase/write pages thereby depleting flash life. Flash pages of this board are functional for a minimum of 10000 erase cycles.

**Testing:**
1) Set `testrunnerFULL_COMMON_IO_ENABLED = 1` in `aws_test_runner_config.h`
2) Set `IOT_TEST_COMMON_IO_FLASH_SUPPORTED = 1` in `iot_test_common_io_config.h`
3) Build, flash, and observe `aws_tests` SES project

Checklist:
----------

- [X] My code is Linted.

### DIFF
--- a/libraries/abstractions/common_io/test/iot_test_common_io.c
+++ b/libraries/abstractions/common_io/test/iot_test_common_io.c
@@ -56,14 +56,6 @@ TEST_GROUP_RUNNER( Common_IO )
 {
     size_t i = 0;
 
-    #ifdef IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED
-        for( i = 0; i < IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED; i++ )
-        {
-            SET_TEST_IOT_PERFCOUNTER_CONFIG( i );
-            RUN_TEST_GROUP( TEST_IOT_PERFCOUNTER );
-        }
-    #endif
-
     /* These already used loop back tests which require minimum of two pins */
     #ifdef IOT_TEST_COMMON_IO_GPIO_SUPPORTED
         for( i = 1; i < IOT_TEST_COMMON_IO_GPIO_SUPPORTED; i++ )
--- a/libraries/abstractions/common_io/test/iot_test_common_io.c
+++ b/libraries/abstractions/common_io/test/iot_test_common_io.c
@@ -56,6 +56,14 @@ TEST_GROUP_RUNNER( Common_IO )
 {
     size_t i = 0;
 
+    #ifdef IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED
+        for( i = 0; i < IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED; i++ )
+        {
+            SET_TEST_IOT_PERFCOUNTER_CONFIG( i );
+            RUN_TEST_GROUP( TEST_IOT_PERFCOUNTER );
+        }
+    #endif
+
     /* These already used loop back tests which require minimum of two pins */
     #ifdef IOT_TEST_COMMON_IO_GPIO_SUPPORTED
         for( i = 1; i < IOT_TEST_COMMON_IO_GPIO_SUPPORTED; i++ )
--- a/libraries/abstractions/common_io/test/test_iot_flash.c
+++ b/libraries/abstractions/common_io/test/test_iot_flash.c
@@ -340,7 +340,6 @@ TEST( TEST_IOT_FLASH, AFQP_IotFlashReadInfo )
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulPageSize & ( pxFlashInfo->ulPageSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulSectorSize & ( pxFlashInfo->ulSectorSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulBlockSize & ( pxFlashInfo->ulBlockSize - 1 ) );
-        TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulFlashSize & ( pxFlashInfo->ulFlashSize - 1 ) );
     }
 
     /* Close flash handle */
--- a/libraries/abstractions/common_io/test/test_iot_flash.c
+++ b/libraries/abstractions/common_io/test/test_iot_flash.c
@@ -340,6 +340,7 @@ TEST( TEST_IOT_FLASH, AFQP_IotFlashReadInfo )
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulPageSize & ( pxFlashInfo->ulPageSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulSectorSize & ( pxFlashInfo->ulSectorSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulBlockSize & ( pxFlashInfo->ulBlockSize - 1 ) );
+        TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulFlashSize & ( pxFlashInfo->ulFlashSize - 1 ) );
     }
 
     /* Close flash handle */
--- a/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
+++ b/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
@@ -1,0 +1,37 @@
+/*
+ * FreeRTOS
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://aws.amazon.com/freertos
+ * http://www.FreeRTOS.org
+ */
+
+/**
+ * @file   iot_flash_config.h
+ * @brief  Additional settings for GPIO, such as CommonIO-to-Board pin mapping.
+ */
+
+#ifndef _AWS_COMMON_IO_FLASH_CONFIG_H_
+#define _AWS_COMMON_IO_FLASH_CONFIG_H_
+
+/* Set defaults which are not overridden */
+#include "iot_flash_config_defaults.h"
+
+#endif /* ifndef _AWS_COMMON_IO_FLASH_CONFIG_H_ */
--- a/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
+++ b/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
@@ -30,6 +30,10 @@
 #ifndef _AWS_COMMON_IO_FLASH_CONFIG_H_
 #define _AWS_COMMON_IO_FLASH_CONFIG_H_
 
+
+#define IOT_LOG_LEVEL_IOT_FLASH    IOT_LOG_NONE
+
+
 /* Set defaults which are not overridden */
 #include "iot_flash_config_defaults.h"
 
--- a/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
+++ b/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
@@ -25,7 +25,6 @@
 
 /**
  * @file   iot_flash_config.h
- * @brief  Additional settings for GPIO, such as CommonIO-to-Board pin mapping.
  */
 
 #ifndef _AWS_COMMON_IO_FLASH_CONFIG_H_
--- a/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_test_common_io_config.h
+++ b/vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_test_common_io_config.h
@@ -33,28 +33,31 @@
 #include "iot_spi.h"
 #include "nrf_drv_spi.h"
 
+/*------------------------FLASH-------------------------------*/
+#define IOT_TEST_COMMON_IO_FLASH_SUPPORTED    0
+
 /*------------------------GPIO-------------------------------*/
-#define IOT_TEST_COMMON_IO_GPIO_SUPPORTED    0
+#define IOT_TEST_COMMON_IO_GPIO_SUPPORTED     0
 
 /*------------------------UART-------------------------------*/
-
-/* UART is supported. */
-/*#define IOT_TEST_COMMON_IO_UART_SUPPORTED                    1 */
+#define IOT_TEST_COMMON_IO_UART_SUPPORTED     0
 
 /*------------------------I2C-------------------------------*/
-
-/* I2C is supported. */
-/*#define IOT_TEST_COMMON_IO_I2C_SUPPORTED                     1 */
+#define IOT_TEST_COMMON_IO_I2C_SUPPORTED      0
 
 /*------------------------SPI-------------------------------*/
-#define IOT_TEST_COMMON_IO_SPI_SUPPORTED    0
+#define IOT_TEST_COMMON_IO_SPI_SUPPORTED      0
 
-/* #define IOT_TEST_COMMON_IO_SPI_LOOPBACK        1 */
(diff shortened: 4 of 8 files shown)


libraries/abstractions/common_io/test/test_iot_flash.c
```diff
@@ -340,7 +340,6 @@ TEST( TEST_IOT_FLASH, AFQP_IotFlashReadInfo )
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulPageSize & ( pxFlashInfo->ulPageSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulSectorSize & ( pxFlashInfo->ulSectorSize - 1 ) );
         TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulBlockSize & ( pxFlashInfo->ulBlockSize - 1 ) );
-        TEST_ASSERT_EQUAL( 0, pxFlashInfo->ulFlashSize & ( pxFlashInfo->ulFlashSize - 1 ) );
```
lundinc2, line 343: Can you provide some background on removing this?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+static inline bool prvIsAligned4K( uint32_t ulAddr )
+{
+    return ( ulAddr & FLASH_SECTOR_MASK_4K ) == 0;
+}
+
+static inline void prvWaitUntilFlashReady( nrf_fstorage_t const * p_fstorage )
+{
+    while( nrf_fstorage_is_busy( p_fstorage ) )
+    {
+        vTaskDelay( pdMS_TO_TICKS( 1 ) );
```
lundinc2, line 126: Haven't read the whole PR, but maybe should return IOT_FLASH_DEVICE_BUSY to the application, and defer waiting to the next level?

vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_test_common_io_config.h
```diff
@@ -33,28 +33,31 @@
 #include "iot_spi.h"
 #include "nrf_drv_spi.h"
 
+/*------------------------FLASH-------------------------------*/
+#define IOT_TEST_COMMON_IO_FLASH_SUPPORTED    0
```
n9wxu, line 37: I think these test generation flags is a bad idea and makes the test suite much more complex that required.

vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @file   iot_flash_config.h
+ */
+
+#ifndef _AWS_COMMON_IO_FLASH_CONFIG_H_
+#define _AWS_COMMON_IO_FLASH_CONFIG_H_
+
+/* Set defaults which are not overridden */
+#include "iot_flash_config_defaults.h"
```
n9wxu, line 34: I don't like header files that simply include 1 other header file.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+    if( xFlashDesc.xCallback != NULL )
+    {
+        xFlashDesc.xCallback( xStatus, xFlashDesc.pvUserContext );
+    }
+}
+
+static void prvGetFlashInfo( void )
+{
+    uint32_t const bootloader_addr = NRF_UICR->NRFFW[ 0 ];
+    uint32_t const page_sz = NRF_FICR->CODEPAGESIZE;
```
lundinc2, line 160: These variables go against the coding style, I think common-io uses the FreeRTOS coding style, instead of the C-SDK

vendors/nordic/boards/nrf52840-dk/ports/common_io/config/iot_flash_config_defaults.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef _AWS_COMMON_IO_FLASH_CONFIG_DEFAULTS_H_
+#define _AWS_COMMON_IO_FLASH_CONFIG_DEFAULTS_H_
+
+#ifndef IOT_LOG_LEVEL_IOT_FLASH
+    #define IOT_LOG_LEVEL_IOT_FLASH    IOT_LOG_NONE
+#endif
+
+/* User application base address. Note, iot_flash.h API does not check against this */
+#ifndef IOT_COMMON_IO_FLASH_BASE_ADDRESS
+    #define IOT_COMMON_IO_FLASH_BASE_ADDRESS    0x00000000
```
n9wxu, line 40: Is this a vendor parameter?  Should this be in some sort of vendor flash defaults to separate vendor defined constants from FreeRTOS constants and User constants?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+    return IOT_FLASH_FUNCTION_NOT_SUPPORTED;
+}
+
+int32_t iot_flash_close( IotFlashHandle_t const pxFlashHandle )
+{
+    int32_t lReturnCode = IOT_FLASH_INVALID_VALUE;
+
+    if( prvIsOperableHandle( pxFlashHandle ) )
+    {
+        if( NRF_SUCCESS == nrf_fstorage_uninit( &iot_fstorage, NULL ) )
```
lundinc2, line 478: This is a nit but it can be nice to separate the function result into a variable outside of the if statement to make debugging easier

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+
+NRF_FSTORAGE_DEF( nrf_fstorage_t iot_fstorage ) =
+{
+    /* Set a handler for fstorage events. */
+    .evt_handler = prvFlashEventHandler,
+};
+
+typedef enum
+{
+    IOT_FLASH_CLOSED = 0u,
```
n9wxu, line 82: Are the values for these enums important?  In general, I feel we should not rely upon the value of an enum.  I have seen to many abuses like flash_state = IOT_FLASH_CLOSED | IOT_FLASH_OPENED;

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_test_common_io_internal.c
```diff
@@ -29,6 +29,18 @@
...
+/* FLASH */
+    void SET_TEST_IOT_FLASH_CONFIG( int testSet )
+    {
+        /* This value depends on linked flash usage. At time of writing, test project ended @ 0x49bc3
+         * Some tests require this to be page-aligned
+         */
+        ultestIotFlashStartOffset = 0x75000;
+    }
+
+#endif /* if defined( IOT_TEST_COMMON_IO_SPI_SUPPORTED ) && ( IOT_TEST_COMMON_IO_SPI_SUPPORTED >= 1 ) */
```
lundinc2, line 42: Copy paste error

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+/* @brief flash context */
+typedef struct IotFlashDescriptor
+{
+    IotFlashInfo_t xFlashInfo;    /* flash info structure */
+    IotFlashCallback_t xCallback; /* callback */
+    void * pvUserContext;         /* user context to provide in callback */
+    uint8_t ucState;
+} IotFlashDescriptor_t;
+
+#define DEFAULT_FLASH_DESC                          \
```
n9wxu, line 95: Is DEFAULT_FLASH_DESC used anywhere besides line 112?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
n9wxu: I agree.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+        pxFlashHandle->xCallback = xCallback;
+        pxFlashHandle->pvUserContext = pvUserContext;
+    }
+    else
+    {
+        IotLogWarn( "%s: Null or closed IotFlashHandle_t(0x%x)", __func__, pxFlashHandle );
+    }
+}
+
+int32_t iot_flash_ioctl( IotFlashHandle_t const pxFlashHandle,
```
n9wxu, line 259: I truly hate the ioctl pattern.  The cyclomatic complexity goes up and the clarity goes down.
Why can't we simply have the functions we mean to have?  Is there a limit to our function count?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
```diff
@@ -0,0 +1,490 @@
...
+         */
+        size_t write_size = 4 * ( ( xBytes + 3 ) / 4 ); /* pad size to multiple of a word */
+        uint8_t write_buf[ write_size ] __attribute__( ( aligned( 4 ) ) );
+        memcpy( write_buf, pvBuffer, xBytes );
+        memset( write_buf + xBytes, FLASH_PADDING_BYTE, write_size - xBytes );
+
+        /* Write and check status*/
+        ret_code_t xReturnCode_NRF = nrf_fstorage_write( &iot_fstorage, ulAddress, write_buf, write_size, NULL );
+
+        switch( xReturnCode_NRF )
```
n9wxu, line 383: This error handler seems to be repeated in many places.  Do we think that a more "mature" version of the library will evolve to have different error responses for each of these functions?

libraries/abstractions/common_io/test/test_iot_flash.c
dachalco: Glad you brought this up. I'm second guessing this change.

The idea was that `pxFlashInfo->ulFlashSize` is limited to _useable_ flash. For this board, our bootloader is stored at the end of total flash region, which I deemed _unuseuable_, thereby excluding some pages from the total of 1 MB (2^20) flash.

Technically, it's also not required that any of these are powers of 2. Though practically, it's a de-facto standard.
I'm thinking I'll re-introduce this condition and adjust value of `ulFlashSize` to include bootloader flash region. Thoughts?

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_test_common_io_internal.c
```diff
@@ -29,6 +29,18 @@
...
 #include "iot_test_common_io_internal.h"
 
+#if defined( IOT_TEST_COMMON_IO_FLASH_SUPPORTED ) && ( IOT_TEST_COMMON_IO_FLASH_SUPPORTED >= 1 )
+/* FLASH */
+    void SET_TEST_IOT_FLASH_CONFIG( int testSet )
+    {
+        /* This value depends on linked flash usage. At time of writing, test project ended @ 0x49bc3
+         * Some tests require this to be page-aligned
+         */
+        ultestIotFlashStartOffset = 0x75000;
```
n9wxu, line 39: Is this magic number specific to this linker, compiler, device, feature set in the test?
Can we derive this magic number from the linker script?

libraries/abstractions/common_io/test/test_iot_flash.c
lundinc2: That sounds reasonable to me, reading the API it doesn't seem to differentiate. I'm unfamiliar with the Nordic APIs, but alternatively the different flash sections could be moved to different handles, so the bootloader can be managed separately. Curious what you think about this, seems the API (to me) is flexible in this regard, but will have a better idea regarding this

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: By 'next level' do you mean above `iot_flash.h`? If so, and the concern is to not delay/reschedule the current task, I see where you're coming from.  However the NVMC does not have an interrupt based interface so if `vTaskDelay` can't be used then we're limited to busy-waits or delay via hardware timer -- all the while still having to block parent API call as its sync. `prvWaitUntilFlashReady` is only called in the `*_sync` functions which, by API definition, can not return until the flash operation is complete.

Or by next level, do you mean having `prvWaitUntilFlashReady` return `IOT_FLASH_DEVICE_BUSY`?

lundinc2: Yes, if the API returns busy to the user, then we defer the decision to them. This allows for more flexibility like yielding the thread or choosing how long to delay. In the case of a library sitting on this one, it can bubble up the BUSY, and make for a more flexible library

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: The handler is different for read_sync, though it looks similar. It is currently the same for write_sync and erase_sync. 
Exactly, left it as such in case we to further customize handler response for each case

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: Honoring what the existing API (`iot_flash.h`) expects. The `common_io/*.h` APIs were the result of extensive discussion. @ejacobus can provide details on those discussions and resulting decisions. 

From my observation, the `ioctls` are useful for drawing a line between "standard" and "non-standard" common_io functions.

Since the `*.h` APIs predate the `*.c` implementations, any non-additive API changes could also mess with customer implementations of `iot_flash.c`, if they wrote one.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_test_common_io_internal.c
dachalco: Specific to our `aws_tests` architecture for this board. The Code Regions are "free-game", though our code save  starts at Code Region 0 and ended at 0x49bc3 -- for this PR's `aws_tests`

I agree that a symbol like `__user_flash_start__` and `__user_flash_end__` would be a useful improvement that should be made for all of our boards. Currently there is no such shared mechanism.
I've added a separate task to research how we can address this across all boards.

<p align="center">

</p>

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: The value doesn't matter. It was specified to be explicit and to reflect similar state values used by other CommonIO drivers

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: Thanks. Addressed in latest commit

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: That's reasonable. Added in latest commit

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_test_common_io_internal.c
dachalco: Thanks. Addressed these comments in latest commit. 

Decided to remove these end-guard comments as their related code blocks were small, so these comments were just redundant and unecessary visual clutter.

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: If you search in the file you'll see all the usages. It's used at descriptor open and close

vendors/nordic/boards/nrf52840-dk/aws_tests/config_files/iot_flash_config.h
dachalco: Added a logging macro so it's not emtpy. This `*_config.h` pattern is common among various modules. Some modules have hard dependencies on such config files, however this flash module does not at the moment.

There's a number of potential enhancements that could be made to our flash layout across all boards, and when that time comes, this stub may become more useful.

Currently, it primarily serves as a stub and to reinforce our `*_config.h` pattern

vendors/nordic/boards/nrf52840-dk/ports/common_io/iot_flash.c
dachalco: I think that would work well for `async`, but per the sync API:

[`iot_flash.h#317`](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/include/iot_flash.h#L317):
```
 *          This is a blocking operation and waits until the number of bytes are written before returning.
```
[`iot_flash.h#332`](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/common_io/include/iot_flash.h#L332):
```
IOT_FLASH_DEVICE_BUSY if another asynchronous operation is currently being executed.
```
however async is currently not supported.

libraries/abstractions/common_io/test/test_iot_flash.c
dachalco: Added this assertion in latest commit

libraries/abstractions/common_io/test/iot_test_common_io.c
```diff
@@ -56,14 +56,6 @@ TEST_GROUP_RUNNER( Common_IO )
 {
     size_t i = 0;
 
-    #ifdef IOT_TEST_COMMON_IO_PERFCOUNTER_SUPPORTED
```
lundinc2, line 59: This shouldn't get removed

dachalco: Oops. Merge mistake during rebase. Fixed in latest commit. Thanks for catching

PR #2326, opened 2020-08-03
PKCS #11 Doxygen

lundinc2:
PKCS #11 Doxygen

Description
-----------
Fix macros not being included in the doxygen. Add dependency graphs. Various formatting fixes

Checklist:
----------

- [ x] My code is Linted.


libraries/freertos_plus/standard/pkcs11/include/iot_pkcs11.h
```diff
@@ -119,32 +139,26 @@
...
  */
 #define pkcs11DER_ENCODED_OID_P256           { 0x06, 0x08, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x03, 0x01, 0x07 }
 
-
-/**
- * @brief Set to 1 if importing private keys is supported.
- *
- * If private key import is not supported, this value should be defined 0 in aws_pkcs11_config.h
- */
-#ifndef pkcs11configIMPORT_PRIVATE_KEYS_SUPPORTED
```
ravibhagavandas, line 143: Seems the default is changed now. Is it intentional?

lundinc2: Yes, I felt like this might lead to some non-obvious behavior. Also this configuration is usually created when doing a qualification and not really touched afterwards, so it's being in the porting guide should be enough

lundinc2: Going to revert it for now, and separate it into a different PR, so this one doesn't have any logic changes

PR #2332, opened 2020-08-05
Send Ping request using MQTT_PING.

gshvang:
Implementing PINGREQ using Managed MQTT API.


libraries/c_sdk/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -806,43 +807,6 @@ void _IotMqtt_ProcessKeepAlive( IotTaskPool_t pTaskPool,
...
-                         IotTaskPool_strerror( taskPoolStatus ) );
-
-            status = false;
-        }
-    }
-    else
-    {
-        EMPTY_ELSE_MARKER;
-    }
-
```
leegeth, line 845: This entire block cannot be removed as that is how the rescehduling happens

PR #2349, opened 2020-08-11
NimBLE AFR: Accommodate 128bit UUID type for comparison with standard CCCD(0x2902)

prasad-alatkar:
Characteristic descriptors are internally added by NimBLE stack, so to avoid duplication we avoid adding descriptors (16 bit UUID = 0x2902) to GATT database. The problem may come when the caller to `prvAddServiceBlob` provides CCCD with 128bit UUID type and standard base UUID. Ideally this should be avoided, but to avoid duplication this PR intends to accommodate 128 bit UUID comparison as well.

Description
-----------

- Add `BLE_GATT_DSC_CLT_CFG_UUID128` macro to convert 16 bit CCCD UUID (`0x2902`) to 128 bit UUID.
- Compare characteristic descriptor UUID with `BLE_GATT_DSC_CLT_CFG_UUID128`.
Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -43,6 +43,8 @@
 #define APP_ID          0
 #define MAX_SERVICES    20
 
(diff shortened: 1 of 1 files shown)


prasad-alatkar:
@ravibhagavandas PTAL (CC: @dhrishi @mahavirj )

PR #2358, opened 2020-08-13
Fix public key deletion and PKCS #11 tests on Optiga X.

lundinc2:
Fix public key deletion and PKCS #11 tests on Optiga X.

Description
-----------
The newly refactored tests were overwriting and deleting the public key stored on the device. Need to do further tests to ensure this is the correct solution, but this will fix the "real" public key being deleted, as well as making the PKCS #11 tests pass.

Checklist:
----------

- [x ] My code is Linted.

### DIFF
--- a/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
+++ b/libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
@@ -1834,9 +1834,9 @@ TEST( Full_PKCS11_EC, AFQP_GenerateKeyPair )
     /* mbedTLS structures for verification. */
     uint8_t ucSecp256r1Oid[] = pkcs11DER_ENCODED_OID_P256; /*"\x06\x08" MBEDTLS_OID_EC_GRP_SECP256R1; */
 
-    /* TODO: We should be destroying credentials first, just in case. Currently a port cannot do this, but once that is complete this will need to be updated. 
+    /* TODO: We should be destroying credentials first, just in case. Currently a port cannot do this, but once that is complete this will need to be updated.
      * xResult = prvDestroyTestCredentials();
-     * TEST_ASSERT_EQUAL_MESSAGE( CKR_OK, xResult, "Failed to destroy credentials before Generating Key Pair" ); 
+     * TEST_ASSERT_EQUAL_MESSAGE( CKR_OK, xResult, "Failed to destroy credentials before Generating Key Pair" );
      * */
     xCurrentCredentials = eNone;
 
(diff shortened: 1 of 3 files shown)


libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -1834,8 +1834,10 @@ TEST( Full_PKCS11_EC, AFQP_GenerateKeyPair )
     /* mbedTLS structures for verification. */
     uint8_t ucSecp256r1Oid[] = pkcs11DER_ENCODED_OID_P256; /*"\x06\x08" MBEDTLS_OID_EC_GRP_SECP256R1; */
 
-    xResult = prvDestroyTestCredentials();
-    TEST_ASSERT_EQUAL_MESSAGE( CKR_OK, xResult, "Failed to destroy credentials before Generating Key Pair" );
+    /* TODO: We should be destroying credentials first, just in case. Currently a port cannot do this, but once that is complete this will need to be updated.
```
gkwicker, line 1837: Why is this important?  Is `xProvisionGenerateKeyPairEC()` not guaranteed to work?  If it can fail and this can only be recovered by destroying credentials, and the pkcs11 abstraction does not have this operation – please create a "feature request" issue to add it in this repo with a link to this PR, and include a link to the feature request in this `TODO` comment.  This will help us ensure that this gets `TODONE`. :-)

If `prvDestroyTestCredentials()` cannot fail in a complete/working port, then let's delete lines 1837-1840 rather than leave a `TODO` in here for something we're not certain is actually a problem.

libraries/abstractions/pkcs11/test/iot_test_pkcs11.c
```diff
@@ -2414,7 +2410,7 @@ TEST( Full_PKCS11_EC, AFQP_GetAttributeValueMultiThread )
         xGlobalTaskParams[ xTaskNumber ].pvTaskData = &xSessionHandle[ xTaskNumber ];
     }
 
-    prvProvisionEcTestCredentials( &xPrivateKey, &xCertificate, &xPublicKey );
+    prvFindObjectTest( &xPrivateKey, &xCertificate, &xPublicKey );
```
ravibhagavandas, line 2413: Is there a way to fail with a valid message, if the credentials are not found.

lundinc2: There are test assert messages in this helper function. It is tested first in AFQP_FindObjects, and then the code is reused in order to reuse objects between the tests. This was a miss in my earlier refactoring, as the goal is to only provision objects in the test cases that are testing that logic.

PR #2367, opened 2020-08-14
PsoC64 re-provisioning changes

solusrex:
PSoC64: remove re-prov_helper dependency on fw-loader 
PSoC64: delete cysecuretools artifacts from AFR tree 

Description
-----------
User is expected to switch to DAPLink mode manually.
It is sufficient to instruct the user to leave in daplink
before running the script. It is much simpler than pointing
to a fw-loader utility.

CY auth keys, hsm_state.json and oem_state.json have to be
populated with cysecuretools command:
cysecuretools -t cy8ckit-064s0s2-4343w init

Checklist:
----------

- [x] My code is Linted.


mingyue86010:
> PSoC64: remove re-prov_helper dependency on fw-loader
> PSoC64: delete cysecuretools artifacts from AFR tree
> 
> ## Description
> User is expected to switch to DAPLink mode manually.
> It is sufficient to instruct the user to leave in daplink
> before running the script. It is much simpler than pointing
> to a fw-loader utility.
> 
> CY auth keys, hsm_state.json and oem_state.json have to be
> populated with cysecuretools command:
> cysecuretools -t cy8ckit-064s0s2-4343w init
> 
> ## Checklist:
> * [x]  I have tested my changes. No regression in existing tests.
> * [x]  My code is Linted.
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Hi Andrei,

The PR looks good. But since it changes the re-provisioning script, I'm wondering if the provisioning guide (https://community.cypress.com/docs/DOC-20043), which is referenced by the getting start guide, need to be updated accordingly. If so, can you follow up with that? 

Thank you!

Ming.

RaymondNgun:
Hi @mingyue86010, no change is required in the guide.  The guide doesn't reference the reprov_helper.py that is changed here.  Also, the guide already has instructions to run "cysecuretools init" which will populate the files that is being deleted here.

PR #2372, opened 2020-08-17
OTA E2E test fix

tgsong:
Description
-----------

- Some test cases are failing because we need to perform some actions before OTA is complete, but it's blocked until we finish building second image. This PR changed to save the first image, and immediately build the second image before we start the OTA session.
- `OtaTest2UpdatesCancel1st` is missing in the all tests list.
- `junit-xml` is not working properly to generate the test report, replaced `junit-xml` with `junitparser`.

Checklist:
----------

- [ ] My code is Linted.


tools/ota_e2e_tests/aws_ota_test/aws_ota_test_case_2_updates_cancel_1st.py
```diff
@@ -74,6 +81,9 @@ def run(self):
         if exec_status == 'QUEUED':
             return OtaTestResult(testName=self._name, result=OtaTestResult.ERROR,
                                  summary='Timeout waiting for OTA job status.')
+        if exec_status == 'SUCCEEDED':
+            return OtaTestResult(testName=self._name, result=OtaTestResult.ERROR,
+                                 summary='OTA update complete too fast before we can cancel the job.')
 
         # Force cancel the first job that's in progress, device should pick up the 2nd update and succeed.
         iot_client = boto3.client('iot')
```
ericyeh1995, line 89: Comment on line 90, though unlikely to fail, is it possible to check the status of cancel_job_execution?

tgsong: We don't need to actually. Before it's failing because the update is already finished. Now if it fails we just let the exception pop up and fails the tests.

PR #2408, opened 2020-08-26
Added --pointer-primitive-check CBMC flag

SaswatPadhi:
Description
-----------
This PR adds the `--pointer-primitive-check` flag to CBMC proofs.

Checklist:
----------
- [x] My code is Linted.


tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
```diff
@@ -45,15 +45,17 @@ void harness() {
   IotHttpsRequestHandle_t reqHandle = allocate_IotRequestHandle();
   if (reqHandle)
     __CPROVER_assume(is_valid_IotRequestHandle(reqHandle));
-  uint32_t nameLen;
-  uint32_t valueLen;
+
```
markrtuttle, line 48: Why the new blank line?

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
```diff
@@ -45,15 +45,17 @@ void harness() {
...
+  size_t nameLen;
+  size_t valueLen;
   __CPROVER_assume(nameLen < CBMC_MAX_OBJECT_SIZE);
   __CPROVER_assume(valueLen < CBMC_MAX_OBJECT_SIZE);
-  char * pName = safeMalloc(nameLen+1);
-  char * pValue = safeMalloc(valueLen+1);
+  char *pName = safeMalloc(nameLen+1);
+  char *pValue = safeMalloc(valueLen+1);
```
markrtuttle, line 54: Is this consistent with the team's uncrustify rules?

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/Makefile.json
```diff
@@ -1,11 +1,6 @@
 {
   "ENTRY": "IotHttpsClient_AddHeader",
 
-  # A CBMC pointer is an object id followed by an offset into the object.
-  # The size of the offset is limited by the size of the object id.
-  "CBMC_OBJECTID_BITS": "7",
-  "CBMC_MAX_OBJECT_SIZE": "\"((UINT32_MAX>>CBMC_OBJECTID_BITS)>>1)\"",
-
   "CBMCFLAGS":
```
markrtuttle, line 4: Why deleting these definitions, where are the definitions you are using?

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
```diff
@@ -11,18 +11,25 @@
...
-  initialize_IotResponseHandle(respHandle);
   if (respHandle) {
     __CPROVER_assume(is_valid_IotResponseHandle(respHandle));
     // Until a CBMC issue is addressed
     // https://github.com/diffblue/cbmc/issues/5004
     // copy an assumption from is_valid above as a simple assignment below.
     respHandle->httpParserInfo.readHeaderParser.data = respHandle;
   }
-  __CPROVER_assume(0 < valueLen && valueLen < UINT32_MAX);
+  initialize_IotResponseHandle(respHandle);
```
markrtuttle, line 21: Why move all of this from an allocate to an initialize function?  This seems inconsistent with current proof styles.

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/Makefile.json
```diff
@@ -45,9 +40,7 @@
   ],
   "DEF":
   [
-    "MAX_ACCEPTED_SIZE={MAX_ACCEPTED_SIZE}",
-    "CBMC_OBJECTID_BITS={CBMC_OBJECTID_BITS}",
-    "CBMC_MAX_OBJECT_SIZE={CBMC_MAX_OBJECT_SIZE}"
+    "MAX_ACCEPTED_SIZE={MAX_ACCEPTED_SIZE}"
```
markrtuttle, line 43: Where is MAX_ACCEPTED_SIZE defined?  I can't see what the change here is.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -12,7 +14,7 @@ IotHttpsRequestHandle_t allocate_IotRequestHandle();
  * without allocating memory.
  */
 void *safeMalloc(size_t xWantedSize) {
-  return nondet_bool() ? malloc(xWantedSize) : NULL;
+  return xWantedSize > 0 && nondet_bool() ? malloc(xWantedSize) : NULL;
```
markrtuttle, line 17: Why add this conjunct?

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -100,15 +102,15 @@ size_t IotNetworkInterfaceReceive( void * pConnection,
 }
 
 IotNetworkError_t IotNetworkInterfaceCallback( void * pConnection,
-					       IotNetworkReceiveCallback_t
-					         receiveCallback,
-					       void * pContext ) {
+                                               IotNetworkReceiveCallback_t
```
markrtuttle, line 105: Looks like a lot of useless whitespace changes, is your editor set to change between tabs and spaces?

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -246,9 +248,21 @@ IotHttpsResponseHandle_t allocate_IotResponseHandle() {
     safeMalloc(sizeof(struct _httpsResponse));
   if(pResponseHandle) {
     size_t headerLen;
-    size_t bodyLen;
+    __CPROVER_assume(headerLen < CBMC_MAX_OBJECT_SIZE);
```
markrtuttle, line 251: Is this really needed here?  Can't this wait for the is_valid predicate?

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -246,9 +248,21 @@ IotHttpsResponseHandle_t allocate_IotResponseHandle() {
...
     pResponseHandle->pHeaders = safeMalloc(headerLen);
+    pResponseHandle->pHeadersEnd = pResponseHandle->pHeadersCur = pResponseHandle->pHeaders;
+    if (pResponseHandle->pHeaders) {
+      pResponseHandle->pHeadersEnd = pResponseHandle->pHeaders + headerLen - 1;
+    }
```
markrtuttle, line 256: This seems very wrong.  Aren't current and end supposed to point to arbitrary locations in the buffer, not necessarily exactly the head and the end.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -246,9 +248,21 @@ IotHttpsResponseHandle_t allocate_IotResponseHandle() {
...
+
+    size_t bodyLen;
+    __CPROVER_assume(bodyLen < CBMC_MAX_OBJECT_SIZE);
     pResponseHandle->pBody = safeMalloc(bodyLen);
+    pResponseHandle->pBodyEnd = pResponseHandle->pBodyCur = pResponseHandle->pBody;
+    if (pResponseHandle->pBody) {
+      pResponseHandle->pBodyEnd = pResponseHandle->pBody + bodyLen - 1;
+    }
+
```
markrtuttle, line 265: Comments for the header apply here to the body.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -268,10 +282,8 @@ initialize_IotResponseHandle(IotHttpsResponseHandle_t pResponseHandle) {
 }
 
 int is_valid_IotResponseHandle(IotHttpsResponseHandle_t pResponseHandle) {
-  int valid_headers =
-    pResponseHandle->pHeaders != NULL;
-  int valid_body =
-    pResponseHandle->pBody != NULL;
+  int valid_headers = pResponseHandle->pHeaders != NULL;
```
markrtuttle, line 285: Why change line breaks?  This was formatted so that all predicate definitions were similar.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
```diff
@@ -325,10 +337,17 @@ IotHttpsRequestHandle_t allocate_IotRequestHandle() {
...
     pRequestHandle->pHttpsResponse = allocate_IotResponseHandle();
     pRequestHandle->pHttpsConnection = allocate_IotConnectionHandle();
+
+    size_t headerLen;
+    __CPROVER_assume(headerLen < CBMC_MAX_OBJECT_SIZE);
     pRequestHandle->pHeaders = safeMalloc(headerLen);
+    pRequestHandle->pHeadersEnd = pRequestHandle->pHeadersCur = pRequestHandle->pHeaders;
+    if (pRequestHandle->pHeaders) {
+      pRequestHandle->pHeadersEnd = pRequestHandle->pHeaders + headerLen - 1;
+    }
```
markrtuttle, line 349: This seems wrong again for similar reasons.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: But it's not consistent. For example, lines 358 and 359 do not use line breaks.

https://github.com/aws/amazon-freertos/pull/2408/files/fde2d074464d339da519ef27d5760ecc47acf20b#diff-a5a8d6b34fac01a0b4afe4254cb313d2R358-R359

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: I see, I thought they should be initialized to the head and end, and can be modified later.
If we leave out the initialization entirely, CBMC is not happy. We should make sure that both pointers fall within the buffer, as you said.

Do you suggest using a `__CPROVER_assume` here?

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/Makefile.json
SaswatPadhi: Line 6 of the same file. The change was just to remove these 2 lines:

```
    "CBMC_OBJECTID_BITS={CBMC_OBJECTID_BITS}",
    "CBMC_MAX_OBJECT_SIZE={CBMC_MAX_OBJECT_SIZE}"
```

These macros now come from `#include ""cbmc.h"`

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: For zero-size allocation, NULL was not being returned and CBMC used to complain that the returned pointer is neither NULL nor valid.

Moreover, the comment right above this function reads:

```
/* Implementation of safe malloc which returns NULL if the requested
 * size is 0.  Warning: The behavior of malloc(0) is platform
 * dependent.  It is possible for malloc(0) to return an address
 * without allocating memory.
 */
```

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: The file had mixed spaces and tabs everywhere. Mostly spaces, so I normalized everything to spaces.

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
SaswatPadhi: I haven't checked the uncrustify rules, but this change makes it consistent with the other harnesses, so I'm hoping it is ...

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/Makefile.json
SaswatPadhi: I added `#included "cbmc.h"` at the top of `global_state_HTTP.c`. These definitions now come from "cbmc.h".

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: I can change both functions to whichever style the team uses, but it should be consistent.

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
SaswatPadhi: I just thought it makes it cleaner and readable --- I added line breaks between:
(1) the initialization of the RequestHandle object itself, (2) the initialization of the header name and value, (3) the actual function call (and assertions, but we don't have any).

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: Without this check, the allocation goes through but header is neither NULL nor valid after the allocation. CBMC wasn't happy about it at the `valid_header_pointers` part the within is_valid. Please see this report: https://dtjjmrnini9ie.cloudfront.net/IotHttpsClient_AddHeader-20200826-191042/out/html/index.html

Do you think we should remove it from the is_valid predicate since it's redundant?

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
SaswatPadhi: Actually I didn't move anything from allocate to initialize. I allocated and initialized the ResponseHandle first, then I allocate the name and value for the header. The way name and value are allocated for header was also inconsistent, and I made it consistent with other files, e.g. https://github.com/aws/amazon-freertos/pull/2408/files/fde2d074464d339da519ef27d5760ecc47acf20b#diff-6eba4bbbc722fd634f2b6364c6badb6dR49-R54

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: As per our discussion, I have now added offsets that are non-deterministically assigned between 0 and headerLen, and are then added to `pheaders`.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: As per our discussion, I have now added offsets that are non-deterministically assigned between 0 and `bodyLen`, and are then added to `pBody`.

tools/cbmc/proofs/HTTP/global_state_HTTP.c
SaswatPadhi: As per our discussion, I have now added offsets that are non-deterministically assigned between 0 and `headerLen`, and are then added to `pHeaders`.

tools/cbmc/.gitignore
```diff
@@ -2,3 +2,5 @@ cbmc.txt
 property.xml
 coverage.xml
 *.goto
+TAGS-*
+html
```
markrtuttle, line 6: Please check the template starter kit and add these things there if they aren't already there.  (Final new line missing.)

tools/cbmc/include/cbmc.h
```diff
@@ -24,8 +27,10 @@
   * the maximum offset into an object in CBMC, and hence a bound on the
   * size of objects in CBMC.
   */
-#define CBMC_BITS               7
-#define CBMC_MAX_OBJECT_SIZE    ( 0xFFFFFFFF >> ( CBMC_BITS + 1 ) )
+#ifndef CBMC_OBJECT_BITS
+#define CBMC_OBJECT_BITS        7
+#endif
+#define CBMC_MAX_OBJECT_SIZE    ( 0xFFFFFFFF >> ( CBMC_OBJECT_BITS + 1 ) )
```
markrtuttle, line 33: Didn't you show to me that something like this is defined by default in a cbmc header file?  Is this a duplication?

tools/cbmc/include/cbmc.h
```diff
@@ -98,3 +103,5 @@ void vPortFree( void * pv )
     ( void ) pv;
     free( pv );
 }
+
+#endif
```
markrtuttle, line 107: Missing new line (here and elsewhere)

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
```diff
@@ -45,15 +45,17 @@ void harness() {
   IotHttpsRequestHandle_t reqHandle = allocate_IotRequestHandle();
   if (reqHandle)
     __CPROVER_assume(is_valid_IotRequestHandle(reqHandle));
-  uint32_t nameLen;
-  uint32_t valueLen;
+
+  size_t nameLen;
+  size_t valueLen;
```
markrtuttle, line 50: You changed the type, can you confirm is this the type in the function prototype?  I see size_t used in the ReadHeader harness below.

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
```diff
@@ -11,18 +11,25 @@
...
+
+  size_t nameLen;
+  size_t valueLen;
+  __CPROVER_assume(nameLen < CBMC_MAX_OBJECT_SIZE);
+  __CPROVER_assume(valueLen < CBMC_MAX_OBJECT_SIZE);
+  char *pName = safeMalloc(nameLen+1);
```
markrtuttle, line 27: Why plus 1?  This is a mistake.  Not all strings are null terminated in this code.  But I see the original AddHeader above uses plus 1.

markrtuttle (changes requested):

tools/cbmc/include/cbmc.h
SaswatPadhi: This is the same file. I only added the `#ifndef CBMC_OBJECT_BITS` around `#define CBMC_OBJECT_BITS` during our Chime call.

tools/cbmc/proofs/HTTP/IotHttpsClient_AddHeader/IotHttpsClient_AddHeader_harness.c
SaswatPadhi: I fixed both this and ReadHeader harnesses. The function signature uses uint32_t, so I switched to that.

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
SaswatPadhi: I made the change to be consistent with AddHeader harness, but you are right, +1 is not necessary and the functions don't require the strings to be null-terminated. Looking at the AddHeader harness, I thought they do and might go out of bound when searching for the end, but they don't and CBMC is happy.

SaswatPadhi: I think I didn't test properly on my local branch. CBMC isn't happy.  It (within `strncmp`) requires the strings to be null terminated.

I have added the `+1` to both AddHeader and ReadHeader. Let me  know what you think.

markrtuttle: This requires review by the author @sarenameas:  Should the length be the string length of name and value or the actual byte length of the buffer containing name and value.

markrtuttle (changes requested):
I have one remaining question that I'd like Sarena, the code author, to answer.

SaswatPadhi:
I requested a review from @sarenameas

tools/cbmc/proofs/HTTP/IotHttpsClient_ReadHeader/IotHttpsClient_ReadHeader_harness.c
sarenameas: the nameLen passed into ReadHeader() is the length of the name of the header field. Since pName is unbounded we have no idea how long the name in the pName buffer is going to be, so this look correct to me..

the valueLen passed into ReadHeader() is the size of the buffer pValue.

SaswatPadhi: @markrtuttle: I discussed this issue offline with @sarenameas and it turns out there is a bug in `IotHttpsClient_AddHeader` implementation in `iot_https_client.c`.

Thank you both for the thorough review!

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2908,28 +2908,32 @@ IotHttpsReturnCode_t IotHttpsClient_AddHeader( IotHttpsRequestHandle_t reqHandle
...
+    HTTPS_ON_ARG_ERROR_MSG_GOTO_CLEANUP( ( ( nameLen != FAST_MACRO_STRLEN( HTTPS_HOST_HEADER ) ) ||
+                                           ( strncmp( pName, HTTPS_HOST_HEADER, FAST_MACRO_STRLEN( HTTPS_HOST_HEADER ) ) != 0 ) ),
                                          IOT_HTTPS_INVALID_PARAMETER,
                                          "Attempting to add auto-generated header %s. This is not allowed.",
                                          HTTPS_HOST_HEADER );
 
     /* Check for auto-generated header "User-Agent". This header is created and placed into the header buffer space
      * in IotHttpsClient_InitializeRequest(). */
-    HTTPS_ON_ARG_ERROR_MSG_GOTO_CLEANUP( strncmp( pName, HTTPS_USER_AGENT_HEADER, FAST_MACRO_STRLEN( HTTPS_USER_AGENT_HEADER ) ) != 0,
+    HTTPS_ON_ARG_ERROR_MSG_GOTO_CLEANUP( ( ( nameLen != FAST_MACRO_STRLEN( HTTPS_USER_AGENT_HEADER ) ) ||
```
sarenameas, line 2935: Thank you for making this change. I have unit tested it: https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/573/ and all pass.

sarenameas (approved):

markrtuttle (approved):

PR #2495, opened 2020-09-29
Update IDE projects for Shim layer

leegeth:
Update IDE projects for Shim layer

Description
-----------
Update IDE projects for Shim layer.
All the project files are updated and verified.

The below 2 projects will be updated as part of a different PR
1. Nordic project.

Checklist:
----------


libraries/abstractions/platform/include/platform/transport_interface.h
```diff
@@ -1,4 +1,5 @@
 /*
```
aggarw13, line 1: The `master` branch does not store a copy of the `transport_interface.h` file, instead just utilizes `coreMQTT` as the SSoT. Could you follow the same for this branch as well?

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1149,7 +1149,7 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
         transport.recv = transportRecv;
 
         /* Fill the values for network buffer. */
-        networkBuffer.pBuffer = &( connToContext[ contextIndex ].buffer );
+        networkBuffer.pBuffer = &( connToContext[ contextIndex ].buffer[0] );
```
aggarw13, line 1152: Was this incorrect way of getting the starting address not causing issues in testing before?

libraries/c_sdk/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -828,7 +826,7 @@ TEST( MQTT_Unit_API, OperationWaitTimeout )
         TEST_ASSERT_NOT_NULL( _pMqttConnection );
 
         /* Set the MQTT Context for the new MQTT Connection*/
-        TEST_ASSERT_EQUAL( IOT_MQTT_SUCCESS, _setContext( _pMqttConnection, _sendDelay ) );
+        TEST_ASSERT_EQUAL( IOT_MQTT_SUCCESS, _setContext( _pMqttConnection, ( TransportSend_t ) _sendDelay ) );
```
aggarw13, line 829: Instead of a explicit casting, can their be 2 different `_sendDelay` functions, one that adheres to network abstraction's send signature, and the other that adheres to the new `TransportSend_t`'s?

libraries/c_sdk/standard/mqtt/test/unit/iot_tests_mqtt_subscription.c
```diff
@@ -217,6 +220,47 @@ static bool _waitForCount( IotMutex_t * pMutex,
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Wait for a reference count to reach a target value, using FreeRTOS semaphores,
+ * subject to a timeout.
+ */
+static bool _waitForCountSemaphore( StaticSemaphore_t * pSem,
```
aggarw13, line 227: If the thought is to prefer use of native FreeRTOS synchronization primitives instead of `IotMutex`, can `_waitForCount` be removed from this file? It would be better to have a single approach used through the file instead of multiple functions present for a similar functionality..

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
leegeth: This was not extensively tested, However this didn't cause any issue in tests on WinSim. It is good to catch the fix through a build warning reported as error.

libraries/abstractions/platform/include/platform/transport_interface.h
leegeth: Yea. Agree. I will do that in my next PR as that will need update to all the project files and retesting.

libraries/c_sdk/standard/mqtt/test/unit/iot_tests_mqtt_api.c
leegeth: I have thought that as an option too. I think that would be a better option. Will add that fix.

libraries/c_sdk/standard/mqtt/test/unit/iot_tests_mqtt_subscription.c
leegeth: I have looked at that choice as well. This is for a mutex used by the library, not just by tests. If I need to update it, I will need to analyse and update that as well. I will keep that task as a different one to analyze and improve it if possible. It will be outside the scope of this PR.

aggarw13 (approved):

aggarg (approved):

PR #2497, opened 2020-09-30
Bump submodules of C-SDK libraries for disabled leaf submodules

aggarw13:
Bump submodule commits of `coreMQTT`, `Device Shadow` and `coreJSON` repositories to utilize the `update=none` configuration made in the library repositories to avoid their leaf submodules to be cloned for `git submodule update --init --recursive` command on hub repository (like this one, `amazon-freertos`)


leegeth (commented):
I thought the plan was to use v1.0.0 version of the libraries. If we move the submodule pointers ahead, our development will be based on the latest master and RC will have to be moved back to v1.0.0 tag.

leegeth (approved):

leegeth:
> I thought the plan was to use v1.0.0 version of the libraries. If we move the submodule pointers ahead, our development will be based on the latest master and RC will have to be moved back to v1.0.0 tag.

Decided that we can update the submodule pointers to the latest master. However, this will mean a new release tag for the libraries.

dan4thewin (approved):

PR #2509, opened 2020-10-04
Make common_io as a library target

aggarw13:
Update `AFR::common_io` module to be built as a library so that application can link against it.

Currently, the module is configured as an `INTERFACE` which makes it as a non-independent library target. By removing the `INTERFACE` configuration, CMake will expose `afr_common_io` as a library target which can be linked against with `target_link_libraries`.

### DIFF
--- a/vendors/st/boards/stm32l475_discovery/CMakeLists.txt
+++ b/vendors/st/boards/stm32l475_discovery/CMakeLists.txt
@@ -158,7 +158,7 @@ target_sources(
 target_include_directories(
     AFR::common_io::mcu_port
     INTERFACE
-        "${AFR_MODULES_ABSTRACTIONS_DIR}/platform/freertos/include"
+        "${AFR_MODULES_ABSTRACTIONS_DIR}/platform/freertos/include/platform"
         # test code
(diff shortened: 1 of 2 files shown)


n9wxu (approved):

vendors/st/boards/stm32l475_discovery/CMakeLists.txt
```diff
@@ -155,6 +155,15 @@ target_sources(
         $<${AFR_IS_TESTING}:${afr_ports_dir}/common_io/iot_test_common_io_internal.c>
 )
 
+target_include_directories(
```
lundinc2, line 158: Did Espressif need an update as well? Or does it still compile

aggarw13: Espressif does compile as validated both locally and from PR build checks

lundinc2 (dismissed):

n9wxu (dismissed):

lundinc2 (approved):

n9wxu (approved):

PR #2517, opened 2020-10-06
Wait for republish in MQTT integration tests for AWS IoT Core

yourslab:
Description
-----------

AWS IoT Core takes some time before resending an unack'd PUBLISH, so a delay is added before resuming the persistent session.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -1554,6 +1554,9 @@ void Restore_Session_Duplicate_Incoming_Publish_Qos1()
 
     /* Make sure that a record was created for the incoming PUBLISH packet. */
     TEST_ASSERT_NOT_EQUAL( MQTT_PACKET_ID_INVALID, context.incomingPublishRecords[ 0 ].packetId );
+    context.incomingPublishRecords[ 0 ].packetId = MQTT_PACKET_ID_INVALID;
+
+    vTaskDelay( pdMS_TO_TICKS( 30 * 1000 ) );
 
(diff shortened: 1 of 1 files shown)


tests/integration_test/core_mqtt_system_test.c
```diff
@@ -1562,6 +1569,17 @@ void Restore_Session_Duplicate_Incoming_Publish_Qos1()
...
 
+    if( testingAgainstAWS )
+    {
+        /* AWS IoT Core takes at least 30 seconds before resending a PUBLISH. */
+        vTaskDelay( pdMS_TO_TICKS( AWS_IOT_CORE_REPUBLISH_INTERVAL_MS ) );
+
+        /* When the resent PUBLISH is received, the DUP flag will not be set to 1.
+         * Therefore, the packet ID of the original PUBLISH must be cleared from
+         * the incoming publish records array. */
+        context.incomingPublishRecords[ 0 ].packetId = MQTT_PACKET_ID_INVALID;
```
aggarw13, line 1580: Why is this required for the `MQTTContext_t`. According to the spec, even if the DUP flag is set, it does not mean that the receiving end had received the original PUBLISH message..

yourslab: IoT Core will not set the DUP flag in the resent publish. However, during the first time we receive the publish, the state array is updated. Therefore, when we receive it again and the DUP flag is not set, the library will return a `MQTTStateCollision` error.

aggarw13 (approved):

dan4thewin (approved):

PR #2523, opened 2020-10-07
Port MQTT Connection Sharing Demo

muneebahmed10:
Description
-----------

Adds the MQTT connection sharing demo from FreeRTOS. Fixes one build failure from #2549

Checklist:
----------

- [ ] My code is Linted.


demos/coreMQTT/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+# AFR coreMQTT demo
+afr_demo_module(core_mqtt)
+
+# Set the default demo that is enabled in the downloaded package.
+afr_set_demo_metadata(ID "CORE_MQTT_DEMO")
```
sarenameas, line 5: So OCW will take this ID and prepend "CONFIG_" and append "_ENALBED", so we have to choose one to be default.

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1906 @@
...
+    if( ret == EXIT_SUCCESS )
+    {
+        /* Form an MQTT connection with a persistent session. */
+        xMQTTStatus = prvMQTTConnect( &globalMqttContext, &xNetworkContext, false );
+        ret = ( xMQTTStatus == MQTTSuccess ) ? EXIT_SUCCESS : EXIT_FAILURE;
+    }
+
+    if( ret == EXIT_SUCCESS )
+    {
+        configASSERT( globalMqttContext.connectStatus = MQTTConnected );
```
sarenameas, line 1830:
```suggestion
configASSERT( globalMqttContext.connectStatus == MQTTConnected );
```
This might be missing in the FreeRTOS/FreeRTOS version?

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1906 @@
...
+            {
+                pxCommand->pxCmdContext->xReturnStatus = xStatus;
+            }
+
+            break;
+
+        case RECONNECT:
+            /* Reconnect TCP. */
+            xNetworkResult = prvDisconnectNetwork( globalMqttContext.transportInterface.pNetworkContext );
+            configASSERT( xNetworkResult == pdPASS );
```
sarenameas, line 1168: Not being able to connect or disconnect from the network is a common failure that could trigger this assert. If this assert is triggered, then the demo will stop in some forever exception loop. This stop keeps "-------DEMO FINISHED-------\n" from printing and then the CI will hand on the demo for 30 minutes. We need to break and return an error for these type of errors.

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1906 @@
...
+/* Compile time error for some undefined configs, and provide default values
+ * for others. */
+#ifndef democonfigMQTT_BROKER_ENDPOINT
+    #define democonfigMQTT_BROKER_ENDPOINT    clientcredentialMQTT_BROKER_ENDPOINT
+#endif
+
+#ifndef democonfigMQTT_BROKER_PORT
+    #define democonfigMQTT_BROKER_PORT    clientcredentialMQTT_BROKER_PORT
+#endif
+
```
sarenameas, line 99: Can you please add this below. The CI will replace this macro in this file directly.
```suggestion

#ifndef democonfigROOT_CA_PEM
    #define democonfigROOT_CA_PEM    tlsATS1_ROOT_CERTIFICATE_PEM
#endif
```

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1906 @@
...
+/* Retry utilities include. */
+#include "retry_utils.h"
+
+/* Transport interface implementation include header for TLS. */
+#include "transport_secure_sockets.h"
+
+/* Credentials for the MQTT broker endpoint. */
+#include "aws_clientcredential.h"
+
```
sarenameas, line 76:
```suggestion

/* Include header for root CA certificates. */
#include "iot_default_root_certificates.h"
```

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1906 @@
...
+    /* Initialize the MQTT broker information. */
+    xServerInfo.pHostName = democonfigMQTT_BROKER_ENDPOINT;
+    xServerInfo.hostNameLength = sizeof( democonfigMQTT_BROKER_ENDPOINT ) - 1U;
+    xServerInfo.port = democonfigMQTT_BROKER_PORT;
+
+    /* Set the Secure Socket configurations. */
+    xSocketConfig.enableTls = true;
+    xSocketConfig.sendTimeoutMs = mqttexampleTRANSPORT_SEND_RECV_TIMEOUT_MS;
+    xSocketConfig.recvTimeoutMs = mqttexampleTRANSPORT_SEND_RECV_TIMEOUT_MS;
+
```
sarenameas, line 845:
```suggestion
    xSocketsConfig.pRootCa = democonfigROOT_CA_PEM;
    xSocketsConfig.rootCaSize = sizeof( democonfigROOT_CA_PEM ); 

```

demos/coreMQTT/mqtt_demo_connection_sharing.c
muneebahmed10: Changed

demos/coreMQTT/mqtt_demo_connection_sharing.c
muneebahmed10: Yeah, it's missing there too

demos/coreMQTT/CMakeLists.txt
muneebahmed10: I put plaintext as the default for now

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1916 @@
...
+    /* Connect to the broker. We connect here with the "clean session" flag set
+     * to true in order to clear any prior state in the broker. We will disconnect
+     * and later form a persistent session, so that it may be resumed if the
+     * network suddenly disconnects. */
+    LogInfo( ( "Creating a TCP connection to %s.\r\n", democonfigMQTT_BROKER_ENDPOINT ) );
+    xNetworkStatus = prvConnectNetwork( &xNetworkContext );
+
+    ret = EXIT_FAILURE;
+
+    if( xNetworkStatus == pdPASS )
```
sarenameas, line 1799: I went to office hours and we have decided to keep the loop but limit it by democonfigMQTT_MAX_DEMO_COUNT

muneebahmed10: Is that a requirement? I think the demo's more straightforward without creating/deleting tasks in a loop, but I added the loop back anyway.

sarenameas: if the loop exists in FreeRTOS/FreeRTOS then we want to try to keep a semblance of it. Thank you

demos/include/iot_demo_runner.h
```diff
@@ -78,6 +78,15 @@
...
     #endif
+#elif defined( CONFIG_CORE_MQTT_CONNECTION_SHARING_DEMO_ENABLED )
+    #define DEMO_entryFUNCTION              RunCoreMqttConnectionSharingDemo
+    #if defined( democonfigMQTT_ECHO_TASK_STACK_SIZE )
+        #undef democonfigDEMO_STACKSIZE
+        #define democonfigDEMO_STACKSIZE    democonfigMQTT_ECHO_TASK_STACK_SIZE
+    #endif
+    /* This demo must have priority at least one greater than idle. */
+    #undef democonfigDEMO_PRIORITY
+    #define democonfigDEMO_PRIORITY         ( tskIDLE_PRIORITY + 1 )
```
sarenameas, line 89: Should this customer configurable?

vendors/xilinx/boards/microzed/aws_demos/config_files/mqtt_demo_connection_sharing_config.h
```diff
@@ -0,0 +1,100 @@
+/*
+ * FreeRTOS Kernel V202010.00
```
sarenameas, line 2: nitpick I think we want to take out the "Kernel" from this.

demos/include/iot_demo_runner.h
muneebahmed10: Ideally yes, I didn't see a way to do it without creating a new `democonfigCONNECTION_SHARING_DEMO_TASK_PRIORITY` macro in `aws_demo_config.h` for all of the boards. Should I do that?

sarenameas: It might help in the future for certain boards that need different priorities

muneebahmed10: Updated

sarenameas (dismissed):

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1959 @@
...
+            LogInfo( ( "Disconnecting TCP connection." ) );
+            xNetworkStatus = prvDisconnectNetwork( &xNetworkContext );
+            ret = ( xNetworkStatus == pdPASS ) ? EXIT_SUCCESS : EXIT_FAILURE;
+            xNetworkConnectionCreated = pdFALSE;
+        }
+
+        if( ret == EXIT_SUCCESS )
+        {
+            LogInfo( ( "RunCoreMqttConnectionSharingDemo() completed an iteration successfully. Total free heap is %u.\r\n",
+                       xPortGetFreeHeapSize() ) );
```
sarenameas, line 1914: Let's delete this xPortGetFreeHeapSize(). 
```suggestion
            LogInfo( ( "RunCoreMqttConnectionSharingDemo() completed an iteration successfully." ) ) );
```

muneebahmed10: Changed

sarenameas (approved):

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1962 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * http://www.FreeRTOS.org
+ * http://aws.amazon.com/freertos
+ */
+
+/*
+ * Demo for showing use of the managed MQTT API shared between multiple tasks.
```
cobusve, line 27: This file is terribly large and seems to be containing a number of different concerns. Can we split this into at least one file per task that is running? It is very hard to understand if it is just one giant file with stuff sprinkled throughout.

If I have e.g. a file called PublishTask.c and SubscribeTask.c it is obvious where to find stuff, separation of concerns AND less code (haystack) to search through for what I am looking for (needle).

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -0,0 +1,1962 @@
...
+/**
+ * @brief The MQTT client identifier used in this example.  Each client identifier
+ * must be unique so edit as required to ensure no two clients connecting to the
+ * same broker use the same client identifier.
+ */
+    #define democonfigCLIENT_IDENTIFIER    clientcredentialIOT_THING_NAME
+#endif
+
+/* Provide default values for undefined configs. */
+#ifndef democonfigMQTT_BROKER_ENDPOINT
```
cobusve, line 87: This is a strange pattern to see in a demo. This is usually what a library does if you are not sure the user will define these things. 

I would suggest that all of this stuffs should be in a DemoConfig.h file or similar which is defining what you want defined.

The demo should be a "demo" of how a user is expected to use the library. I do not expect anyone writing a real application to need config #ifdefs to add in config defaults for things they did not configure. I would expect a config file somewhere, and with that config file a description of what each config item does.

Lots of examples in the FreeRTOS Kernel demos of this e.g. https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS/Demo/WIN32-MSVC/main_blinky.c

cobusve (approved):

demos/coreMQTT/mqtt_demo_connection_sharing.c
muneebahmed10: I understand. I followed the same pattern as the other demos which had comments in the [demo config files](https://github.com/aws/amazon-freertos/blob/master/vendors/pc/boards/windows/aws_demos/config_files/mqtt_demo_mutual_auth_config.h) and the ifndef in the actual [demo file](https://github.com/aws/amazon-freertos/blob/master/demos/coreMQTT/mqtt_demo_mutual_auth.c#L94). Should that be changed for all demos?

PR #2534, opened 2020-10-08
OTA: Abort the image if job status update fails

pvyawaha:
Abort the update image if self-test jobs status fails.

Description
-----------

The change adds a check before activating the image to abort if the publish for job status update failed. The image will be activated only when the publish is successful. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
+++ b/libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
@@ -1131,7 +1131,23 @@ static OTA_Err_t prvProcessDataHandler( OTA_EventData_t * pxEventData )
 
             if( xErr != kOTA_Err_None )
             {
-                OTA_LOG_L2( "[%s] Failed to update job status %d\r\n", OTA_METHOD_NAME, xErr );
+                OTA_LOG_L1( "[%s] Failed to update job status %d\r\n", OTA_METHOD_NAME, xErr );
+
+                /* Job status update failed, call the platform specific code to abort the image. */
+                xErr = xOTA_Agent.xPALCallbacks.xSetPlatformImageState( xOTA_Agent.ulServerFileID, eOTA_ImageState_Aborted );
+
+                if( xErr != kOTA_Err_None )
+                {
+                    OTA_LOG_L1( "[%s] Error trying to set platform image state (0x%08x)\r\n", OTA_METHOD_NAME, ( int32_t ) xErr );
+                }
+
+                /* Let application know of the fail event. */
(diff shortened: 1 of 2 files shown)


gkwicker (changes requested):
One question, see comment.

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1131,7 +1131,23 @@ static OTA_Err_t prvProcessDataHandler( OTA_EventData_t * pxEventData )
 
             if( xErr != kOTA_Err_None )
             {
-                OTA_LOG_L2( "[%s] Failed to update job status %d\r\n", OTA_METHOD_NAME, xErr );
+                OTA_LOG_L1( "[%s] Failed to update job status %d\r\n", OTA_METHOD_NAME, xErr );
+
+                /* Job status update failed, call the platform specific code to abort the image. */
+                xErr = xOTA_Agent.xPALCallbacks.xSetPlatformImageState( xOTA_Agent.ulServerFileID, eOTA_ImageState_Aborted );
+
+                if( xErr != kOTA_Err_None )
```
gkwicker, line 1139: Under what conditions might `xErr != kOTA_Err_None`?  For example if this could be due to a busy/unavailable condition, should you sleep/retry here?  And what should an application do if this can never complete?  As the code is written now there's no way for the application to know that the PAL was never successfully able to set the image state to `eOTA_ImageState_Aborted`.

pvyawaha: For errros we can update the job status with error codes but it is not applicable here as we are in this if block when the updating job status failed. The xSetPlatformImageState function is implemented in the OTA pal, where it calls the actual platform drivers to set the image state. This can be flash location or file system depending on the platform. The OTA pal should have retries if retrying can be successful for those platform drivers. The application is abstracted from the platform state and it is handled by the ota agent by matching states of the job document and the platform before committing new image.

tgsong (approved):

libraries/freertos_plus/aws/ota/src/aws_iot_ota_agent.c
```diff
@@ -1131,7 +1131,23 @@ static OTA_Err_t prvProcessDataHandler( OTA_EventData_t * pxEventData )
...
+                {
+                    OTA_LOG_L1( "[%s] Error trying to set platform image state (0x%08x)\r\n", OTA_METHOD_NAME, ( int32_t ) xErr );
```
gkwicker, line 1141: should you call the complete callback on line 1145 with a different error code if the PAL has failed here?

gkwicker (approved):

PR #2548, opened 2020-10-11
Port MQTT Keep Alive demo from FreeRTOS/FreeRTOS

sarenameas:
https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/MQTT_Keep_Alive/DemoTasks/KeepAliveMQTTExample.c

- Add text about how this demo uses MQTT_ReceiveLoop instead of MQTT_ProcessLoop, so it shows how to handle the keep-alive in the application.
- Add the demo to iot_demo_runner.h
- Add mqtt_demo_keep_alive_config.h to all boards.
- Change all transport_freertos uses to transport_secure_sockets.
- Return EXIT_SUCCESS and EXIT_FAILURE in the demo, so necessitates depleting non-fatal configASSERTs.
- Fixes in other demos for consistency

Checklist:
----------

- [ ] My code is Linted.


demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+ */
+
+/*
+ * Demo for showing use of the managed MQTT API.
+ *
+ * The Example shown below uses this API to create MQTT messages and
+ * send them over the connection established using Secure Sockets.
+ * The example is single threaded and uses statically allocated memory;
+ * it uses QOS0 and therefore does not implement any retransmission
+ * mechanism for Publish messages.
```
aggarw13, line 34:
```suggestion
/*
 * Demo that shows use of the MQTT API without its keep-alive feature. This demo instead implements the keep-alive functionality in the application. 
 *
 * The example shown below uses this API to create MQTT messages and
 * send them over the TCP connection established using Secure Sockets. 
 * It shows how the MQTT API can be used without the keep-alive feature,
 * so that the application can implements its own keep-alive functionality
 * for MQTT. The example is single threaded and uses statically allocated memory;
 * it uses QOS0 and therefore does not implement any retransmission
 * mechanism for Publish messages.
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+
+/**
+ * @brief Dimensions a file scope buffer currently used to send and receive MQTT data
+ * from a socket.
+ */
+#define mqttexampleSHARED_BUFFER_SIZE                ( 500U )
+
+/**
+ * @brief Time to wait between each cycle of the demo implemented by
+ * RunCoreMqttPlaintextDemo().
```
aggarw13, line 131:
```suggestion
 * RunCoreMqttKeepAliveDemo().
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+/**
+ * @brief Timeout for MQTT_ReceiveLoop in milliseconds.
+ */
+#define mqttexampleRECEIVE_LOOP_TIMEOUT_MS           ( 500U )
+
+/**
+ * @brief Keep alive time reported to the broker while establishing an MQTT connection.
+ *
+ * It is the responsibility of the Client to ensure that the interval between
+ * Control Packets being sent does not exceed the this Keep Alive value. In the
```
aggarw13, line 144:
```suggestion
 * It is the responsibility of the client to ensure that the interval between
 * control packets being sent does not exceed the this keep-alive value. In the
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+ * absence of sending any other Control Packets, the Client MUST send a
```
aggarw13, line 145:
```suggestion
 * absence of sending any other control packets, the client MUST send a
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+
+/**
+ * @brief Set to true when PINGREQ is sent then false when PINGRESP is received.
+ */
+static volatile bool xWaitingForPingResp = false;
+
+/**
+ * @brief Set to true when an error occurs in the prvKeepAliveTimerCallback().
+ */
+static volatile bool xKeepAliveError = false;
```
aggarw13, line 377: What is the purpose of making these _shared_ memory variables `volatile`?

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1172 @@
...
+    ( void ) pNetworkInterface;
+
+
+    ulGlobalEntryTimeMs = prvGetTimeMs();
+
+    for( ; ulDemoRunCount < democonfigMQTT_MAX_DEMO_COUNT; ulDemoRunCount++ )
+    {
+        /* Serialize a PINGREQ packet to send upon invoking the keep-alive timer
+         * callback. */
+        xMQTTStatus = MQTT_SerializePingreq( &xPingReqBuffer );
```
aggarw13, line 450: Can this be kept before the loop? The buffer needs to be populated only once with the PINGREQ packet

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Because they are accessed by multiple threads.

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Then I would have to put the loop into an if case.

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Sure

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Thanks

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: sure i will get rid of all weird capitalizations.

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+
+        if( xDemoStatus == pdPASS )
+        {
+            /* Start the timer for keep-alive. */
+            xTimerStatus = xTimerStart( xKeepAliveTimer, 0 );
+
+            if( xTimerStatus != pdPASS )
+            {
+                LogError( ( "xTimerStart() failed to start KeepAliveTimer." ) );
+                xDemoStatus = pdFAIL;
```
aggarw13, line 523: Can the `xTimerStatus` variable be avoided to instead just use `xDemoStatus`?
```suggestion
            xDemoStatus = xTimerStart( xKeepAliveTimer, 0 );

            if( xDemoStatus != pdPASS )
            {
                LogError( ( "xTimerStart() failed to start KeepAliveTimer." ) );
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+                     * back to the application. */
+                    LogInfo( ( "Attempt to receive publish message from broker." ) );
+                    xMQTTStatus = MQTT_ReceiveLoop( &xMQTTContext,
+                                                    mqttexampleRECEIVE_LOOP_TIMEOUT_MS );
+
+                    if( xMQTTStatus != MQTTSuccess )
+                    {
+                        LogError( ( "MQTT_ReceiveLoop() failed with status %s.",
+                                    MQTT_Status_strerror( xMQTTStatus ) ) );
+                        xDemoStatus = pdFAIL;
```
aggarw13, line 560: If one iteration of the loop fails, should the loop continue executing?
In the `mqtt_demo_mutual_auth` demo, I have put the `xDemoStatus == pdPASS` condition in the condition expression of the loop to break out of the loop whenever failure occurs. (Alternatively, you can put a `break` here)

aggarw13: Also, would be good to make similar changes in the plaintext and serializer demos

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+        /* Whether the demo operations preceeding are successful or not, we
+         * always disconnect an open connection to free up system resources. */
+        if( xIsConnectionEstablished == pdTRUE )
+        {
+            /* Send an MQTT Disconnect packet over the already connected TCP socket.
+             * There is no corresponding response for the disconnect packet. After
+             * sending the disconnect, the client must close the network connection. */
+            LogInfo( ( "Disconnecting the MQTT connection with %s.",
+                       democonfigMQTT_BROKER_ENDPOINT ) );
+            xMQTTStatus = MQTT_Disconnect( &xMQTTContext );
```
aggarw13, line 602: Should `MQTT_Disconnect` be called if `MQTT_Connect` had failed after a successful establishment of the TCP connection?

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+        }
+
+        /* Always stop the keep-alive timer for the next iteration. */
+        xTimerStatus = xTimerStop( xKeepAliveTimer, 0 );
+
+        if( xTimerStatus != pdPASS )
+        {
+            LogError( ( "xTimerStart() failed to stop KeepAliveTimer." ) );
+            xDemoStatus = pdFAIL;
+        }
```
aggarw13, line 632: Can avoid the use of `xTimerStatus`
```suggestion
        xDemoStatus = xTimerStop( xKeepAliveTimer, 0 );

        if( xDemoStatus != pdPASS )
        {
            LogError( ( "xTimerStart() failed to stop KeepAliveTimer." ) );
        }
```

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+        /* Check for errors in the keep-alive timer. */
+        if( xKeepAliveError == true )
+        {
+            LogError( ( "An error occurred in KeepAliveTimer task." ) );
+            xDemoStatus = pdFAIL;
+        }
+
+        if( xDemoStatus == pdPASS )
+        {
+            /* Reset SUBACK status for each topic filter after completion of
```
aggarw13, line 643: Should the reset logic be _always_ called irrespective of the `xDemoStatus` value?

demos/coreMQTT/mqtt_demo_keep_alive.c
```diff
@@ -0,0 +1,1177 @@
...
+    return xReturnStatus;
+}
+/*-----------------------------------------------------------*/
+
+static BaseType_t prvMQTTPublishToTopic( MQTTContext_t * pxMQTTContext )
+{
+    MQTTStatus_t xResult;
+    MQTTPublishInfo_t xMQTTPublishInfo;
+    BaseType_t xTimerStatus;
+    BaseType_t xReturnStatus = pdPASS;
```
aggarw13, line 914: Here, the `xTimerStatus` can be avoided as well.
```suggestion
    BaseType_t xReturnStatus = pdPASS;
```

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: The demo does not continue if one iteration fails.
https://github.com/aws/amazon-freertos/pull/2548/files#diff-e1a2055bf6afd86af80b7f6e76e1a28cR662

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: It's not unsafe.

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Sure

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: it's unnecessary if the demo fails.  It doesn't matter.

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: Sure

demos/coreMQTT/mqtt_demo_keep_alive.c
sarenameas: I was having a brain fart on some logic, moved it out.

demos/coreMQTT/mqtt_demo_keep_alive.c
aggarw13: My main point is that the error message about `MQTT_Disconnect` failure would be misleading when the `MQTT_Connect` fail had itself not succeeded

sarenameas: I will turn it into a warn.

sarenameas: In the case where everything passed except MQTT_Disconnect(), we still want to consider that a failing demo. That is amiss in the mutual auth demo.

aggarw13 (approved):

vendors/espressif/boards/esp32/aws_demos/config_files/mqtt_demo_keep_alive_config.h
```diff
@@ -0,0 +1,125 @@
...
+ * to your machine.
+ * 5. After verifying that a Mosquitto broker is running successfully, update
+ * the config democonfigMQTT_BROKER_ENDPOINT to the local IP address of the
+ * Windows host machine. Please note that "localhost" or address "127.0.0.1"
+ * will not work as this example is running on a Windows Simulator and not on
+ * Windows host natively. Also note that, if the Windows host is using a
+ * Virtual Private Network(VPN), connection to the Mosquitto broker may not
+ * work.
+ *
+ * As an alternative option, a publicly hosted Mosquitto broker can also be
```
gkwicker, line 101: I would delete lines 101-105 wherever they appear in these files.  This text proposes an option, but then immediately follows that by telling the reader that they shouldn't do it.  Better to simply not include this at all.

gkwicker (changes requested):
See comment, delete all references to the test.mosquitto.org broker.

vendors/espressif/boards/esp32/aws_demos/config_files/mqtt_demo_keep_alive_config.h
sarenameas: thanks

gkwicker (approved):

aggarw13 (approved):

PR #2555, opened 2020-10-12
BLE AFQP test update

ravibhagavandas:
BLE AFQP test update

Description
-----------
1. Make Pairing request callback with Consent as optional requirement to pass qualification tests.
2. Fix the MTU callback handling in Nordic BLE port, to pass the qualification tests. 
Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/ble_hal/test/ble_test_scipts/startTests_afqp.py
+++ b/libraries/abstractions/ble_hal/test/ble_test_scipts/startTests_afqp.py
@@ -32,6 +32,7 @@ import time
 from testClass import runTest
 from bleAdapter import bleAdapter
 
+ENABLE_TC_AFQP_READ_WRITE_LONG = 1
 
 def main():
     scan_filter = dict()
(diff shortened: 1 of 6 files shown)


cobusve (approved):

libraries/abstractions/ble_hal/test/ble_test_scipts/startTests_afqp.py
```diff
@@ -32,6 +32,7 @@
 from testClass import runTest
 from bleAdapter import bleAdapter
 
+ENABLE_TC_AFQP_READ_WRITE_LONG = 1
```
lundinc2, line 35: Is this only used as a boolean? Might lead to some confusion later when using an int as a bool

lundinc2 (approved):

libraries/abstractions/ble_hal/test/ble_test_scipts/startTests_afqp.py
ravibhagavandas: Right, the config was left as an integer, so as to match the corresponding config on the DUT side. Eventually this needs to be moved to a config shared by both DUT and RPI python tests.

lundinc2: Oh I see, thanks !

PR #2556, opened 2020-10-12
OTA e2e test fix for ESP32

leegeth:
OTA e2e test fix for ESP32.

Description
-----------
OTA e2e test fix for ESP32.
Issue: The return of 0 from Network interface was an error. However, it was interpreted as a timeout in transport interface.
Fix: Handle Network interface failure in transport send.

Checklist:
----------

### DIFF
--- a/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
@@ -896,7 +896,7 @@ static int32_t transportSend( const NetworkContext_t * pNetworkContext,
     /* Sending the bytes on the network using Network Interface. */
     bytesSend = pNetworkContext->pNetworkInterface->send( pNetworkContext->pNetworkConnection, ( const uint8_t * ) pMessage, bytesToSend );
 
-    if( bytesSend < 0 )
+    if( bytesSend <= 0 )
     {
         /* Network Send Interface return negative value in case of any socket error,
          * unifying the error codes here for socket error and timeout to comply with the MQTT LTS Library.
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -896,7 +896,7 @@ static int32_t transportSend( const NetworkContext_t * pNetworkContext,
     /* Sending the bytes on the network using Network Interface. */
     bytesSend = pNetworkContext->pNetworkInterface->send( pNetworkContext->pNetworkConnection, ( const uint8_t * ) pMessage, bytesToSend );
 
-    if( bytesSend < 0 )
+    if( bytesSend <= 0 )
```
aggarw13, line 899: `/* Network Send Interface return negative value in case of any socket error,*/`
Does this comment need to be updated as well?

leegeth: Thanks. Addressed it

aggarw13 (approved):

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -896,9 +896,9 @@ static int32_t transportSend( const NetworkContext_t * pNetworkContext,
...
```
gkwicker, line 899: Is there a unit test to cover this case?  If so, does the test need to be updated?

gkwicker (approved):

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
leegeth: Thanks. No there is no unit test case covering this case for the MQTT Shim.

PR #2562, opened 2020-10-13
Added repeat mode and secure mode cmdline args support

onsankawai:
Repeat mode and Secure mode cmdline args support for Echo Server

Description
-----------
- Added repeat mode cmdline args support: `-repeat` & `-interval`
- Added secure mode cmdline args support: `-cert` & `-key`
- Updated Readme.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tools/echo_server/echo_server.go
+++ b/tools/echo_server/echo_server.go
@@ -39,33 +39,38 @@ import (
 	"time"
 )
 
-const readTimeoutSecond = 300
+const (
+	readTimeoutSecond = 300
+	portFlagName = "port"
+	repeatFlagName = "repeat"
+	intervalFlagName = "interval"
+	certFlagName = "cert"
+	keyFlagName = "key"
+)
 
 // Argument struct for JSON configuration
 type Argument struct {
-	Verbose    bool   `json:"verbose"`
-	Logging    bool   `json:"logging"`
-	Secure     bool   `json:"secure-connection"`
-	ServerPort string `json:"server-port"`
-	ServerCert string `json:"server-certificate-location"`
-	ServerKey  string `json:"server-key-location"`
+	Verbose    				bool   `json:"verbose"`
+	Logging    				bool   `json:"logging"`
+	Secure     				bool   `json:"secure-connection"`
+	RepeatMode 				bool   `json:"repeat-mode"`
+	RepeatIntervalSeconds	int    `json:"repeat-interval-seconds"`
+	ServerPort 				string `json:"server-port"`
+	ServerCert 				string `json:"server-certificate"`
+	ServerKey  				string `json:"server-key"`
 }
 
-func secureEcho(certPath string, keyPath string, port string, verbose bool) {
+func secureEcho(config *Argument) {
 
 	// load certificates
-	servertCert, err := tls.LoadX509KeyPair(certPath, keyPath)
+	serverCertBytes := []byte(config.ServerCert)
+	servertCert, err := tls.X509KeyPair(serverCertBytes, []byte(config.ServerKey))
 	if err != nil {
 		log.Fatalf("Error %s while loading server certificates", err)
 	}
 
-	serverCA, err := ioutil.ReadFile(certPath)
-	if err != nil {
-		log.Fatalf("Error %s while reading server certificates", err)
-	}
-
 	serverCAPool := x509.NewCertPool()
-	serverCAPool.AppendCertsFromPEM(serverCA)
+	serverCAPool.AppendCertsFromPEM(serverCertBytes)
 
 	//Configure TLS
 	tlsConfig := tls.Config{Certificates: []tls.Certificate{servertCert},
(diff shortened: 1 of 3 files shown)


tools/echo_server/echo_server.go
```diff
@@ -167,6 +215,11 @@ func logSetup() {
 }
 
 func main() {
+	serverPort := flag.String(portFlagName, "8883", "Port that will be used for this server.")
+	repeatMode := flag.Bool(repeatFlagName, false, "Flag indicating whether the echo server is in repeat mode.")
+	repeatIntervalSeconds := flag.Int(intervalFlagName, 1, "The repeat interval in seconds when in repeat mode.")
+	serverCert := flag.String(certFlagName, "", "X509 Server certificate for TLS connection. TLS will be enabled if this flag is set.")
```
lundinc2, line 221: What is the reason for switching to specifying the certs and key values in the commandline? It seems more cumbersome than providing a file path. Particularly I am worried it may be hard to use on Windows (I personally tend to use zsh on MacOS and Unix).

tools/echo_server/echo_server.go
```diff
@@ -189,5 +242,22 @@ func main() {
...
+		if f.Name == portFlagName {
+			config.ServerPort = *serverPort
+		} else if f.Name == repeatFlagName {
+			config.RepeatMode = *repeatMode
+		} else if f.Name == intervalFlagName {
+			config.RepeatIntervalSeconds = *repeatIntervalSeconds
+		} else if f.Name == certFlagName {
+			config.ServerCert = *serverCert
+			config.Secure = true
+		} else if f.Name == keyFlagName {
```
lundinc2, line 256: If secure is enabled when specifying the key or server cert, we may as well just remove the secure option. Also should probably check that both are defined.

tools/echo_server/readme.md
```diff
@@ -47,15 +51,23 @@ The JSON file contains the following options:
...
 }
 ```
 # Running the Echo Server From the Command Line
 `go run echo_server.go`
 ## With a Custom Config Location
 `go run echo_server.go -config={config_file_path}`
+## With Repeat Mode on and a repeat interval of 3 seconds
+`go run echo_server.go -repeat=true -interval=3`
+## With Server Certificate and Key passed from command line args
+`go run echo_server.go -cert='<x509-cert-content>' -key='<x509-key-content>'`
```
lundinc2, line 68: This breaks the current config method, so will need to update the docs accordingly.

tools/echo_server/echo_server.go
```diff
@@ -113,12 +124,48 @@ func echoServerThread(port string, tlsConfig *tls.Config, verbose bool) {
...
+				go repeatEchoing(config, connection)
+			} else {
+				go readWrite(config, connection)
+			}
 		}
 	}
 }
 
-func readWrite(connection net.Conn, verbose bool) {
+func repeatEchoing(config *Argument, connection net.Conn) {
```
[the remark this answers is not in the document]
onsankawai: Will update this in next rev.

tools/echo_server/echo_server.go
onsankawai: Because in this way it is easy for the IDT to run the echo server on an EC2 instance with a user-data script directly.
But I agree that this will be inconvenient for user who are using the config file. I will restore the certPath / keyPath options in next rev.

In next rev both file paths in config file and cmdline args can be used, with cmdline args having a higher priority.
i.e. (pseudo code)

```
if certFromCmdline exist {
  cert = loadCert(certFromCmdline)
} else {
  cert = loadCertFromPath(certPath)
}

```

tools/echo_server/echo_server.go
onsankawai: I want to leave the secure option for users who want to use config file instead, it will be easier for them to toggle secure mode without making too many changes to the config file.

I agree that we should enable secure mode ONLY WHEN both cmdline args are present, will make this change in next rev.

tools/echo_server/readme.md
onsankawai: I will make this very clear in next rev:

Revised config file options:
```
"server-certificate": "<cert-content>",
"server-key": "<key-content>",
"server-certificate-location": "./certs/server.pem",
"server-key-location": "./certs/server.key"
```

New cmdline args:
`-cert` and `-key` 

Priority:
`-cert` > `server-certificate` in config file > `server-certificate-location` in config file
`-key` > `server-key` in config file > `server-key-location` in config file

tools/echo_server/readme.md
```diff
@@ -41,6 +41,10 @@ The JSON file contains the following options:
     1. The repeat interval in seconds. When in repeat mode, the echo server will wait for this amount of time until the next echoing.
 1. server-port
     1. Specify which port to open a socket on.
+1. server-certificate
+    1. Server certificate in X509 format. When both `server-certificate` and `server-certificate-location` are specified, this option will be used instead.
+1. server-key
+    1. Server private key in X509 format. When both `server-key` and `server-key-location` are specified, this option will be used instead.
```
lundinc2, line 47: Nit but the key will not be in X509, as it is a standard only related to certificates

tools/echo_server/readme.md
```diff
@@ -54,8 +58,11 @@ The JSON file contains the following options:
     "repeat-mode": false,
     "repeat-interval-seconds": 1,
     "server-port": "9000",
+    "server-certificate": "<cert-content>",
+    "server-key": "<key-content>",
     "server-certificate-location": "./certs/server.pem",
     "server-key-location": "./certs/server.key"
+
```
lundinc2, line 65: nit: new line should be removed

tools/echo_server/echo_server.go
```diff
@@ -58,13 +58,35 @@ type Argument struct {
 	ServerPort 				string `json:"server-port"`
 	ServerCert 				string `json:"server-certificate"`
 	ServerKey  				string `json:"server-key"`
+	ServerCertPath          string `json:"server-certificate-location"`
+	ServerKeyPath           string `json:"server-key-location"`
```
lundinc2, line 62: Do you mind running the go formatter to enforce the formatting

tools/echo_server/readme.md
onsankawai: fixed in next rev.

tools/echo_server/readme.md
onsankawai: fixed in next rev.

tools/echo_server/echo_server.go
onsankawai: fixed in next rev.

lundinc2 (approved):

lundinc2:
@onsankawai thanks for this contribution!

PR #2589, opened 2020-10-16
MQTT Shim Layer Pub/Sub thread safe

DanielYEHsieh:
Symptom:
MQTT Shim Layer removed the task pool and scheduling for publish/subscribe and disconnect operations.
Applications (tests/demos) invokes IotMqtt_Publish/IotMqtt_Subscribe and invokes MQTT Shim Layer with lower priority than the receive task will encounter issues when receive task gets an ACK and cleans up the pending operation before it is added into the list by the application.
t1. created operation.
t2. send out pkt.
t3. receive ack. (will not send complete since the operation is not in the list yet.)
t4. insert operation into pending list.
t5. application waits till timeout because the operation will not be notified as complete.

Same pattern happens to any Libraries who relies MQTT Shim Layer (example: Shadow).

Solution:
We've extend the mutex critical section for MQTT Shim Layer to avoid issues caused by tests/demos main thread priority lower than receive task.
libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
libraries/c_sdk/standard/mqtt/src/iot_mqtt_operation.c

For Shadows, other libraries like DEFENDER who are relying on MQTT Shim Layer, to avoid changing critical sections for all these libraries (which might involve more issues), we've will increase the demos/tests thread and make sure it is higher than receive task.
We've ensured the priority change on 5 boards (PC, ESP * 2, infineon * 2, Nuvoton), and will update other boards in another PR.

Results:
demos
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/1255
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/1256

tests (nuvoton eth board doesn't exists for tests others all passed)
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/1254


libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -733,7 +734,6 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
         IotListDouble_InsertHead( &( mqttConnection->pendingResponse ),
                                   &( pSubscriptionOperation->link ) );
 
-        IotMutex_Unlock( &( mqttConnection->referencesMutex ) );
 
         /* Processing operation after sending it on the network. */
         _IotMqtt_ProcessOperation( pSubscriptionOperation );
```
leegeth, line 739: This function has internally mutex lock and unlock on the same reference mutex. So this change may result in a deadlock.

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1761,10 +1763,9 @@ IotMqttError_t IotMqtt_Publish( IotMqttConnection_t mqttConnection,
         {
             EMPTY_ELSE_MARKER;
         }
-
-        IOT_GOTO_CLEANUP();
     }
 
+    IotMutex_Unlock( &( mqttConnection->referencesMutex ) );
```
leegeth, line 1768: Same comment as that of subscription.

DanielYEHsieh: fixed in 6894004

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
DanielYEHsieh: fixed in 6894004

leegeth (dismissed):

leegeth (approved):

leegeth (approved):

DanielYEHsieh:
As discussed we add back mainTEST_RUNNER_TASK_PRIORITY for Nuvoton to avoid thread safe issues in Shadow in commit cbc7302

DanielYEHsieh:
### Symptom:
MQTT Shim Layer removed the task pool and scheduling publish/subscribe and disconnect operations.
Applications (tests/demos) invokes IotMqtt_Publish/IotMqtt_Subscribe and invokes MQTT Shim Layer with lower priority than the receive task will encounter issues when receive task gets an ACK and cleans up the pending operation before it is added into the list by the application.
t1. created operation.
t2. send out pkt.
t3. receive ack.
t4. insert operation into pending list.
t5. application waits till timeout because the operation will not be notified as complete.

Same pattern happens to any Libraries who relies MQTT Shim Layer (example: Shadow).

### Solution:
1. We've extend the mutex critical section for MQTT Shim Layer to avoid issues caused by tests/demos main thread priority lower than receive task.
libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
libraries/c_sdk/standard/mqtt/src/iot_mqtt_operation.c

2. For Shadows, other libraries like DEFENDER who are relying on MQTT Shim Layer, to avoid changing critical sections for all these libraries (which might involve more issues), we've will increase the demos/tests thread and make sure it is higher than receive task.

aggarg (approved):

PR #2592, opened 2020-10-16
Remove assert for context index in destroy MQTT connection.

leegeth:
Remove assert for context index in destroy MQTT connection.

Description
-----------
Problem: In an OTA test an issue was observed when the MQTT connect was not successful. The program hit an assert.
Root cause: The MQTT connection was created without a MQTT context associated with it in an error case. Trying to remove the MQTT context resulted in an assert.
Fix: Remove the assert and replace it with conditional checks to expect the error case when destroying a connection.

Checklist:
----------

### DIFF
--- a/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
@@ -521,9 +521,17 @@ static void _destroyMqttConnection( _mqttConnection_t * pMqttConnection )
     int8_t contextIndex = -1;
     bool mutexStatus = true;
 
+    IotMqtt_Assert( pMqttConnection != NULL );
+
     /* Getting MQTT Context for the specified MQTT Connection. */
     contextIndex = _IotMqtt_getContextIndexFromConnection( pMqttConnection );
-    IotMqtt_Assert( contextIndex != -1 );
+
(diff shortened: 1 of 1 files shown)


sarenameas (approved):

n9wxu (approved):

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -521,9 +521,17 @@ static void _destroyMqttConnection( _mqttConnection_t * pMqttConnection )
...
     bool mutexStatus = true;
 
+    IotMqtt_Assert( pMqttConnection != NULL );
+
     /* Getting MQTT Context for the specified MQTT Connection. */
     contextIndex = _IotMqtt_getContextIndexFromConnection( pMqttConnection );
-    IotMqtt_Assert( contextIndex != -1 );
+
+    /* An MQTT connection can be created without having MQTT Context associated.
+     * This can happen in case of an error in #IotMqtt_Connect() API. */
```
aggarw13, line 530: How does the `_destroyMqttConnection` function get invoked if the `IotMqtt_Connect` API failed?

aggarw13: Thanks for the clarification.

PR #2612, opened 2020-10-21
Mbedtls buffer size increase

lundinc2:
Mbedtls buffer size increase

Description
-----------
Set mbed TLS configuration to the default max buffer size as defined by RFC.  

Checklist:
----------

- [x] My code is Linted.


libraries/3rdparty/mbedtls_config/aws_mbedtls_config.h
```diff
@@ -3246,7 +3246,7 @@
  * Uncomment to set the maximum plaintext size of the outgoing I/O buffer
  * independently of the incoming I/O buffer.
  */
-//#define MBEDTLS_SSL_OUT_CONTENT_LEN             16384
+#define MBEDTLS_SSL_OUT_CONTENT_LEN             8192
```
aggarw13, line 3249: Why is this change needed? If the goal is to use the fragment length of 16K, then mbedTLS requires that both the _in_ and _out_ buffer lengths to be set to a value of 16K...
From the mbedTLS documentation:
"_The MFL values that mbedtls_ssl_conf_max_frag_len() accepts are those defined in the standard: 512, 1024, 2048 and 4096 bytes. If mbedtls_ssl_conf_max_frag_len() is not called, the MFL defaults to the minimum of the size of the internal incoming and outgoing datagram buffers as configured by MBEDTLS_SSL_IN_CONTENT_LEN and MBEDTLS_SSL_OUT_CONTENT_LEN, respectively, which is at most 16384 bytes (the maximum record payload size according to the TLS/DTLS standards, see for example TLS 1.2 RFC 5246, Section 6.2.1 and DTLS 1.2 RFC 6347, Section 3.2.3)_"
```

lundinc2: We can save some RAM with this. If this causes a failure it's more obvious, as the client will fail the handshake. If the content in buffer is too small, it will just drop the fragment.

mbed TLS does not require the buffers be the same size, this is a recommended way for saving on some RAM in the case where you do not know what size fragments to expect from the server.

gkwicker:
Closing this one.

PR #2627, opened 2020-10-26
Add BLE dependency on MQTT for FreeRTOS console

aggarw13:
Equivalent of #2626

### DIFF
--- a/libraries/core_mqtt_demo_dependencies.cmake
+++ b/libraries/core_mqtt_demo_dependencies.cmake
@@ -100,3 +100,14 @@ if(TARGET AFR::wifi::mcu_port)
             AFR::wifi
     )
 endif()
+
+# Add dependency on BLE module so that the BLE library is auto-included
+# when selecting core MQTT library on FreeRTOS console for boards that
+# support BLE.
+if(BLE_SUPPORTED)
+    afr_module_dependencies(
+        ${AFR_CURRENT_MODULE}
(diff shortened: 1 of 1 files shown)


ravibhagavandas (approved):

libraries/core_mqtt_demo_dependencies.cmake
```diff
@@ -99,4 +99,15 @@ if(TARGET AFR::wifi::mcu_port)
         PUBLIC
             AFR::wifi
     )
+endif()
+
+# Add dependency on BLE module so that the BLE library is auto-included
+# when selecting core MQTT library on FreeRTOS console for boards that
```
gkwicker, line 105: Is this just for the console use case?  If someone just clones this repo and builds this demo, would they need lines 107-112?  If so I'd reword the comment on lines 104-106:

```sh
# Add dependency on BLE module for boards that support BLE.
```

gkwicker (changes requested):
See comment

libraries/core_mqtt_demo_dependencies.cmake
aggarw13: Nope, these changes do not affect the demo configuration as the demos themselves have the `AFR::ble` dependency specified in their CMake configuration. 
These changes are purely for the FreeRTOS console functionality.

gkwicker (approved):

PR #2629, opened 2020-10-26
Remove unused client credentials configuration in HTTP demos

aggarw13:
The HTTP demos connecting to S3 contain connection congifuration logic with client certificate and private key. This is not required as the AWS S3 TLS connection does not require client certificate. The client authentication is performed through use of _pre-signed URLs_ in the demos instead.

### DIFF
--- a/demos/https/iot_demo_https_s3_download_async.c
+++ b/demos/https/iot_demo_https_s3_download_async.c
@@ -1086,7 +1086,9 @@ int RunHttpsAsyncDownloadDemo( bool awsIotMqttMode,
         IOT_SET_AND_GOTO_CLEANUP( EXIT_FAILURE );
     }
 
-    /* Set the connection configurations. */
+    /* Set the connection configurations.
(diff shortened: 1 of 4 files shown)


demos/https/iot_demo_https_s3_download_async.c
```diff
@@ -1086,18 +1086,16 @@ int RunHttpsAsyncDownloadDemo( bool awsIotMqttMode,
         IOT_SET_AND_GOTO_CLEANUP( EXIT_FAILURE );
     }
 
-    /* Set the connection configurations. */
+    /* Set the connection configurations.
+     * Note: TLS Connection to AWS S3 service does not client certificate.
```
sarenameas, line 1090:
```suggestion
     * Note: TLS Connection to AWS S3 service does not need a client certificate.
```

sarenameas (approved):

gkwicker (approved):

PR #2637, opened 2020-10-27
Fix TI issue of Server Cert validation in MQTT demos

aggarw13:
### Problem

MQTT Mutual Auth and MQTT Connection sharing demos, and the `coreMQTT_Integration_AWS_IoT_Compatible` tests fail on the TI board with the following error:
```
21 43338 [pthread] [SimpleLinkSockEventHandler ERROR]: Root CA in file system did not sign the chain.

22 43349 [pthread] [SimpleLinkSockEventHandler ERROR]: Handshake Failed - socket 0, reason: -688.
```

### Cause
The TI board requires that a custom Root CA provided have its certificate self-signed (i.e. the Root CA not be "_issued" by another CA), otherwise, it treats the CA as an _intermediate_ CA. 
More details on the TI board TLS stack is provided in this customer issue on the TI forum: https://e2e.ti.com/support/wireless-connectivity/wifi/f/968/t/789112\?CC3220SF-SL-ERROR-BSD-ESEC-ASN-NO-SIGNER-E-688-

### Solution
Update the TI specific config files for the demos to use the ** Starfield Class 2 Certification Authority** as the Root CA.

### DIFF
--- a/demos/coreMQTT/mqtt_demo_basic_tls.c
+++ b/demos/coreMQTT/mqtt_demo_basic_tls.c
@@ -84,7 +84,7 @@
  * @brief The root CA certificate belonging to the broker.
  */
 #ifndef democonfigROOT_CA_PEM
-    #define democonfigROOT_CA_PEM    tlsATS1_ROOT_CERTIFICATE_PEM
+    #define democonfigROOT_CA_PEM    "Please provide Root CA for the MQTT Basic TLS demo."
 #endif
 
 /**
--- a/demos/coreMQTT/mqtt_demo_basic_tls.c
+++ b/demos/coreMQTT/mqtt_demo_basic_tls.c
@@ -82,9 +82,13 @@
 
 /**
  * @brief The root CA certificate belonging to the broker.
+ *
+ * @note This demo cannot be run against AWS IoT as it requires client authentication
+ * for TLS connection. Please define a valid value of the server Root CA certificate relevant to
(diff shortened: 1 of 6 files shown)


demos/coreMQTT/mqtt_demo_basic_tls.c
```diff
@@ -84,7 +84,7 @@
  * @brief The root CA certificate belonging to the broker.
  */
 #ifndef democonfigROOT_CA_PEM
-    #define democonfigROOT_CA_PEM    tlsATS1_ROOT_CERTIFICATE_PEM
```
lundinc2, line 87: Maybe should be a #error

lundinc2 (dismissed):

demos/coreMQTT/mqtt_demo_basic_tls.c
aggarw13: Have to use `#define` instead of `#error` until the "PR Build checks" can be updated to define this macro at build time. Once, the CI job is updated to define this macro, I will change it to a `#error` for better user experience about the significance of the macro.

lundinc2 (approved):

demos/coreMQTT/mqtt_demo_basic_tls.c
cobusve: Just take care because #error is not part of the C standard, it may not work on all compilers

cobusve (approved):

PR #2639, opened 2020-10-27
afr.cmake: Add BUILD_CLONE_SUBMODULES option

pfalcon:
This option is similar to the option under the same name already used
in some submodules of the repository. If set to OFF, it skips automatic
submodule update during building. This is useful for focused builds
of particular targets, which use only a subset of submodules (which
then need to be updated externally). In particular in continuous
integration contexts, this can save build time/bandwidth.

Signed-off-by: Paul Sokolovsky <paul.sokolovsky@linaro.org>

Description
-----------

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tools/cmake/afr.cmake
+++ b/tools/cmake/afr.cmake
@@ -27,19 +27,23 @@ set(NETWORK_MANAGER_SOURCES
     "Network manager common source files."
 )
 
+option(BUILD_CLONE_SUBMODULES "Clone any required Git submodules. When OFF, submodules must be manually cloned." ON)
+
 # Set regular version and Git commit version.
 set(AFR_VERSION "${PROJECT_VERSION}")
 set(AFR_VERSION_VCS "Unknown" CACHE INTERNAL "")
 # Check if we're in a Git repository.
 find_package(Git)
 if(Git_FOUND AND EXISTS "${AFR_ROOT_DIR}/.git")
-    message(STATUS "Submodule update")
-    # TODO: Update submodule only if it hasn't been checked out (check if directory is empty).
-    execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-                    RESULT_VARIABLE GIT_SUBMOD_RESULT)
-    if(NOT GIT_SUBMOD_RESULT EQUAL "0")
-        message(FATAL_ERROR "git submodule update --init failed with ${GIT_SUBMOD_RESULT}, please checkout submodules")
+    if(${BUILD_CLONE_SUBMODULES})
+        message(STATUS "Submodule update")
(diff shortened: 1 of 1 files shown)


pfalcon:
Diff with whitespace changes filtered can show that changes are indeed minimal: https://github.com/aws/amazon-freertos/pull/2639/files?w=1

pfalcon:
@galak: FYI

ravibhagavandas:
@pfalcon 

Thanks for submitting the PR. Could you rename the option to `CLONE_SUBMODULES` to be consistent with other variables.

pfalcon:
@ravibhagavandas: Thanks for the review, and I'm certainly happy to make any changes to get this into mergeable state. I however was curious what you mean by "to be consistent with other variables". Looking at the exact file being modified, https://github.com/aws/amazon-freertos/blob/master/tools/cmake/afr.cmake, I see that it defines a few other CMake `option`s, but the only thing common among them seems to be that all they have `AFR_` prefix.

On the other hand, as I mentioned in the original description ("This option is similar to the option under the same name already used in some submodules of the repository."), the option named `BUILD_CLONE_SUBMODULES` is used in a number of git submodules which are included in this repository. Let me provide specific list:

* device-shadow-for-aws-iot-embedded-sdk: https://github.com/aws/device-shadow-for-aws-iot-embedded-sdk/blob/master/test/CMakeLists.txt#L23
* coreMQTT: https://github.com/FreeRTOS/coreMQTT/blob/master/test/CMakeLists.txt#L23
* coreJSON: https://github.com/FreeRTOS/coreJSON/blob/master/test/CMakeLists.txt#L25
* FreeRTOS-Plus-TCP: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/master/test/unit-test/CMakeLists.txt#L23
* corePKCS11: https://github.com/FreeRTOS/corePKCS11/blob/master/test/unit-test/CMakeLists.txt#L23

So, BUILD_CLONE_SUBMODULES is used in 5 other `amazon-freertos` submodules, and that's why I decided to use the same name here. Again, I'm happy to change it to whatever required, just taking a chance to understand amazon-freertos's conventions better.

What I did for now is to add a 2nd commit with the proposed rename. If you confirm we should go ahead with it, I'll squash it into the first commit. Otherwise, it can be dropped again. Thanks.

ravibhagavandas:
@pfalcon 

Thanks for the clarification. I felt that `BUILD_` seemed redundant and just `CLONE_SUBMODULES` or `AFR_CLONE_SUBMODULES` seemed a better naming.

Having said that I do see its being used at other places in CMakeFiles, thanks for pointing that. We can keep that same name to be consistent and drop the new commit for that.

pfalcon:
> We can keep that same name to be consistent and drop the new commit for that.

Thanks, re-pushed.

ravibhagavandas (approved):

leegeth (approved):

cobusve (approved):

PR #2640, opened 2020-10-27
Fix MISRA error for transport secure socket

andysun2015:
Description
-----------

Fix the following MISRA errors in transport secure socket:
- misra c 5.2
- misra c 10.4
- misra c 14.3
- misra c 15.7

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
+++ b/libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
@@ -96,7 +96,7 @@ static TransportSocketStatus_t connectToServer( Socket_t tcpSocket,
 
 /*-----------------------------------------------------------*/
 
-int32_t SecureSocketsTransport_Send( const NetworkContext_t * pNetworkContext,
+int32_t SecureSocketsTransport_Send( NetworkContext_t * pNetworkContext,
                                      const void * pMessage,
                                      size_t bytesToSend )
 {
@@ -145,7 +145,7 @@ int32_t SecureSocketsTransport_Send( const NetworkContext_t * pNetworkContext,
 
 /*-----------------------------------------------------------*/
 
-int32_t SecureSocketsTransport_Recv( const NetworkContext_t * pNetworkContext,
+int32_t SecureSocketsTransport_Recv( NetworkContext_t * pNetworkContext,
                                      void * pBuffer,
                                      size_t bytesToRecv )
 {
--- a/libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
+++ b/libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
@@ -215,9 +215,9 @@ static int32_t tlsSetup( const SocketsConfig_t * pSocketsConfig,
 {
     int32_t secureSocketStatus = SOCKETS_ERROR_NONE;
 
-    configASSERT( tcpSocket != SOCKETS_INVALID_SOCKET );
-    configASSERT( pSocketsConfig != NULL );
-    configASSERT( pHostName != NULL );
+    configASSERT( (tcpSocket != SOCKETS_INVALID_SOCKET) ? pdTRUE:pdFALSE );
+    configASSERT( (pSocketsConfig != NULL) ? pdTRUE:pdFALSE );
+    configASSERT( (pHostName != NULL) ? pdTRUE:pdFALSE );
 
     /* ALPN options for AWS IoT. */
     /* ppcALPNProtos is unused. putting here to align behavior in IotNetworkAfr_Create. */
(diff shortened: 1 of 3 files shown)


libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
```diff
@@ -96,7 +96,7 @@ static TransportSocketStatus_t connectToServer( Socket_t tcpSocket,
 
 /*-----------------------------------------------------------*/
 
-int32_t SecureSocketsTransport_Send( NetworkContext_t * pNetworkContext,
+int32_t SecureSocketsTransport_Send( const NetworkContext_t * pNetworkContext,
```
yourslab, line 99: `transport_interface.h` was updated to remove the `const` fron send/recv. I know this is fixing MISRA 8.13, but this rule is just advisory. Other transport send/recv implementations (like in product 3) need to modify the `NetworkContext` object. I think this warning should just be suppressed.

andysun2015: If that's our common consensus, then I will remove the ```const```.

libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
```diff
@@ -210,9 +214,9 @@ static int32_t tlsSetup( const SocketsConfig_t * pSocketsConfig,
 {
     int32_t secureSocketStatus = SOCKETS_ERROR_NONE;
 
-    configASSERT( tcpSocket != SOCKETS_INVALID_SOCKET );
-    configASSERT( pSocketsConfig != NULL );
-    configASSERT( pHostName != NULL );
+    configASSERT( ( tcpSocket != SOCKETS_INVALID_SOCKET ) ? pdTRUE : pdFALSE );
```
muneebahmed10, line 217: I don't understand why we need this ternary here. In the C SDK, we disable asserts before checking MISRA. Can the same be done here? Otherwise, this looks awkward to use every time we use a `configASSERT`

libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
```diff
@@ -160,7 +160,7 @@ int32_t SecureSocketsTransport_Recv( NetworkContext_t * pNetworkContext,
                     pBuffer, bytesToRecv, ( void * ) pNetworkContext ) );
         bytesReceived = SOCKETS_EINVAL;
     }
-    else if( ( pNetworkContext != NULL ) && ( pNetworkContext->tcpSocket == SOCKETS_INVALID_SOCKET ) )
+    else if( pNetworkContext->tcpSocket == SOCKETS_INVALID_SOCKET )
```
muneebahmed10, line 163: Looks like the 11.2 violations is due to `SOCKETS_INVALID_SOCKET`, which is `#define`d as `SOCKETS_INVALID_SOCKET    ( ( Socket_t ) ~0U )`, so this violation will appear every time the `SOCKETS_INVALID_SOCKET` macro is used. Since this macro is part of secure sockets, can allow this for now

andysun2015: This happened due to the incomplete type from ```Socket_t ``` checked by 11.2 and seems like only happened in windows toolchain instead of linux toolchain. There may be something wrong in windows toolchain because this structure should be defined. So I think this one should be allowed.

libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
andysun2015: If so, I can disable it because I run MISRA with ```configASSERT```. Thanks for your comment.

libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
```diff
@@ -145,6 +149,10 @@ int32_t SecureSocketsTransport_Send( NetworkContext_t * pNetworkContext,
 
 /*-----------------------------------------------------------*/
 
+/* MISRA Rule 18.3 flags the following line for not using the const qualifier
+ * on `pNetworkContext`. Indeed, the object pointed by it is not modified
+ * by Secure Sockets, but other implementations of `TransportRecv_t` may do so. */
+/* coverity[misra_c_2012_rule_8_13_violation] */
```
muneebahmed10, line 155: Same comment as the other PR, if it doesn't actually suppress the violations, we should leave it out

muneebahmed10 (approved):

yourslab (approved):

cobusve (approved):

PR #2647, opened 2020-10-28
Fix MISRA error for transport secure socket 

andysun2015:
Description
-----------

Fix the following MISRA errors in transport secure socket:
- misra c 5.2
- misra c 10.4
- misra c 14.3
- misra c 15.7

Checklist:
----------

- [ ] My code is Linted.


yourslab:
I think we still need to suppress 8.13 above the function definition. Maybe also for plaintext. This is how I did it in CSDK:
https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1347/files#diff-28f0f9b269a9715d8bf94b70d182c8d5aefba2d6f361764cea2efaf735b7438cR665-R669

libraries/abstractions/transport/secure_sockets/transport_secure_sockets.c
```diff
@@ -145,6 +149,10 @@ int32_t SecureSocketsTransport_Send( NetworkContext_t * pNetworkContext,
 
 /*-----------------------------------------------------------*/
 
+/* MISRA Rule 18.3 flags the following line for not using the const qualifier
+ * on `pNetworkContext`. Indeed, the object pointed by it is not modified
+ * by Secure Sockets, but other implementations of `TransportRecv_t` may do so. */
+/* coverity[misra_c_2012_rule_8_13_violation] */
```
muneebahmed10, line 155: In my experience, this does not actually work in suppressing 8.13 violations. It'd be better to remove this message if it doesn't actually work in doing so

yourslab: I would say remove `/* coverity[misra_c_2012_rule_8_13_violation] */`, but the comment can stay. Although, this says 18.3 instead of 8.13.

andysun2015: oops~my bad, thanks for the correction.

muneebahmed10 (approved):

yourslab (approved):

cobusve (approved):

PR #2653, opened 2020-10-28
NimBLE AFR: Save full AFR service instead of only service pointers

prasad-alatkar:
AFR: Allocate memory to save services internally instead of depending on upper layer. 

Description
-----------

- For each service allocate full AFR service space (`BTService_t`), save the content received from upper layer.
- Free the space when cleaning up GATT.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -514,6 +514,7 @@ void vESPBTGATTServerCleanup( void )
         for( index = 0; index < serviceCnt; index++ )
         {
             prvCleanupService( &espServices[ index ] );
+            vPortFree( ( void * ) afrServices[ index ] );
         }
 
         serviceCnt = 0;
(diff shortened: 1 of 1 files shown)


prasad-alatkar:
@ravibhagavandas PTAL. This change is needed when upper layer does not want to own responsibility of `pxService` allocated. ( CC: @dhrishi @mahavirj)

vendors/espressif/boards/esp32/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -856,7 +858,17 @@ BTStatus_t prvAddServiceBlob( uint8_t ucServerIf,
         if( serviceCnt < MAX_SERVICES )
         {
             pSvc = &espServices[ serviceCnt ];
-            afrServices[ serviceCnt ] = pxService;
+            afrFullService = pvPortCalloc( 1, sizeof( BTService_t ) );
```
ravibhagavandas, line 861: @prasad-alatkar Could you explain why an internal copy was required. As far as my understanding these pointers are to static memory allocated throughout program context.

prasad-alatkar: @ravibhagavandas , the original request came from DPK team regarding this. Though the pointers are pointing to static memory (i.e. `afrServices`), but the array `afrServices` points to `pxService` received from upper layer in `prvAddServiceBlob`. Now, the upper layer may free this memory at their will (apparently the case with DPK), so to avoid dependency on upper layer I believe this PR is needed. Please let me know your thoughts on this.

ravibhagavandas (approved):

aggarg (approved):

cobusve (approved):

PR #2654, opened 2020-10-28
Add mutex around connection to context array in MQTT shim

leegeth:
Add mutex around connection to context array in MQTT shim

Description
-----------
Add mutex around connection to context array in MQTT shim.
This is to provide thread safety while updating the connection to context array.
This is to provide thread safety when more than one connection is created using MQTT shim.

Checklist:
----------

### DIFF
--- a/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
@@ -621,8 +621,8 @@ static void _destroyMqttConnection( _mqttConnection_t * pMqttConnection )
             /* Free the MQTT Context for the MQTT connection to be destroyed. */
             _IotMqtt_removeContext( pMqttConnection );
 
-            /* Lock mutex before updating the #connToContext array. */
-            connContextMutexStatus = IotMutex_TakeRecursive( &connContextMutex );
+            /* Release mutex. */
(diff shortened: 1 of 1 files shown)


yourslab (dismissed):

libraries/c_sdk/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -606,8 +613,19 @@ static void _destroyMqttConnection( _mqttConnection_t * pMqttConnection )
...
+        /* Lock mutex before updating the #connToContext array. */
+        connContextMutexStatus = IotMutex_TakeRecursive( &connContextMutex );
+
+        if( connContextMutexStatus == true )
+        {
+            /* Free the MQTT Context for the MQTT connection to be destroyed. */
+            _IotMqtt_removeContext( pMqttConnection );
+
+            /* Lock mutex before updating the #connToContext array. */
+            connContextMutexStatus = IotMutex_TakeRecursive( &connContextMutex );
```
dan4thewin, line 625: Do you need to take it again?  Should this be a give?

leegeth: No. Thanks for catching it. Will fix it.

dan4thewin (approved):

PR #2658, opened 2020-10-29
Fix CMake configuration of OTA demo

aggarw13:
The OTA demo has a hard-dependency on the old MQTT library. This PR updates the CMake configuration of the demo to have a direct dependency on `AFR::mqtt` target.


tgsong (approved):

aggarw13:
Closing the PR as the same change was made in #2628

PR #2665, opened 2020-10-29
Test: Fix unit test not building

alfred2g:
Fix Unit Test not building

Description
-----------
Fix the build framework for the UT, transport, and ble

Checklist:
----------


libraries/c_sdk/standard/ble/include/iot_ble_mqtt_serialize.h
```diff
@@ -32,6 +33,8 @@
 #ifndef IOT_BLE_MQTT_SERIALIZE_H
 #define IOT_BLE_MQTT_SERIALIZE_H
 
+#define bool    int
```
aggarw13, line 36: This seems like a dangerous change... It wouldn't work for compilers that have C99 support.

alfred2g: bool won't work with c89 compilers 
so we need to fix all this file

libraries/c_sdk/standard/ble/utest/ble_mqtt_transport/CMakeLists.txt
```diff
@@ -64,6 +66,8 @@
             "${AFR_ROOT_DIR}/freertos_kernel/include/"
             "${AFR_ROOT_DIR}/libraries/c_sdk/standard/common/include"
             "${AFR_ROOT_DIR}/libraries/coreMQTT/source/include"
+            "${AFR_ROOT_DIR}/libraries/coreMQTT/test/unit-test"
```
aggarw13, line 69: Why does this need to have the `coreMQTT/test/unit-test` path in the include list?

libraries/abstractions/transport/CMakeLists.txt
```diff
@@ -35,4 +35,3 @@ afr_module_include_dirs(
        "${transport_interface_dir}"
        "${src_dir}"
 )
-
```
aggarw13, line 38: We usually do put newlines at the end of files...

aggarw13 (dismissed):

aggarg (dismissed):

aggarg (approved):

PR #2681, opened 2020-11-02
Build and check CBMC proofs concurrently with litani

markrtuttle:
This pull request modifies the method of building and checking CBMC proofs to use litani.  

Litani is a build system based on ninja that allows us to build and check CBMC proofs concurrently, and to produce a summary that gives the current status of proof progress, and ends with an overall summary of the proof results.  Moving to a litani build also makes it possible to move to a new implementation of continuous integration.

### DIFF
--- a/tools/cbmc/proofs/Makefile.template
+++ b/tools/cbmc/proofs/Makefile.template
@@ -50,16 +50,6 @@ INSTFLAGS = \
 # ____________________________________________________________________
 # Rules
 #
-# Rules for building a:FR object files. These are not harness-specific,
-# and several harnesses might depend on a particular a:FR object, so
-# define them all once here.
-
-@RULE_GOTO@
-	$(GOTO_CC) @COMPILE_ONLY@ @RULE_OUTPUT@ $(INC) $(CFLAGS) @RULE_INPUT@
-
-# ____________________________________________________________________
-# Rules
-#
 # Rules for patching
 
 patch:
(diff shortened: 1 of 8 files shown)


tools/cbmc/proofs/Makefile.template
```diff
@@ -73,14 +63,27 @@ unpatch:
...
+CWD=$(abspath .)
+
+gotos/%.goto: %.c
+	mkdir -p $(dir $@)
 	$(GOTO_CC) @COMPILE_ONLY@ @RULE_OUTPUT@ $(INC) $(CFLAGS) @RULE_INPUT@
 
-$(ENTRY)1.goto: $(OBJS)
-	$(GOTO_CC) @COMPILE_LINK@ @RULE_OUTPUT@ --function harness $(OBJS)
+queue_datastructure.h: gotos/$(FREERTOS)/freertos_kernel/queue.goto
+	python3 @TYPE_HEADER_SCRIPT@ --binary $(CWD)/gotos$(FREERTOS)/freertos_kernel/queue.goto --c-file $(FREERTOS)/freertos_kernel/queue.c
```
karkhaz, line 80: * There should be a `/` in `$(CWD)/gotos$(FREERTOS)` between `gotos` and `$(FREERTOS)`
* Currently, the prerequisite is `gotos/$(FREERTOS)/...` but the argument is `$(CWD)/gotos/...`. I would prefer if the argument was just `$@` so that we don't write the same thing twice and have it go out-of-date. I believe there's no need for `$(CWD)` because this command will be run in the proof directory, but IMO you should still have it there just so that an absolute path shows up in the log.

tools/cbmc/proofs/Makefile.template
```diff
@@ -110,11 +113,13 @@ $(ENTRY).goto: $(ENTRY)5.goto
 # Rules for running CBMC
 
 goto:
```
karkhaz, line 115: Do you not want to do the `_goto`, `_report`, etc trick that we do for other repositories, so that people can just run `make report` in the proof directory and have it 'just work' with litani?

tools/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,313 @@
...
+def main():
+    args = get_args()
+    set_up_logging(args.verbose)
+
+    proof_root = pathlib.Path(__file__).resolve().parent
+    litani = get_litani_path(proof_root)
+
+    if not args.no_standalone:
+        run_cmd(
+            ["./prepare.py"], check=True, cwd=str(proof_root))
```
karkhaz, line 281: Is this idempotent now?

karkhaz (changes requested):

tools/cbmc/proofs/Makefile.template
markrtuttle: I have addressed all of these issues with `--binary @RULE_INPUT@`.  This required modifying the type header script to expand this relative path to an absolute patch before running the goto-instrument command that uses this path.  The script changes directory to a temporary directory before running goto-instrument, so an absolute path is required, probably explaining  the reason for the $(CWD) prefix used in the prior version of this Makefile.template.

tools/cbmc/proofs/Makefile.template
markrtuttle: No, we are not changing the targets themselves.  We are using litani to run `make goto`, we are not using `make goto` to run litani as we do in the starter kit.

tools/cbmc/proofs/run-cbmc-proofs.py
markrtuttle: Yes, the patch module that this script calls now checks for the patching touch file before patching.

karkhaz (approved):

nateglims (approved):

n9wxu (approved):

PR #2687, opened 2020-11-03
Fix MQTT LWT tests failure due to MQTT keep alive timeout

leegeth:
Fix MQTT LWT tests failure due to MQTT keep alive timeout.

Description
-----------
Fix MQTT LWT tests failure due to MQTT keep alive timeout.
Problem: LWT tests were consistently failing on boards with slower stack.
Root cause: The MQTT connection established were getting timing out and closed by the broker on boards with slower network stacks.
Fix: Increase the keep alive timeout.

Checklist:
----------

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -212,7 +212,7 @@
  * @brief Time interval in seconds at which an MQTT PINGREQ need to be sent to
  * broker.
  */
-#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS      ( 5U )
+#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS      ( 20U )
 
 /**
  * @brief The number of milliseconds to wait for AWS IoT Core Message Broker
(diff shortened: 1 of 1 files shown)


tests/integration_test/core_mqtt_system_test.c
```diff
@@ -212,7 +212,7 @@
  * @brief Time interval in seconds at which an MQTT PINGREQ need to be sent to
  * broker.
  */
-#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS      ( 5U )
+#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS      ( 20U )
```
muneebahmed10, line 215: Is it the broker that was disconnecting us? If a value < 30 is given to AWS IoT, the server should treat it as if it were 30 seconds. Does this value need to be increased more?

muneebahmed10 (dismissed):

tests/integration_test/core_mqtt_system_test.c
leegeth: Thanks for sending the [documentation pointing out this behavior in AWS IoT](https://docs.amazonaws.cn/en_us/general/latest/gr/iot-core.html#iot-protocol-limits). I have seen that 20 seconds fixes the issue already, while it was consistently failing for 5. I have tested with 20 and 60 and both were successful. I will make it 60 to make the test more robust.

muneebahmed10 (approved):
Seems a bit long for the keep alive test, but we can refactor the tests at a later time to not take so much time

aggarg (approved):

PR #2725, opened 2020-11-09
Fix memory leak in tearDown of integration tests

yourslab:
Description
-----------

If the assertion on the returned status of `MQTT_Disconnect` fails, then `SecureSocketsTransport_Disconnect` will never be called. This means that the memory will remain allocated while a new connection will be established for the next test case, causing a memory leak.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -883,17 +883,19 @@ void testSetUp()
 /* Called after each test method. */
 void testTearDown()
 {
-    MQTTStatus_t status;
+    MQTTStatus_t mqttStatus;
+    TransportSocketStatus_t transportStatus;
 
     /* Terminate MQTT connection. */
-    status = MQTT_Disconnect( &context );
+    mqttStatus = MQTT_Disconnect( &context );
 
     /* Terminate TLS session and TCP connection. */
-    ( void ) SecureSocketsTransport_Disconnect( &networkContext );
+    transportStatus = SecureSocketsTransport_Disconnect( &networkContext );
 
-    /* Make any assertions at the end that may prevent #SecureSocketsTransport_Disconnect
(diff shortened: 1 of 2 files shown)


aggarw13 (dismissed):

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -883,11 +883,17 @@ void testSetUp()
...
 {
+    MQTTStatus_t status;
+
     /* Terminate MQTT connection. */
-    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_Disconnect( &context ) );
+    status = MQTT_Disconnect( &context );
 
     /* Terminate TLS session and TCP connection. */
     ( void ) SecureSocketsTransport_Disconnect( &networkContext );
+
```
gkwicker, line 893: Why are we throwing away the return value from `SecureSocketsTransport_Disconnect()` on line 892?

tests/integration_test/shadow_system_test.c
```diff
@@ -817,11 +817,17 @@ TEST_SETUP( deviceShadow_Integration )
...
 {
+    MQTTStatus_t status;
+
     /* Terminate MQTT connection. */
-    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_Disconnect( &context ) );
+    status = MQTT_Disconnect( &context );
 
     /* Terminate TLS session and TCP connection. */
     ( void ) SecureSocketsTransport_Disconnect( &networkContext );
+
```
gkwicker, line 827: Same comment as previous file – why are we discarding the return value on line 826?  It seems like we should be validating it just as we validate the return value of `MQTT_Disconnect()`.

gkwicker (commented):
See comments.

aggarw13 (approved):

gkwicker (approved):

PR #2741, opened 2020-11-12
Enable logging in the coreMQTT library for integration tests only

sarenameas:
This PR enables logging in the MQTT library only when the coreMQTT integration tests are enabled.
Ci builds one test at a time and therefore MQTT unit tests won't have logging enabled.

This PR increases the MQTT_Process loop timeout a bit and forces logging in the MQTT library for integration tests. This helps us to see if errors are from MQTT_Process loop not having enough time (which has been the case with manual inspection). This will also print any network errors when trying to connect.

The integration tests do not force error cases like the unit tests do, as functionality is tested against a third-party MQTT broker. Because error inducing cases are not part of the integration  tests, errors logs in the library are less likely to clobber Unity PASS and FAIL prints.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -80,7 +80,7 @@
 #endif
 
 #ifndef LIBRARY_LOG_LEVEL
-    #define LIBRARY_LOG_LEVEL    LOG_ERROR
+    #define LIBRARY_LOG_LEVEL    LOG_DEBUG
 #endif
 
 #include "logging_stack.h"
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -80,7 +80,7 @@
 #endif
 
 #ifndef LIBRARY_LOG_LEVEL
-    #define LIBRARY_LOG_LEVEL    LOG_DEBUG
+    #define LIBRARY_LOG_LEVEL    LOG_INFO
 #endif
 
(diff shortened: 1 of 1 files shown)


yourslab (dismissed):

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -226,7 +243,7 @@
  * PUBLISH message and ack responses for QoS 1 and QoS 2 communications
  * with the broker.
  */
-#define MQTT_PROCESS_LOOP_TIMEOUT_MS          ( 700U )
+#define MQTT_PROCESS_LOOP_TIMEOUT_MS          ( 1000U )
```
yourslab, line 246: Although, this increase seems insignificant. What if we try increasing the timeout to 10 seconds? Based on the data, the latency can possibly be that long.

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -62,22 +58,43 @@
...
 #ifndef LIBRARY_LOG_NAME
     #define LIBRARY_LOG_NAME    "TEST"
 #endif
 
+#ifndef LIBRARY_LOG_LEVEL
+    #define LIBRARY_LOG_LEVEL    LOG_DEBUG
+#endif
+
+/* Logging configuration for the MQTT library. */
+#ifndef LIBRARY_LOG_NAME
```
sarenameas, line 78: Just realized this won't work.

PR #2744, opened 2020-11-13
Increase all coreMQTT integration test timeouts.

sarenameas:
All intermittent failures in the last release were due to not receiving ACKs post MQTT_ProcessLoop and not receiving the CONNACK in MQTT_Connect. 

This increases the timeouts to 10 seconds in all areas to reduce failure probability. The MQTT keep alive timeout remains the same at 60 seconds. No other tests have a dependency on these timeouts. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -201,12 +201,12 @@
 /**
  * @brief Transport timeout in milliseconds for transport send and receive.
  */
-#define TRANSPORT_SEND_RECV_TIMEOUT_MS        ( 200U )
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS        ( 1000U )
 
 /**
  * @brief Timeout for receiving CONNACK packet in milli seconds.
  */
-#define CONNACK_RECV_TIMEOUT_MS               ( 1000U )
+#define CONNACK_RECV_TIMEOUT_MS               ( 10000U )
 
 /**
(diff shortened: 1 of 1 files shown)


yourslab (approved):

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -226,7 +226,7 @@
  * PUBLISH message and ack responses for QoS 1 and QoS 2 communications
  * with the broker.
  */
-#define MQTT_PROCESS_LOOP_TIMEOUT_MS          ( 700U )
+#define MQTT_PROCESS_LOOP_TIMEOUT_MS          ( 10000U )
```
muneebahmed10, line 229: Increasing the wait from 0.7 to 10 seconds every time the process loop is called seems like it would waste time in a lot of cases. If the only reason for this is to ensure that an ack is received, then wouldn't this be more efficient?
```c
void waitForAck( bool *flag ) {
    int counter = 0;
    TEST_ASSERT_FALSE( *flag );
    while( !(*flag) && ( counter++ < 10 ) ) {
        /* Event callback will set flag when receiving appropriate ack. This will
         * wait for at most TRANSPORT_SEND_RECV_TIMEOUT_MS. */
        TEST_ASSERT_EQUAL( MqttSuccess, MQTT_ProcessLoop( &context, 0 ) );
    }
    TEST_ASSERT_TRUE( *flag );
}
  
```

sarenameas: Sure thanks

PR #2749, opened 2020-11-14
coreMQTT_integration Tests: Implement polling up to a count on the blocking function, MQTT_ProcessLoop.

sarenameas:
This PR moves all test asserts that check for a flag to be set, to a function that polls on MQTT_ProcessLoop. 

Flags are set in the event callback invoked from MQTT_ProcessLoop successfully receiving a packet from the network.

97% of intermittent test failures in the last release were because of MQTT_ProcessLoop and MQTT_Connect not having enough time to receive certain specified packets. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -1709,10 +1709,7 @@ TEST( coreMQTT_Integration, Publish_With_Retain_Flag )
 
     /* Make sure that the library invoked the event callback with the incoming PUBLISH from
      * the broker containing the "retained" flag set. */
-    if( !receivedRetainedMessage )
-    {
-        waitForPacket( &receivedRetainedMessage, MQTTSuccess );
-    }
+    waitForPacket( &receivedRetainedMessage, MQTTSuccess );
 
     /* Reset the global variables for the remainder of the test. */
     receivedPubAck = false;
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -463,8 +463,8 @@ static void resumePersistentSession();
  * @param[in] pFlag Packet flag to test for true.
  * @param[in] expectedStatus Status code expected to be returned from MQTT_Process
  */
-static inline void waitForPacket( bool * pFlag,
-                                  MQTTStatus_t expectedStatus );
+static void waitForPacket( bool * pFlag,
+                           MQTTStatus_t expectedStatus );
 
 /*-----------------------------------------------------------*/
 
@@ -883,8 +883,8 @@ static void resumePersistentSession()
     TEST_ASSERT_TRUE( persistentSession );
 }
 
-static inline void waitForPacket( bool * pFlag,
-                                  MQTTStatus_t expectedStatus )
+static void waitForPacket( bool * pFlag,
+                           MQTTStatus_t expectedStatus )
 {
     uint8_t count = 0;
     MQTTStatus_t currentStatus = MQTTSuccess;
@@ -1709,7 +1709,10 @@ TEST( coreMQTT_Integration, Publish_With_Retain_Flag )
 
     /* Make sure that the library invoked the event callback with the incoming PUBLISH from
      * the broker containing the "retained" flag set. */
-    waitForPacket( &receivedRetainedMessage, MQTTSuccess );
+    if( !receivedRetainedMessage )
+    {
+        waitForPacket( &receivedRetainedMessage, MQTTSuccess );
+    }
 
     /* Reset the global variables for the remainder of the test. */
     receivedPubAck = false;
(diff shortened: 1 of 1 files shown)


tests/integration_test/core_mqtt_system_test.c
```diff
@@ -1688,22 +1698,18 @@ TEST( coreMQTT_Integration, Publish_With_Retain_Flag )
...
     TEST_ASSERT_FALSE( receivedSubAck );
-    TEST_ASSERT_EQUAL( MQTTSuccess,
-                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
-    TEST_ASSERT_TRUE( receivedSubAck );
+    waitForPacket( &receivedSubAck, MQTTSuccess );
 
     /* Make sure that the library invoked the event callback with the incoming PUBLISH from
      * the broker containing the "retained" flag set. */
-    TEST_ASSERT_TRUE( receivedRetainedMessage );
+    waitForPacket( &receivedRetainedMessage, MQTTSuccess );
```
muneebahmed10, line 1712: Is this the only instance where the flag we're waiting on isn't guaranteed to be `false` before calling the function? Otherwise, I see that `TEST_ASSERT_FALSE` is called before every other invocation of `waitForPacket`, so could be moved into it.

This isn't as pretty, but it would save one line for every other call of `waitForPacket`:
```c
/* If the retained message was not received in the last wait, wait again. */
if( !receivedRetainedMessage )
{
    waitForPacket( &receivedRetainedMessge, MQTTSuccess );
}
```

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -442,6 +454,18 @@ static void startPersistentSession();
...
+/**
+ * @brief Helper function to test for a specified incoming packet from the
+ * broker. This helper function allows us to wait for the specified incoming
+ * packet without unnecessarily blocking on MQTT_ProcessLoop() for beyond its
+ * arrival.
+ *
+ * @param[in] pFlag Packet flag to test for true.
+ * @param[in] expectedStatus Status code expected to be returned from MQTT_Process
+ */
+static inline void waitForPacket( bool * pFlag,
```
leegeth, line 466: why inline?

tests/integration_test/core_mqtt_system_test.c
sarenameas: sure thanks

tests/integration_test/core_mqtt_system_test.c
sarenameas: was experimenting left on accident will remove thanks.

tests/integration_test/core_mqtt_system_test.c
sarenameas: No that makes sense. I think it looks fine.

leegeth (dismissed):

muneebahmed10 (dismissed):

muneebahmed10 (dismissed):

yourslab (approved):

yourslab (dismissed):
How come we don't apply the same for lines 1049, 1251, 1613? There may be reasons that I'm missing.

yourslab:
Wow you can approve twice lolololol

sarenameas:
> How come we don't apply the same for lines 1049, 1251, 1613? There may be reasons that I'm missing.

1049 checks for a flag to remain false.
1251 and 1613 has not shown any intermittent failures in receiving the next publish packet expected. There are also no flags to indicate a publish was received. There may be some flaws with these test that can be analyzed in a separate scope.

sarenameas:
> How come we don't apply the same for lines 1049, 1251, 1613? There may be reasons that I'm missing.

Will handle incoming publish checking separately.

muneebahmed10 (dismissed):

yourslab (dismissed):

leegeth (dismissed):

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -120,128 +120,134 @@
...
+#define AWS_IOT_CORE_REPUBLISH_INTERVAL_MS         ( 30000U )
 
 /**
  * @brief Timeout for MQTT_ProcessLoop() function in milliseconds.
  * The timeout value is appropriately chosen for receiving an incoming
  * PUBLISH message and ack responses for QoS 1 and QoS 2 communications
  * with the broker.
  */
-#define MQTT_PROCESS_LOOP_TIMEOUT_MS          ( 700U )
+#define MQTT_PROCESS_LOOP_TIMEOUT_MS               ( 700U )
```
gkwicker, line 229: The comment on lines 225-227 doesn't seem appropriate now.

gkwicker (changes requested):
See comment

tests/integration_test/core_mqtt_system_test.c
sarenameas: Delete thanks.

yourslab (approved):

leegeth (approved):

gkwicker (approved):

PR #2755, opened 2020-11-17
coreMQTT_Integration test: Implement waiting on a publish packet, even one that is duplicate.

sarenameas:
This PR stabilizes the tests that wait on a publish packet from the broker. 
Publish packets, just like ACK packets, could take a long time to be received over the network from the broker.

A duplicate publish from the broker will not invoke the event callback from MQTT ProcessLoop. If the event callback is not invoked the test source file scoped boolean flags indicating that a packet was received cannot be set. 
This change updates waiting on a packet to input a boolean expression as a flag instead of a bool pointer. This helps us wait for a change in the incoming publish record in addition to a boolean global.

This PR ALSO clears retained messages in the broker, for a clean start to tests. AWS IoT does not support retained messages, so it is excluded. 

Description
-----------

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -1353,9 +1353,7 @@ void Connect_LWT()
 
     /* Establish a second TCP connection with the server endpoint, then
      * a TLS session. The server info and credentials can be reused. */
-    TEST_ASSERT_EQUAL( TRANSPORT_SOCKET_STATUS_SUCCESS, SecureSocketsTransport_Connect( &secondNetworkContext,
-                                                                                        &serverInfo,
-                                                                                        &socketsConfig ) );
+    TEST_ASSERT_TRUE( connectToServerWithBackoffRetries( &secondNetworkContext ) );
 
     /* Establish MQTT session on top of the TCP+TLS connection. */
     useLWTClientIdentifier = true;
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -218,7 +218,7 @@
  * @brief Time interval in seconds at which an MQTT PINGREQ need to be sent to
  * broker.
  */
-#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS           ( 30U )
+#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS           ( 60U )
 
 /**
  * @brief The number of milliseconds to wait for AWS IoT Core Message Broker
@@ -235,7 +235,7 @@
  * @brief The maximum number of times to call MQTT_ProcessLoop() when polling
  * for a specific packet from the broker.
  */
-#define MQTT_PROCESS_LOOP_PACKET_WAIT_COUNT_MAX    ( 60U )
+#define MQTT_PROCESS_LOOP_PACKET_WAIT_COUNT_MAX    ( 30U )
 
 /**
  * @brief The MQTT message published in this example.
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -218,7 +218,7 @@
  * @brief Time interval in seconds at which an MQTT PINGREQ need to be sent to
  * broker.
  */
-#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS           ( 60U )
+#define MQTT_KEEP_ALIVE_INTERVAL_SECONDS           ( 30U )
 
 /**
  * @brief The number of milliseconds to wait for AWS IoT Core Message Broker
--- a/tests/integration_test/core_mqtt_system_test.c
+++ b/tests/integration_test/core_mqtt_system_test.c
@@ -820,6 +820,7 @@ static MQTTStatus_t subscribeToTopic( MQTTContext_t * pContext,
     MQTTSubscribeInfo_t pSubscriptionList[ 1 ];
 
     assert( pContext != NULL );
+    assert( pTopic != NULL );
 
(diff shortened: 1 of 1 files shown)


tests/integration_test/core_mqtt_system_test.c
```diff
@@ -1606,11 +1742,10 @@ void Restore_Session_Duplicate_Incoming_Publish_Qos1()
 
     /* Process the duplicate incoming QoS 1 PUBLISH that will be sent by the broker
      * to re-attempt the PUBLISH operation. */
-    TEST_ASSERT_EQUAL( MQTTSuccess,
-                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
-
+    TEST_ASSERT_NOT_EQUAL( MQTT_PACKET_ID_INVALID, context.incomingPublishRecords[ 0 ].packetId );
     /* Make sure that the library cleared the record for the incoming QoS 1 PUBLISH packet. */
-    TEST_ASSERT_EQUAL( MQTT_PACKET_ID_INVALID, context.incomingPublishRecords[ 0 ].packetId );
+    WAIT_FOR_PACKET( ( MQTT_PACKET_ID_INVALID == context.incomingPublishRecords[ 0 ].packetId ),
```
sarenameas, line 1747: The function-like macro helps to wait on this duplicate publish condition.

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -78,6 +74,16 @@
 #endif
 #include "logging_stack.h"
 
+/* The logging configuration macros are defined above to ensure they are not
+ * superceded by definitions in the following header files. */
+
+/* Include OpenSSL implementation of transport interface. */
```
muneebahmed10, line 80: OpenSSL?

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -419,6 +459,64 @@ static void eventCallback( MQTTContext_t * pContext,
...
+ *
+ * @return MQTTSuccess if successful, error code otherwise. See #MQTT_Publish
+ * for possible return codes.
+ */
+static MQTTStatus_t publishToTopic( MQTTContext_t * pContext,
+                                    const char * pTopic,
+                                    const char * pPayload,
+                                    bool setRetainFlag,
+                                    bool isDuplicate,
+                                    MQTTQoS_t qos,
```
muneebahmed10, line 507: These are all of the fields for an `MQTTPublishInfo_t`, except for the sizes of the strings. At this point, would it just be easier to pass in the struct? We could even have a default initializer macro since most of these fields don't change.

tests/integration_test/core_mqtt_system_test.c
sarenameas: Will update that

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -246,6 +252,34 @@
...
+ * incoming packet without unnecessarily blocking on MQTT_ProcessLoop() for
+ * beyond its arrival.
+ *
+ * By using a macro instead of a function, the line number in the test case will
+ * be printed in a failure.
+ *
+ * @param[in] flag Packet flag to test for true.
+ * @param[in] expectedStatus Status code expected to be returned from MQTT_Process
+ */
+#define WAIT_FOR_PACKET( flag, expectedStatus )                                         \
```
yourslab, line 267: What advantage does this have over the function? Seems like the input-parameter types are always expected to be (bool, MQTTStatus_t).

tests/integration_test/core_mqtt_system_test.c
sarenameas: sure

sarenameas: I don't think one is easier over the other lol

tests/integration_test/core_mqtt_system_test.c
```diff
@@ -781,6 +898,26 @@ static MQTTStatus_t publishToTopic( MQTTContext_t * pContext,
                          packetId );
 }
 
+static void clearRetainedMessage( MQTTContext_t * pContext,
+                                  const char * pTopic )
+{
+    assert( pContext != NULL );
```
yourslab, line 904: Can also assert pTopic != NULL?

sarenameas: Sure, but since the other helpers i hadn't here as well.

muneebahmed10 (approved):

tests/integration_test/core_mqtt_system_test.c
cobusve: I would like to understand the reasoning for checking the one parameter here but not the other. Saying this was done incorrectly elsewhere is not sufficient reason to dismiss this question, if it is wrong elsewhere that also needs to be fixed.

That said it seems strange to check parameters on static functions, it is often better to ensure we always pass valid date into these and assert earlier (before calling the function something else must have failed - shift left here)

cobusve (changes requested):
We need to resolve the parameter checks as per my comment please

tests/integration_test/core_mqtt_system_test.c
sarenameas: This is fixed in https://github.com/aws/amazon-freertos/pull/2755/commits/e349b75ed0a722ed8a367695aa0550e040f2312a

sarenameas: This is simply a defensive developer check on parameters that should never be NULL.

cobusve (approved):

PR #2763, opened 2020-11-18
Add coreHTTP Integration Tests

sukhmanm:
Adding integration tests for the coreHTTP library:
- Tests simple GET, HEAD, POST, and PUT requests, as well as a chunked streaming response and a line-feed only response.


tests/integration_test/core_http_system_test.c
```diff
@@ -0,0 +1,629 @@
...
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+/* Include header for root CA certificates. */
+#include "iot_default_root_certificates.h"
+
+/* Include paths for public enums, structures, and macros. */
+#include "core_http_client.h"
```
sarenameas, line 47: We need to move this to after `#include "logging_stack.h" otherwise the logging definitions in the config will supercede the ones in this test.

tests/integration_test/core_http_system_test.c
```diff
@@ -0,0 +1,629 @@
...
+
+/* Include Secure Sockets implementation of transport interface. */
+#include "transport_secure_sockets.h"
```
sarenameas, line 50: We need to move these headers to after `#include "logging_stack.h" otherwise the logging definitions in the config will supercede the ones in this test.

sarenameas (approved):

n9wxu (approved):

PR #2767, opened 2020-11-19
Submodule FreeRTOS/backoffAlgorithm and update demos and tests for new API

aggarw13:
Replace the `retry_utils` library folder with submodule to the refactored library, now hosted, in [FreeRTOS/backoffAlgorithm](https://github.com/FreeRTOS/backoffAlgorithm) 
Update the demos and integration tests to use the refactored library API. 

**Note**: For the random number generator function provided to the refactored `retry_utils` library, the **PKCS11** module has been used to so that if the board platform supports a **True Random Number Generator**, the back-off calculation can take advantage of that.


aggarw13:
All files under `libraries/abstractions/retry_utils` have been copied from the C-SDK repository. This location will eventually be replaced with a submodule to a new repository that will host the retry utility library.

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
```diff
@@ -233,13 +282,77 @@ static uint32_t prvGetTimeMs( void );
...
+    BaseType_t xStatus = pdFAIL;
+
+    CK_RV xResult = CKR_OK;
+    CK_FUNCTION_LIST_PTR pxFunctionList = NULL;
+    CK_SESSION_HANDLE xSession = CK_INVALID_HANDLE;
+
+    /* Get list of functions supported by the PKCS11 port. */
+    xResult = C_GetFunctionList( &pxFunctionList );
+
+    if( ( xResult != CKR_OK ) || ( pxFunctionList != NULL ) )
```
lundinc2, line 308: Should this be == NULL?

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
```diff
@@ -233,13 +282,77 @@ static uint32_t prvGetTimeMs( void );
...
+    {
+        LogError( ( "Failed to seed random number generator. "
+                    "PCKS11 API, C_GetFunctionList, failed." ) );
+        xStatus = pdFAIL;
+    }
+
+    if( xStatus == pdPASS )
+    {
+        /* Initialize PKCS11 module and create a new session. */
+        xResult = xInitializePkcs11Session( &xSession );
```
lundinc2, line 318: May want to clean up by calling `C_CloseSession`

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -447,7 +502,11 @@ int RunCoreMqttMutualAuthDemo( bool awsIotMqttMode,
      */
     ulGlobalEntryTimeMs = prvGetTimeMs();
 
-    for( ; ulDemoRunCount < democonfigMQTT_MAX_DEMO_COUNT; ulDemoRunCount++ )
+    /* Seed randon number generator for use with retry utils library for
+     * backoff calculation during failed connection retry attempts. */
+    xDemoStatus = prvSeedRandomNumberGenerator();
+
+    for( ; ( ( xDemoStatus == pdPASS ) && ( ulDemoRunCount < democonfigMQTT_MAX_DEMO_COUNT ) ); ulDemoRunCount++ )
```
leegeth, line 509: ( xDemoStatus == pdPASS ) in this condition will fail the retries we are doing. Please move it out of this for loop condition.

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
```diff
@@ -233,13 +282,86 @@ static uint32_t prvGetTimeMs( void );
...
+{
+    const uint32_t ulMultiplier = 0x015a4e35UL, ulIncrement = 1UL;
+
+    ulNextRand = ( ulMultiplier * ulNextRand ) + ulIncrement;
+    return( ( int32_t ) ( ulNextRand >> 16UL ) & 0x7fffUL );
+}
+
+/*-----------------------------------------------------------*/
+
+static BaseType_t prvSeedRandomNumberGenerator()
```
leegeth, line 297: Where is this function invoked?

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -917,14 +973,87 @@ static MQTTStatus_t prvResumeSession( bool xSessionPresent )
...
+{
+    const uint32_t ulMultiplier = 0x015a4e35UL, ulIncrement = 1UL;
+
+    ulNextRand = ( ulMultiplier * ulNextRand ) + ulIncrement;
+    return( ( int32_t ) ( ulNextRand >> 16UL ) & 0x7fffUL );
+}
+
+/*-----------------------------------------------------------*/
+
+static BaseType_t prvSeedRandomNumberGenerator()
```
leegeth, line 986: I see that we are having copies of the same code in a few places. Is it possible to avoid this redundancy?

aggarw13: It is a valid concern, I think that the utility wrapper of generating a random number with PKCS11 can be part of the `corePKCS11` repository so that both the `amazon-freertos` and `FreeRTOS/FreeRTOS` can utilize it. 
What are your thoughts on it @lundinc2 ?

For now, that change would be beyond the scope of this PR.

aggarw13: I have commonized the logic of using PKCS11 functions into a wrapper function, `xPkcs11GenerateRandomNumber`, and created a separate library named `AFR::pkcs11_helpers` that the demos link against.

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
aggarw13: Thanks for catching it. I have added a call-site for this function in this file in the `prvConnectToServerWithBackoffRetries` function.

demos/coreMQTT/mqtt_demo_mutual_auth.c
aggarw13: Have introduced a new boolean variable that represents only whether the random number generation was successful.

leegeth (dismissed):

leegeth (dismissed):

tests/integration_test/core_http_system_test.c
```diff
@@ -258,14 +308,55 @@ static int32_t transportSendStub( NetworkContext_t * pNetworkContext,
...
+
+static int32_t generateRandomNumber()
+{
+    const uint32_t multiplier = 0x015a4e35UL, increment = 1UL;
+
+    nextRand = ( multiplier * nextRand ) + increment;
+    return( ( int32_t ) ( nextRand >> 16UL ) & 0x7fffUL );
+}
+
+static void seedRandomNumberGenerator()
```
leegeth, line 320: Where is this seeding done in this test?

aggarw13: Good catch, have added the call in `TEST_SETUP`

leegeth (dismissed):

demos/common/http_demo_helpers/http_demo_utils.c
```diff
@@ -26,27 +26,129 @@
...
+ * @return pdPASS if seeding the randon number generator was successful;
+ * otherwise pdFAIL for failure.
+ */
+static BaseType_t prvSeedRandomNumberGenerator();
+
+/*-----------------------------------------------------------*/
+
+static int32_t prvGenerateRandomNumber()
+{
+    const uint32_t ulMultiplier = 0x015a4e35UL, ulIncrement = 1UL;
```
gkwicker, line 102: Why does this routine do anything other than call `xPkcs11GenerateRandomNumber()`?  The RNG code on lines 104-105 looks to have issues (see comment below).  If you just call the PKCS#11 routine you can delete that.

gkwicker (changes requested):
See comments

demos/common/http_demo_helpers/http_demo_utils.c
```diff
@@ -28,25 +28,127 @@
...
+ * from devices in connection retries is mitigated.
+ *
+ * @return pdPASS if seeding the randon number generator was successful;
+ * otherwise pdFAIL for failure.
+ */
+static BaseType_t prvSeedRandomNumberGenerator();
+
+/*-----------------------------------------------------------*/
+
+static int32_t prvGenerateRandomNumber()
```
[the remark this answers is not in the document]
aggarw13: The signed return type is used to allow the RNG function to return an error. For example, the PKCS11 `C_GenerateRandom` function can return an error.

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -917,14 +972,34 @@ static MQTTStatus_t prvResumeSession( bool xSessionPresent )
 
 /*-----------------------------------------------------------*/
 
+static int32_t prvGenerateRandomNumber()
```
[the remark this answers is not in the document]
aggarw13: >Why does this routine contain anything other than a call to xPkcs11GenerateRandomNumber()?

Creating a PKCS11 session to generate a random number is memory (and potentially performance consuming if using a HW based RNG). Thus, based on suggestion from @lundinc, we decided to only _seed_ the pseudo random number generator with PKCS11

demos/common/http_demo_helpers/http_demo_utils.c
aggarw13: Have answered the reason of not using PKCS11 in every iteration of random number generation in the reply [here](https://github.com/aws/amazon-freertos/pull/2767#discussion_r529068149)

demos/common/http_demo_helpers/http_demo_utils.c
```diff
@@ -28,25 +28,127 @@
...
+{
+    const uint32_t ulMultiplier = 0x015a4e35UL, ulIncrement = 1UL;
+
+    ulNextRand = ( ulMultiplier * ulNextRand ) + ulIncrement;
+    return( ( int32_t ) ( ulNextRand >> 16UL ) & 0x7fffUL );
```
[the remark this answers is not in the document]
aggarw13: This pseudo RNG logic has just been taken from the existing demos. Example -  https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Demo/coreMQTT_Windows_Simulator/Common/main.c#L207-L227

>  The right shift on line 105 reduces this to 16 bits. Why?
As the next backoff period takes a 16 bit unsigned value in the [backoffAlgorithm library API](https://github.com/FreeRTOS/backoffAlgorithm/blob/main/source/include/backoff_algorithm.h#L130-L147), we don't need the RNG to create a larger bit width value. (The reason for choosing 16 bit value is that the largest 16 bit value represents a ~64 seconds value for the largest backoff period)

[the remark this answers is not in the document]
aggarw13: However, we can avoid discarding the randomness and return a 31 bit unsigned value to take advantage of the randomness that is completely in line with the [RNG interface](https://github.com/FreeRTOS/backoffAlgorithm/blob/main/source/include/backoff_algorithm.h#L45-L57) exposed by the backoffAlgorithm library API

demos/common/http_demo_helpers/http_demo_utils.c
```diff
@@ -28,25 +28,127 @@
...
+#define CONNECTION_RETRY_BACKOFF_BASE_MS         ( 500U )
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief A global storage variable that is used by the pseudo random number generator.
+ * The random number generator is used for calculating exponential back-off
+ * with jitter for retry attempts of failed network operations with the broker.
+ */
+static uint32_t ulNextRand;
```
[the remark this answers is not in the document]
aggarw13: Have removed this global from all demos and tests

demos/common/http_demo_helpers/http_demo_utils.c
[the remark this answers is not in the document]
aggarw13: We discussed separately and decided to remove use of SW random number generators completely (in this PR) to suggest users to use HW random number generator in every calculation for retry logic

leegeth (dismissed):

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
```diff
@@ -205,26 +192,15 @@ static bool mqttSessionEstablished = false;
...
- * The PKCS11 module is used to seed the random number generator
- * as it allows access to a True Random Number Generator (TRNG) if the
- * vendor platform supports it.
- * It is recommended to use a device-specific unique entropy source
- * to seed the random number generator so that probability of collisions
- * from devices in connection retries is mitigated.
- *
- * @return pdPASS if seeding the randon number generator was successful;
- * otherwise pdFAIL for failure.
+ * @return pdPASS if generating the randon number was successful; otherwise pdFAIL
```
gkwicker, line 200: What does this `@return` statement refer to?  It doesn't match the implementation of `prvGenerateRandomNumber()`.  Also, spelling: __random__.

demos/coreMQTT/mqtt_demo_connection_sharing.c
```diff
@@ -409,26 +409,15 @@ typedef struct publishElement
...
- * to seed the random number generator so that probability of collisions
- * from devices in connection retries is mitigated.
+ * The PKCS11 module is used to generate the random number as it allows access
+ * to a True Random Number Generator (TRNG) if the vendor platform supports it.
+ * It is recommended to use a device-specific unique random number generator so
+ * that probability of collisions from devices in connection retries is mitigated.
  *
- * @return pdPASS if seeding the randon number generator was successful;
- * otherwise pdFAIL for failure.
+ * @return pdPASS if generating the randon number was successful; otherwise pdFAIL
```
gkwicker, line 417: Same comment as earlier, this comment doesn't match.

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -234,26 +234,15 @@
...
- * to seed the random number generator so that probability of collisions
- * from devices in connection retries is mitigated.
+ * The PKCS11 module is used to generate the random number as it allows access
+ * to a True Random Number Generator (TRNG) if the vendor platform supports it.
+ * It is recommended to use a device-specific unique random number generator so
+ * that probability of collisions from devices in connection retries is mitigated.
  *
- * @return pdPASS if seeding the randon number generator was successful;
- * otherwise pdFAIL for failure.
+ * @return pdPASS if generating the randon number was successful; otherwise pdFAIL
```
gkwicker, line 242: Same comment here.

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -505,11 +483,7 @@ int RunCoreMqttMutualAuthDemo( bool awsIotMqttMode,
      */
     ulGlobalEntryTimeMs = prvGetTimeMs();
 
-    /* Seed randon number generator for use with backoff algorithm library for backoff calculation
-     * during retry attempts of failed network operations with the broker. */
-    xRngSeedStatus = prvSeedRandomNumberGenerator();
-
-    for( ; ( ( xRngSeedStatus == pdPASS ) && ( ulDemoRunCount < democonfigMQTT_MAX_DEMO_COUNT ) ); ulDemoRunCount++ )
+    for( ; ( ulDemoRunCount < democonfigMQTT_MAX_DEMO_COUNT ); ulDemoRunCount++ )
```
gkwicker, line 486: This usage is a little out of the ordinary, so the reader is forced to scan backwards to find the last assignment of `ulDemoRunCount`.  Make it easier for them by just doing the normal thing here:

```c
for( ulDemoRunCount = 0 ; ...
```

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -670,19 +644,21 @@ int RunCoreMqttMutualAuthDemo( bool awsIotMqttMode,
...
-    return( ( int32_t ) ( ulNextRand & 0x7ffffffUL );
-}
+    uint32_t ulRandomNum;
 
-/*-----------------------------------------------------------*/
+    /* Use the PKCS11 module to generate a random number. */
+    if( xPkcs11GenerateRandomNumber( ( uint8_t * ) &ulRandomNum,
+                                     ( sizeof( ulRandomNum ) == pdPASS ) ) )
+    {
+        ulRandomNum = ( ulRandomNum & INT32_MAX );
```
gkwicker, line 653: This is probably okay, but it assumes that the upper and lower halves of the distribution are completely independent and it throws away a bit of randomness.  This is the reason that `xPkcs11GenerateRandomNumber()` returns a _status_ which is independent from the _value_ returned BTW.  Any time status and value are combined in a return value, you're forced to make tradeoffs like this.  Here, we're eliminating 2B worth of possible random values just so that we can have a single negative return value.

The best way to solve this class of problem is to preseve the range of values.  Does `prvGenerateRandomNumber()` even need to exist, or could this demo simply call `xPkcs11GenerateRandomNumber()`?  That would be simpler and it would preserve all bits of randomness.

gkwicker (changes requested):
See comments

demos/coreMQTT/mqtt_demo_mutual_auth.c
aggarw13: This wrapper function is required for the RNG interface type in `FreeRTOS/backoffAlgorithm` which is defined as `typedef int32_t ( * BackoffAlgorithm_RNG_t )();`

As the backoff algorithm library only generates a backoff value that is 16 bits wide, losing the MSB worth of randomness in a 32 bit wide integer is fine (imo) here as the library ultimately normalizes the random value down to the max jitter window which cannot be more than 16 bits wide..

demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
aggarw13: Good catch. Have fixed the comment and the typo.

demos/coreMQTT/mqtt_demo_mutual_auth.c
aggarw13: Updated for improved user readability

gkwicker (dismissed):

karkhaz:
Hi! You'll need to rebase onto latest master for the proofs to run. Thanks!

gkwicker (approved):

PR #2770, opened 2020-11-19
Shim from HTTP V1 library to coreHTTP (HTTPClient_Send)

yourslab:
Description
-----------

This updates the HTTP V1 deprecated library to call `HTTPClient_Send` from the coreHTTP LTS library in order to send HTTP requests. In addition, task pool is removed as a dependency in order to allow for statically allocated tasks. This is done by using a dispatch queue to hold HTTP requests that are ready to be sent to the server.  

Checklist:
----------

### DIFF
--- a/libraries/c_sdk/standard/https/src/iot_https_client.c
+++ b/libraries/c_sdk/standard/https/src/iot_https_client.c
@@ -2412,6 +2412,7 @@ IotHttpsReturnCode_t IotHttpsClient_Init( void )
         }
     #endif /* if IOT_HTTPS_DISPATCH_USE_STATIC_MEMORY != 1 */
     dispatchQueue = NULL;
+
     /* Delete the tasks that send requests from the dispatch queue. */
     for( dispatchTaskIndex = 0; dispatchTaskIndex < IOT_HTTPS_DISPATCH_TASK_COUNT; ++dispatchTaskIndex )
     {
@@ -2425,6 +2426,7 @@ IotHttpsReturnCode_t IotHttpsClient_Init( void )
             httpsDispatchTask[ dispatchTaskIndex ] = NULL;
         }
     }
+
     HTTPS_FUNCTION_CLEANUP_END();
 }
 
@@ -2586,6 +2588,7 @@ void IotHttpsClient_Cleanup( void )
         }
     #endif /* if IOT_HTTPS_DISPATCH_USE_STATIC_MEMORY != 1 */
     dispatchQueue = NULL;
+
     /* Delete the tasks that send requests from the dispatch queue. */
     for( dispatchTaskIndex = 0; dispatchTaskIndex < IOT_HTTPS_DISPATCH_TASK_COUNT; ++dispatchTaskIndex )
     {
--- a/libraries/c_sdk/standard/https/src/iot_https_client.c
+++ b/libraries/c_sdk/standard/https/src/iot_https_client.c
@@ -531,7 +531,9 @@ static IotHttpsReturnCode_t _flushHttpsNetworkData( _httpsConnection_t * pHttpsC
  * @param[in] pJob - Pointer to the HTTP request sending job.
  * @param[in] pUserContext - Pointer to an HTTP request, passed as an opaque context.
  */
-static void _sendHttpsRequest( _httpsRequest_t * pUserContext );
+static void _sendHttpsRequest( IotTaskPool_t pTaskPool,
+                               IotTaskPoolJob_t pJob,
+                               void * pUserContext );
 
 /**
  * @brief Receive the HTTPS body specific to an asynchronous type of response.
@@ -2176,16 +2178,22 @@ static IotHttpsReturnCode_t _sendHttpsHeadersAndBody( _httpsConnection_t * pHttp
 
 /*-----------------------------------------------------------*/
 
-static void _sendHttpsRequest( _httpsRequest_t * pHttpsRequest )
+static void _sendHttpsRequest( IotTaskPool_t pTaskPool,
+                               IotTaskPoolJob_t pJob,
+                               void * pUserContext )
 {
     HTTPS_FUNCTION_ENTRY( IOT_HTTPS_OK );
 
+    _httpsRequest_t * pHttpsRequest = ( _httpsRequest_t * ) ( pUserContext );
     _httpsConnection_t * pHttpsConnection = pHttpsRequest->pHttpsConnection;
     _httpsResponse_t * pHttpsResponse = pHttpsRequest->pHttpsResponse;
     IotHttpsReturnCode_t disconnectStatus = IOT_HTTPS_OK;
     IotHttpsReturnCode_t scheduleStatus = IOT_HTTPS_OK;
     IotLink_t * pQItem = NULL;
     _httpsRequest_t * pNextHttpsRequest = NULL;
+
+    ( void ) pTaskPool;
+    ( void ) pJob;
 
     IotLogDebug( "Task with request ID: %p started.", pHttpsRequest );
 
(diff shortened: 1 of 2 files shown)


libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2622,7 +2552,16 @@ static IotHttpsReturnCode_t _shimConvertStatus( HTTPStatus_t coreHttpStatus )
...
+    uint8_t dispatchTaskIndex = 0;
+
+    /* Delete the tasks that send requests from the dispatch queue. */
+    for( dispatchTaskIndex = 0; dispatchTaskIndex < IOT_HTTPS_DISPATCH_TASK_COUNT; ++dispatchTaskIndex )
+    {
+        if( httpsDispatchTask[ dispatchTaskIndex ] != NULL )
+        {
+            vTaskDelete( httpsDispatchTask[ dispatchTaskIndex ] );
+        }
+    }
```
sarenameas, line 2564: Do we also need to add vQueueDelete() for the dispatchQueue?

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2464,6 +2344,56 @@ IotHttpsReturnCode_t IotHttpsClient_Init( void )
...
+                                                                        dispatchTaskStack,
+                                                                        &dispatchTaskBuffer );
+
+            if( httpsDispatchTask[ dispatchTaskIndex ] == NULL )
+            {
+                IotLogError( "Failed to allocate static memory for request task.", status );
+                HTTPS_SET_AND_GOTO_CLEANUP( IOT_HTTPS_INTERNAL_ERROR );
+            }
+        #else /* ifdef IOT_HTTPS_DISPATCH_USE_STATIC_MEMORY */
+            taskCreationResult = xTaskCreate( _dispatchTaskRoutine,
```
sarenameas, line 2382: How will you clean up these non-static tasks when the HTTP library is done?

libraries/c_sdk/standard/https/src/private/iot_https_internal.h
```diff
@@ -184,13 +184,43 @@
 #ifndef IOT_HTTPS_MAX_ALPN_PROTOCOLS_LENGTH
     #define IOT_HTTPS_MAX_ALPN_PROTOCOLS_LENGTH    ( 255 ) /* The maximum alpn protocols length is chosen arbitrarily. */
 #endif
-/* Provide the User-Agent Value definition from coreHTTP. */
+#ifndef IOT_HTTPS_QUEUE_RECV_TICKS
+    #define IOT_HTTPS_QUEUE_RECV_TICKS             pdMS_TO_TICKS( 200U ) /* The ticks to wait for a #xQueueReceive to complete. */
+#endif
+#ifndef IOT_HTTPS_QUEUE_SEND_TICKS
+    #define IOT_HTTPS_QUEUE_SEND_TICKS             portMAX_DELAY /* The ticks to wait for a #xQueueSendToBack to complete. */
```
sarenameas, line 191: I dont think that this should be the max by default because I don't think the orignal function IotTaskPool_Schedule() blocked at all.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2464,6 +2344,56 @@ IotHttpsReturnCode_t IotHttpsClient_Init( void )
...
+            {
+                IotLogError( "Failed to allocate dynamic memory for request task.", status );
+                HTTPS_SET_AND_GOTO_CLEANUP( IOT_HTTPS_INTERNAL_ERROR );
+            }
+        #endif /* ifdef IOT_HTTPS_DISPATCH_USE_STATIC_MEMORY */
+    }
+
 /* This code prints debugging information and is, therefore, compiled only when
  * log level is set to IOT_LOG_DEBUG. */
     #if ( LIBRARY_LOG_LEVEL == IOT_LOG_DEBUG )
```
sarenameas, line 2399: HTTPS_FUNCTION_EXIT_NO_CLEANUP(); might need to change to deleting the queue if it is created and deleting any non null tasks created, upon error.

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -2324,36 +2189,48 @@ static void _sendHttpsRequest( IotTaskPool_t pTaskPool,
...
-    if( taskPoolStatus != IOT_TASKPOOL_SUCCESS )
+    /* Occurs when the queue remains full for #IOT_HTTPS_QUEUE_SEND_TICKS. */
+    if( queueStatus != pdTRUE )
     {
-        IotLogError( "Error creating a taskpool job for request servicing. Error code: %d", taskPoolStatus );
+        IotLogError( "Failed to add request to full dispatch queue. Error code: %d", ( int ) queueStatus );
         HTTPS_SET_AND_GOTO_CLEANUP( IOT_HTTPS_INTERNAL_ERROR );
     }
 
-    taskPoolStatus = IotTaskPool_Schedule( IOT_SYSTEM_TASKPOOL, pHttpsConnection->taskPoolJob, 0 );
```
sarenameas, line 2349: The taskpool scheduling function never had to block because each job itself was a linked list node. I am a little bit concerned about the behavior change here.

yourslab: Completely agree. The reason why I went with a FreeRTOS queue here rather than the original linear container was for the efficiencies it provides:

> When a task attempts to read from an empty queue the task will be placed into the Blocked state (so it is not consuming any CPU time and other tasks can run) until either data becomes available on the queue, or the block time expires.
> 
> When a task attempts to write to a full queue the task will be placed into the Blocked state (so it is not consuming any CPU time and other tasks can run) until either space becomes available in the queue, or the block time expires.
> 
> If more than one task block on the same queue then the task with the highest priority will be the task that is unblocked first.

To emulate the original behavior, it would be best to set the timeout to 0 and return `IOT_HTTPS_INTERNAL_ERROR` when the queue is full. The original task pool returned the same when it ran out of memory to create a task. Slightly differing in behavior, however, I would argue this implementation is better because we can log queue size rather than having the customer estimate how many tasks from the pool were allocated for HTTP before running out of memory.

libraries/c_sdk/standard/https/src/private/iot_https_internal.h
yourslab: Thanks for catching. I actually switched these up. It is the RECV_TICKS that should block indefinitely (as we do not want to allocate CPU time when there are no requests to process), and I will be setting this to SEND_TICKS to 0 to emulate the original non-blocking behavior.

libraries/c_sdk/standard/https/src/iot_https_client.c
yourslab: > NOTE: The idle task is responsible for freeing the RTOS kernel allocated memory from tasks that have been deleted. It is therefore important that the idle task is not starved of microcontroller processing time if your application makes any calls to vTaskDelete (). Memory allocated by the task code is not automatically freed, and should be freed before the task is deleted.

From my understanding of this, once we call `xTaskDelete`, then the idle task will be responsible for clean up for a non-static task.

libraries/c_sdk/standard/https/src/iot_https_client.c
yourslab: Thanks, missed that.

libraries/c_sdk/standard/https/src/iot_https_client.c
sarenameas: Excuse me I missed the httpsDispatchTask[ dispatchTaskIndex ] that stores the task in the parameter. Thanks.

sarenameas (dismissed):

sarenameas (approved):

libraries/c_sdk/standard/https/src/iot_https_client.c
```diff
@@ -1068,6 +1022,57 @@ static IotHttpsReturnCode_t _receiveHttpsBodySync( _httpsResponse_t * pHttpsResp
...
+ */
+struct NetworkContext
+{
+    void * pNetworkConnection;                       /**< @brief The network connection used for sending packets on the network. */
+    const IotNetworkInterface_t * pNetworkInterface; /**< @brief The network interface used to send packets on the network using the above network connection. */
+};
+
+/*-----------------------------------------------------------*/
+
+static int32_t transportRecv( NetworkContext_t * pNetworkContext,
```
gkwicker, line 1037: I might add a comment here mentioning that this function is a dummy because the legacy library does not use it.  A reader may have missed the earlier comment.

gkwicker (approved):

libraries/c_sdk/standard/https/src/iot_https_client.c
yourslab: Sure, I will add that comment as part of a separate PR. Come to think of it, there are more doc changes that I'd like to make based on the changes here. At least this way we can start testing in CI already. Thanks Gary!

PR #2798, opened 2020-11-25
Fix metadata issue of coreMQTT demo not getting enabled

aggarw13:
With PR#2767, the introduction of the internal `AFR::pkcs11_helpers` module,that is added as a dependency on coreMQTT (and other demos), the coreMQTT demos were not getting enabled by CMake build when enabling the `AFR::core_mqtt_demo_dependencies` module. This PR fixes the issue by adding the `AFR::pkcs11_helpers` module as a dependency for `AFR::core_mqtt_demo_dependencies` module so that the INTERNAL module can be enabled and thereby, allow the coreMQTT demo (and other demos) to be enabled

### DIFF
--- a/libraries/core_mqtt_demo_dependencies.cmake
+++ b/libraries/core_mqtt_demo_dependencies.cmake
@@ -83,6 +83,17 @@ afr_module_dependencies(
         AFR::pkcs11_helpers
 )
 
+# Add dependency on PKCS11 Helpers module, that is required
+# by the Secure Sockets based coreMQTT demo, ONLY if the board
+# supports the PKCS11 module.
+if(TARGET AFR::pkcs11_implementation::mcu_port)
+    afr_module_dependencies(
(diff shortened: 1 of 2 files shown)


pavanmr94 (dismissed):

libraries/core_mqtt_demo_dependencies.cmake
```diff
@@ -79,16 +79,28 @@ afr_module_dependencies(
     ${AFR_CURRENT_MODULE}
     PUBLIC
         AFR::core_mqtt
+        AFR::backoff_algorithm
+        AFR::pkcs11_helpers
```
aggarw13, line 83:
```suggestion
```

ravibhagavandas (approved):

leegeth (approved):

cobusve (approved):

PR #2803, opened 2020-11-28
Fix build failures in Pic32

leegeth:
Fix build failures in Pic32

Description
-----------
Fix build failures in Pic32.
Root Cause: The builds were failing as the quotes in preprocessor macro for mbedtls config file was trimmed and resulting in includes like #include aws_mbedtls_config.h missing the quotes.
Fix: Add an escape character to the preprocessor macro definition.

Checklist:
----------

### DIFF
--- a/projects/microchip/curiosity_pic32mzef/mplab/aws_demos/nbproject/configurations.xml
+++ b/projects/microchip/curiosity_pic32mzef/mplab/aws_demos/nbproject/configurations.xml
@@ -1118,7 +1118,7 @@
 				<property key="place-data-into-section" value="false"/>
 				<property key="post-instruction-scheduling" value="default"/>
 				<property key="pre-instruction-scheduling" value="default"/>
-				<property key="preprocessor-macros" value="MBEDTLS_CONFIG_FILE=\&quot;aws_mbedtls_config.h\&quot;;CONFIG_MEDTLS_USE_AFR_MEMORY;__free_rtos__"/>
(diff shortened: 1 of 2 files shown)


projects/microchip/curiosity_pic32mzef/mplab/aws_demos/nbproject/configurations.xml
```diff
@@ -1118,7 +1118,7 @@
 				<property key="place-data-into-section" value="false"/>
 				<property key="post-instruction-scheduling" value="default"/>
 				<property key="pre-instruction-scheduling" value="default"/>
-				<property key="preprocessor-macros" value="MBEDTLS_CONFIG_FILE=\&quot;aws_mbedtls_config.h\&quot;;CONFIG_MEDTLS_USE_AFR_MEMORY;__free_rtos__"/>
+				<property key="preprocessor-macros" value="MBEDTLS_CONFIG_FILE=\\\&quot;aws_mbedtls_config.h\\\&quot;;CONFIG_MEDTLS_USE_AFR_MEMORY;__free_rtos__"/>
```
yourslab, line 1121: I recall that single backslash is used for newer versions of MPLAB while older versions use triple backslash. Could something have changed in CI that caused it to revert back to triple backslash?

leegeth: Yes. That is the issue. The root cause is that some of the CI systems are using older versions of the MPLAB and some are using new. This needs to be addressed in  CI. This PR is a way to unblock if we decide to choose to stick with the version that requires 3 `\`s

leegeth:
Closing this PR as decision is made to update the MPLAB version to latest to all the CI host machines.

PR #2813, opened 2020-12-02
Remove visibility of Device Defender library for unsupported boards on FreeRTOS console

aggarw13:
As the Device Defender demo added in PR#2799 is only supported on limited number of boards, we do not want the Defender Library to be visible for unsupported boards in their libraries list on the custom configuration wizard of FreeRTOS console


aggarw13:
Closing the PR as we decided to show the Device Defender library for all boards (that support TCP/IP networking stack) on the FreeRTOS console.

PR #2857, opened 2020-12-10
Fix S3 multi threading demo failure on NXP

leegeth:
Fix S3 multi threading demo failure on NXP

Description
-----------
Root cause: The S3 multi threaded demo allocates memory for the queues needed in the demo. With the default size for the Queue elements, the demo fails to allocate memory.
Fix: Optimize the size of the Queue elements for http demo for NXP.

Checklist:
----------

### DIFF
--- a/vendors/nxp/boards/lpc54018iotmodule/aws_demos/config_files/http_demo_s3_download_multithreaded_config.h
+++ b/vendors/nxp/boards/lpc54018iotmodule/aws_demos/config_files/http_demo_s3_download_multithreaded_config.h
@@ -123,7 +123,7 @@
  * be less than democonfigUSER_BUFFER_LENGTH. We don't expect S3 to send more
  * than 1024 bytes of headers.
  */
-#define democonfigUSER_BUFFER_LENGTH                ( 1024 )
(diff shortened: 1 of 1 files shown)


vendors/nxp/boards/lpc54018iotmodule/aws_demos/config_files/http_demo_s3_download_multithreaded_config.h
```diff
@@ -123,7 +123,7 @@
  * be less than democonfigUSER_BUFFER_LENGTH. We don't expect S3 to send more
  * than 1024 bytes of headers.
```
aggarw13, line 124: If S3 cannot send more than 1024 bytes of header, shouldn;t the buffer be larger than that to leave space for the response body?

leegeth: Yea. Talked to @sukhmanm and updated the config

sukhmanm (approved):

aggarg (approved):

Issue #2906, opened 2020-12-22
All save network requests fail after sending a delete network request if there are no networks saved


dgranahan:
**Describe the bug**
If no networks are saved and a delete network request is issued to index 0, for example, the delete network request will succeed.

In https://github.com/aws/amazon-freertos/blob/master/libraries/c_sdk/standard/ble/src/services/wifi_provisioning/iot_ble_wifi_provisioning.c `_deleteNetworkTask` calls `_popNetwork` which calls `WIFI_NetworkDelete`.

The implementation of `WIFI_NetworkDelete` for ESP32 (https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c#L1071-L1127) will always return success even with no networks saved. This is because `xRegistry.usNumNetworks` will be zero so `usIndex < xRegistry.usNumNetworks` will always be false so `eWiFiSuccess` will be returned.

Jumping back to `_popNetwork` if `WIFI_NetworkDelete` returns success it will decrement `wifiProvisioning.numNetworks`. This will cause `numNetworks` which is an unsigned int to underflow and become `INT_MAX`.

With `numNetworks` now at `INT_MAX`, any attempt to add a new network will fail in `_addNetworkTask` because of the max saved networks check.

```
        if( wifiProvisioning.numNetworks < IOT_BLE_WIFI_PROVISIONING_MAX_SAVED_NETWORKS )
        {
            ret = _addNewNetwork( &wifiProvisioning.addNetworkRequest.network,
                                  wifiProvisioning.addNetworkRequest.connect );
        }
        else
        {
            IotLogError( "Failed to add a new network, max networks limit (%d) reached", IOT_BLE_WIFI_PROVISIONING_MAX_SAVED_NETWORKS );
            ret = eWiFiFailure;
        }
```

**System information**
Espressif ESP32

**Expected behavior**
Issuing a delete network request even when no networks are saved should not break the ability to connect to another network.

**To reproduce**

1. Clear all networks so no networks are saved.
2. Issue a delete network request with index 0 which will return success.
3. Issue a save network request with connect true which will fail.

muneebahmed10:
Hi @dgranahan,

Thank you for bringing this to our attention. `numNetworks` should not be decremented when nothing is deleted, so I've created #2909 to fix this issue, and I'll let you know once it's been merged. Please let us know if you have any further questions.

Thank you

muneebahmed10:
[...]

PR #2909, opened 2020-12-23
Fix WiFi network deletion for ESP32

muneebahmed10:
Description
-----------

#2906 raises the following issue: in https://github.com/aws/amazon-freertos/blob/d92499008954c68a28c5ba58f9d7798a8f787ec4/libraries/c_sdk/standard/ble/src/services/wifi_provisioning/iot_ble_wifi_provisioning.c#L1550, `_deleteNetworkTask` calls [`_popNetwork`](https://github.com/aws/amazon-freertos/blob/master/libraries/c_sdk/standard/ble/src/services/wifi_provisioning/iot_ble_wifi_provisioning.c#L1192) with a `NULL` network profile, so [`WIFI_NetworkGet`](https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c#L1020) is not called, and `ret` remains `eWiFiSuccess`. When [`WIFI_NetworkDelete`](https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c#L1071) is called, it is then possible, if no networks are saved, for `usIndex < xRegistry.usNumNetworks`, so [`WIFI_NetworkDelete`](https://github.com/aws/amazon-freertos/blob/master/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c#L1071) returns success without deleting anything. `_popNetwork` will then decrement `wifiProvisioning.numNetworks`, causing it to underflow to UINT_MAX.

This fixes the issue by always returning `eWiFiFailure` when `usIndex >= xRegistry.usNumNetworks`.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c
+++ b/vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c
@@ -1108,11 +1108,11 @@ WIFIReturnCode_t WIFI_NetworkDelete( uint16_t usIndex )
 			{
 				xRet = nvs_commit( xNvsHandle );
 			}
-		}
 
-		if( xRet == ESP_OK )
-		{
-			xWiFiRet = eWiFiSuccess;
+			if( xRet == ESP_OK )
+			{
+				xWiFiRet = eWiFiSuccess;
+			}
 		}
 
 		// Close


vendors/espressif/boards/esp32/ports/wifi/iot_wifi.c
```diff
@@ -1108,11 +1108,11 @@ WIFIReturnCode_t WIFI_NetworkDelete( uint16_t usIndex )
 			{
 				xRet = nvs_commit( xNvsHandle );
 			}
-		}
 
-		if( xRet == ESP_OK )
-		{
-			xWiFiRet = eWiFiSuccess;
+			if( xRet == ESP_OK )
```
dachalco, line 1112: Perhaps, replacing[ entry to this _success_ block](https://github.com/aws/amazon-freertos/pull/2909/files#diff-b2e9f40dcf41678b337ff2ac1fa8ae4c06535ea4296c33fe419c319b4e90c245R1092)
```c
if( xRet == ESP_OK && usIndex < xRegistry.usNumNetworks )
```
with the following instead:
```c
if( xRet == ESP_OK && (0 < usIndex && usIndex < xRegistry.usNumNetworks ))
```
The latter spares runtime of block [1094, 1110] and may be more straightforward failure path when `usIndex == 0`

dachalco (changes requested):

dachalco (approved):

cobusve (approved):

PR #2912, opened 2020-12-23
Fix Cypress Psoc64 OCW missing project files.

mingyue86010:
Description
-----------
Add a Cmake function to add the files directly to console/source_paths.txt without adding to a module.
This function should be only uesd when any additional files need to be included into OCW build combo but does not belong to any module.

Cypress added some sharable makefile proeject files to ${AFR_ROOT_DIR}/projects/cypress/make_support/, outside the aws_demos folder. Use the above function to add those files into build combo.

Checklist:
----------

- [ ] My code is Linted.


tools/cmake/afr_metadata.cmake
```diff
@@ -358,6 +364,15 @@ function(afr_write_metadata)
     )
 endfunction()
 
+# This function is used for any additional files that need
+# be included into the OCW build combo
```
aggarw13, line 368:
```suggestion
# be included in the ZIP downloaded from FreeRTOS console.
```

tools/cmake/afr_metadata.cmake
```diff
@@ -358,6 +364,15 @@ function(afr_write_metadata)
...
+function(afr_files_to_console_metadata)
```
aggarw13, line 369: If this function is called multiple times in the same build combination, wouldn't the variable be reset?
It may be better to update this function to support appending to the `AFR_FILES_TO_CONSOLE_METADATA` list if called multiple times.

tools/cmake/afr_metadata.cmake
```diff
@@ -358,6 +364,15 @@ function(afr_write_metadata)
     )
 endfunction()
 
+# This function is used for any additional files that need
+# be included in the ZIP downloaded from FreeRTOS console.
+function(afr_files_to_console_metadata)
+    set(prop_var AFR_FILES_TO_CONSOLE_METADATA)
+    set(${prop_var} "" CACHE INTERNAL "")
+    afr_cache_append(${prop_var} ${ARGN})
```
aggarw13, line 372: If this function is called multiple times in the same build combination, wouldn't the variable be overwritten?
It may be better to update this function to support appending to the `AFR_FILES_TO_CONSOLE_METADATA` list if called multiple times.
```suggestion
function(afr_files_to_console_metadata)
    set(prop_var AFR_FILES_TO_CONSOLE_METADATA)
    if( NOT DEFINED ${prop_var} )
       set(${prop_var} "" CACHE INTERNAL "")
    endif()
    afr_cache_append(${prop_var} ${ARGN})
```

mingyue86010: You are right~

aggarw13 (dismissed):

cobusve (dismissed):

cobusve (approved):

PR #2914, opened 2020-12-28
Secure sockets test group simplify

DanielYEHsieh:
## Description
1. [FreeRTOS adds cellular LTE-M interface library to support cellular IoT based applications](https://aws.amazon.com/about-aws/whats-new/2020/12/freertos-adds-cellular-lte-m-library-support-cellular-iot-based-applications/), using Cellular, the echo server is no longer in local network like using wi-Fi or Eth, this PR made some suggested changes for Secure Socket Test group making it more realistic for scenarios that echo servers are in a remote network.( https://quip-amazon.com/KNMVAxus311S/Secure-Sockets-Test-Plan-Changes-proposal)
2. These changes improves the coverage for some boards in existing CI, see below test results

## Testing
1. [IDT log ](https://s3.console.aws.amazon.com/s3/buckets/dhs-share/object/edit_public_read_access?region=us-east-1&showversions=false)testing Cellular module B96 + Nuvoton M487.
2. CI testing logs:
a. All passed for pc*,numaker_iot_m487_wifi*,esp*,infineon* boards
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/2100/
b. Improving coverage for ti*,st*,xilinx* boards (the coverage can improve more if we adjust the default values for ST baords)
(with changes)
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/2107/
(without changes)
https://amazon-freertos-ci.corp.amazon.com/job/im_v2_master/job/nightly_custom_job/2106/

### DIFF
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -127,8 +127,8 @@ static volatile EventGroupHandle_t xSyncEventGroup = NULL;
 /* Bit definitions used with the xSyncEventGroup event group to allow the
  * prvEchoClientTxTask() and prvEchoClientRxTask() tasks to synchronize before
  * commencing a new cycle with a different socket. */
-#define tcptestTX_TASK_BIT    ( 0x01 << 1 )
-#define tcptestRX_TASK_BIT    ( 0x01 << 2 )
+#define tcptestTX_TASK_BIT                                  ( 0x01 << 1 )
+#define tcptestRX_TASK_BIT                                  ( 0x01 << 2 )
 
 #ifndef Threadsafe_SameSocketDifferentTasksFrameSize
     #define Threadsafe_SameSocketDifferentTasksFrameSize    ( 2 * 1500 )
@@ -1543,6 +1543,7 @@ static void prvSOCKETS_NonBlocking_Test( Server_t xConn )
                  */
                 xNumBytesReceived += lNumBytes;
             }
+
             if( lNumBytes < 0 )
             {
                 break;
@@ -1551,6 +1552,7 @@ static void prvSOCKETS_NonBlocking_Test( Server_t xConn )
             xEndTime = xTaskGetTickCount();
         }
         while( xMessageLength > xNumBytesReceived );
+
         xResult = prvCheckTimeout( xStartTime, xEndTime, xTimeout );
         TEST_ASSERT_EQUAL_INT32_MESSAGE( pdPASS, xResult, "Receive timeout was outside of acceptable range" );
 
@@ -1940,7 +1942,6 @@ static void prvSOCKETS_SendRecv_VaryLength( Server_t xConn )
         /* Send each message length ulMaxLoopCount times. */
         for( ulI = 0; ulI < ulMaxLoopCount; ulI++ )
         {
-
             memset( pucTxBuffer, tcptestTX_BUFFER_FILLER, tcptestBUFFER_SIZE );
 
             prvCreateTxData( ( char * ) pucTxBuffer,
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -130,8 +130,8 @@ static volatile EventGroupHandle_t xSyncEventGroup = NULL;
 #define tcptestTX_TASK_BIT                                  ( 0x01 << 1 )
 #define tcptestRX_TASK_BIT                                  ( 0x01 << 2 )
 
-#ifndef Threadsafe_SameSocketDifferentTasksFrameSize
-    #define Threadsafe_SameSocketDifferentTasksFrameSize    ( 2 * 1500 )
+#ifndef tcptestTHREAD_SAFE_SAME_SOCKET_DIFFERENT_TASKS_FRAME_SIZE
+    #define tcptestTHREAD_SAFE_SAME_SOCKET_DIFFERENT_TASKS_FRAME_SIZE    tcptestTWICE_MAX_FRAME_SIZE
 #endif
 
 /* Array setup for 2x an Ethernet II data section */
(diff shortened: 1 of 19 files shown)


libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -127,9 +127,12 @@ static volatile EventGroupHandle_t xSyncEventGroup = NULL;
 /* Bit definitions used with the xSyncEventGroup event group to allow the
  * prvEchoClientTxTask() and prvEchoClientRxTask() tasks to synchronize before
  * commencing a new cycle with a different socket. */
-#define tcptestTX_TASK_BIT    ( 0x01 << 1 )
-#define tcptestRX_TASK_BIT    ( 0x01 << 2 )
+#define tcptestTX_TASK_BIT                                  ( 0x01 << 1 )
+#define tcptestRX_TASK_BIT                                  ( 0x01 << 2 )
 
+#ifndef Threadsafe_SameSocketDifferentTasksFrameSize
+    #define Threadsafe_SameSocketDifferentTasksFrameSize    ( 2 * 1500 )
```
yourslab, line 134: It looks like this can be defined from `tcptestTWICE_MAX_FRAME_SIZE`?
```
#define Threadsafe_SameSocketDifferentTasksFrameSize tcptestTWICE_MAX_FRAME_SIZE
```

yourslab: Also, the naming convention is a bit different. How come we are no longer prefixing with `tcptest`?

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -1544,11 +1544,17 @@ static void prvSOCKETS_NonBlocking_Test( Server_t xConn )
...
+            }
+
             xEndTime = xTaskGetTickCount();
         }
-        while( ( ( xEndTime - xStartTime ) < xWaitTime ) && ( xMessageLength > xNumBytesReceived ) );
+        while( xMessageLength > xNumBytesReceived );
 
-        TEST_ASSERT_EQUAL_INT32_MESSAGE( xMessageLength, xNumBytesReceived, "Data was not received \r\n" );
+        xResult = prvCheckTimeout( xStartTime, xEndTime, xTimeout );
+        TEST_ASSERT_EQUAL_INT32_MESSAGE( pdPASS, xResult, "Receive timeout was outside of acceptable range" );
```
yourslab, line 1557: I think using `TEST_ASSERT_EQUAL_MESSAGE` would be a better choice here because `xResult` is of type `BaseType_t`.

yourslab: Is this check no longer necessary?
```
TEST_ASSERT_EQUAL_INT32_MESSAGE( xMessageLength, xNumBytesReceived, "Data was not received \r\n" );
```

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -3053,7 +3059,7 @@ static void prvThreadSafeDifferentSocketsDifferentTasks( void * pvParameters )
                                           ipconfigTCP_MSS - xReceivedBytes,        /* The size of the buffer provided to receive the data. */
                                           0 );                                     /* No flags. */
 
-                if( xReturned <= 0 )
+                if( xReturned < 0 )
```
yourslab, line 3062: If `xReturned` is always 0, is it possible to be stuck in this loop forever? Should we also check for a timeout here the same way we do in the other tests?

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -423,7 +426,7 @@ typedef struct
 #define tcptestMAX_LOOPS_ECHO_TEST            tcptestMAX_ECHO_TEST_MODES
 
 #define tcptestECHO_TEST_LOW_PRIORITY         tskIDLE_PRIORITY
-#define tcptestECHO_TEST_HIGH_PRIORITY        ( configMAX_PRIORITIES - 1 )
+#define tcptestECHO_TEST_HIGH_PRIORITY        ( configMAX_PRIORITIES - 2 )
```
yourslab, line 429: What is the reason for changing this?

DanielYEHsieh: This comes from here:
https://quip-amazon.com/KNMVAxus311S/Secure-Sockets-Test-Plan-Changes-proposal#cNK9CAbPC6R

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
DanielYEHsieh: xReturned should not always get 0, because it will receive data from the echo server. Considering the received data could be send back from a remote server, which is not guaranteed latency, it should be expected that some SOCKETS_Recv will timeout and received xReturned = 0. But it should not get SOCKETS_Recv = 0 every time. We can add a check for prvCheckTimeout, but this test case seems not checking that oriented, but we can add it if we decide to.

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
yourslab: Also, everytime we call `prvCheckTimeout`, we end up calling:
```
TEST_ASSERT_EQUAL_INT32_MESSAGE( pdPASS, xResult, "Receive timeout was outside of acceptable range" );
```
In that case, would it be better to just remove this check inside `prvCheckTimeout`?

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
yourslab: It won't always return 0 but what will happen when the echo server stops returning any data from the payload? It is possible to create a rogue echo server that will cause an infinite loop in this test to occur. It is a corner case, but I think we have to account for that IMO.

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
DanielYEHsieh: Thanks for the suggestion, fixed in 6765b43

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
DanielYEHsieh: Thanks for the suggestion, fixed in ba515b4.
Starting a new CI tests will wait for the results and resolve this comment if it gets ideal result.

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
yourslab: We discussed in chat that a potential infinite loop for these tests specifically is probably not much of a concern.

gkwicker (approved):

yourslab (approved):

PR #2955, opened 2021-01-26
Add CLI input mode to test runner

aggarw13:
This PR adds a another solution ( separate from the configurable delay config added in PR #2950 ) for the problem of race condition between AWS IoT Device Tester (that runs on host machine) getting serial console access to device and the test runner executing tests on the device.

This PR adds support to the test runner to wait for accepting a "start" command from the serial console (based on FreeRTOS+CLI library) before beginning execution of tests. This can allow the AWS IoT Device Tester to ensure that the device is serially accessible before tests begin executing on the device.
This feature can be enabled with the `AWS_TEST_RUNNER_ENABLE_CLI_INPUT` config in the `aws_test_runner_config.h` file. 

Here is the log that shows interacting with the test runner when the feature is enable. Note that after the `start` command is provided on the serial prompt, the tests starts executing. 

```
Welcome to the Test Runner console.
[Type the "start" command to beging executing tests]
>help

help:
 Lists all the registered commands

start : Instructs test runner to start executing tests on the board.
[Press Enter to run the previous command]
>start
---------STARTING TESTS---------
TEST(Full_PKCS11_StartFinish, AFQP_StartFinish_FirstTest) PASS
TEST(Full_PKCS11_StartFinish, AFQP_GetFunctionList) PASS
```

**Note**: This feature uses FreeRTOS+CLI library along with Common I/O interface for UART communication. Thus, this feature is only supported by boards that provide ports for Common I/O interface.


libraries/freertos_plus/standard/freertos_plus_cli/source/FreeRTOS_CLI_Console.c
```diff
@@ -107,12 +102,14 @@ void FreeRTOS_CLIEnterConsoleLoop( xConsoleIO_t consoleIO,
     }
 }
 
-static void processInputBuffer( xConsoleIO_t consoleIO,
-                                int32_t inputSize,
-                                char * pCommandBuffer,
-                                size_t commandBufferLength,
-                                char * pOutputBuffer,
-                                size_t outpuBufferLength )
+void FreeRTOS_CLI_ProcessInputBuffer( xConsoleIO_t consoleIO,
```
ravibhagavandas, line 105: This duplicates some code in `FreeRTOS_CLI_ProcessInputBuffer` to handle the exiting the CLI loop. Can we keep code same and instead add a new command such as `exit` for the loop to exit gracefully.

tests/common/aws_test_runner.c
```diff
@@ -278,13 +388,20 @@ void TEST_RUNNER_RunTests_task( void * pvParameters )
...
+        /* Add sufficient delay before starting tests on device to allow
+         * device to be available as serial port connection to the host machine
+         * OS.
+         * The serial console is used by host machine to view device logs. */
+        vTaskDelay( pdMS_TO_TICKS( AWS_TEST_RUNNER_DELAY_MS ) );
 
-    UNITY_BEGIN();
+        UNITY_BEGIN();
+
+        RunTests();
```
ravibhagavandas, line 403: It seems the memory leak test wont start before the test begin
`testrunnerFULL_MEMORYLEAK_ENABLED `

tests/common/aws_test_runner.c
```diff
@@ -266,6 +275,107 @@ static void RunTests( void )
...
+        ( void ) xWriteBufferLen;
+        ( void ) pcCommandString;
+
+        /* Set flag to indicate reception of command from serial console to start */
+        /* executing tests. */
+        xReceivedCommand = pdTRUE;
+
+        UNITY_BEGIN();
+
+        RunTests();
```
ravibhagavandas, line 305: Needs to handle `UNITY_END()` ?

tests/common/aws_test_runner.c
```diff
@@ -266,6 +275,107 @@ static void RunTests( void )
...
+                                          size_t xWriteBufferLen,
+                                          const int8_t * pcCommandString )
+    {
+        ( void ) pcWriteBuffer;
+        ( void ) xWriteBufferLen;
+        ( void ) pcCommandString;
+
+        /* Set flag to indicate reception of command from serial console to start */
+        /* executing tests. */
+        xReceivedCommand = pdTRUE;
```
ravibhagavandas, line 301: An "OK" response back to the peer here will help in debugging a broken CLI connection to the actual test issues.

tests/common/aws_test_runner.c
aggarw13: `UNITY_END` is called after the console CLI exists.

tests/common/aws_test_runner.c
aggarw13: Added the response to acknowledge processing of command.

tests/common/aws_test_runner.c
aggarw13: Fixed by moving all the test setup, execution of tests and teardown logic in common function.

tests/common/aws_test_runner.c
aggarw13: Have moved all the test logic (of setup, executing tests and tear down) to common function, which is called from the handler now.

aggarw13:
`/bot run checks`

libraries/freertos_plus/standard/freertos_plus_cli/source/FreeRTOS_CLI_Console.c
aggarw13: We discussed offline and decided that we can add a `exit` command support to the `FreeRTOS_CLIEnterConsoleLoop` function to allow the console loop to be terminated in a separate PR.

ravibhagavandas (approved):

cobusve (approved):

tests/common/aws_test_runner.c
aggarw13: Per offline discussion, removed the response for now as the response would be visible only at the end of the test execution. We can consider adding the response string later if there are architectural changes of running the CLI console in a different task than the test runner.

PR #2969, opened 2021-02-04
Rename aws_application_version.h to iot_application_version.h

paulbartell:
Rename aws_application_version.h to iot_application_version.h

Description
-----------
Rename aws_application_version.h to iot_application_version.h since this file is not necessarily AWS related.

Checklist:
----------
- [x] My code is Linted.


aggarw13:
Seems like there is a submodule update for the FreeRTOS+TCP library in the change set. Is that an intended part of the change?

paulbartell:
@aggarw13 Yes. The submodule update is intentional, since one of the files in that repo references aws_application_version.h. 

It seems odd for there to be this type of dependency. Working that out with @AniruddhaKanhere in the PR over here: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/188

PR #2983, opened 2021-02-09
Retry subscribing to topic filter when SUBACK not received

yourslab:
Description
-----------

For Shadow, Defender, and Jobs demos, the function `SubscribeToTopic` would have returned successfully even though the SUBACK wasn't received. This is due to the fact that `MQTT_ProcessLoop` returns success even when there is no data to receive and does not wait to receive any packet type. This fixes it so as to add the logic from the MQTT Mutual Auth demo to retry the subscription whenever the SUBACK is not received within the timeout and also returns failure once all attempts are exhausted.

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2988, opened 2021-02-10
Fix comments and unused function in mqtt_demo_helpers.c

yourslab:
Description
-----------

Just removes an empty function that I forgot to remove in PR #2983. Also cleans up some docstrings that were copy and pasted from another set of macros.

Checklist:
----------

### DIFF
--- a/demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
+++ b/demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
@@ -579,6 +579,7 @@ BaseType_t SubscribeToTopic( MQTTContext_t * pxMqttContext,
         {
             LogError( ( "SUBACK never arrived for subscription attempt to topic %s.",
                         xTopicFilterContext.pcTopicFilter ) );
+            break;
         }
         else
         {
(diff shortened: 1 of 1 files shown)


demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
```diff
@@ -604,9 +600,9 @@ BaseType_t SubscribeToTopic( MQTTContext_t * pxMqttContext,
                 xBackoffStatus = prvBackoffForRetry( &xRetryParams );
             }
         }
-    } while( ( xFailedSubscribeToTopic == pdTRUE ) && ( xBackoffStatus == pdPASS ) );
+    } while( ( xSubscribeStatus == pdFAIL ) && ( xBackoffStatus == pdPASS ) );
```
aggarw13, line 603: Wouldn't this result in retrying subscription request when the response is never received from the server? Should we just break out of the loop if a response is not received from the broker?

yourslab: Yes, it will fail and end up retrying. It should just break out of the loop when `xSubscribeStatus == pdFAIL`.

aggarw13 (approved):

dan4thewin (approved):

PR #2992, opened 2021-02-15
Fix prvConnectToServerWithBackoffRetries to correctly return connection status

lundinc2:
Description
-----------
The function `prvConnectToServerWithBackoffRetries` was always returning `pdPASS` regardless of whether or not the connection was successful. This caused the demo to segfault as it ended up using an invalid socket to establish a MQTT session with the broker in case of a connection failure.

This PR updates  `prvConnectToServerWithBackoffRetries` to return `pdFAIL` in case of connection failure. This ensures that the demo does not attempt to establish a MQTT session with the broker in case of a connection failure.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/coreMQTT/mqtt_demo_mutual_auth.c
+++ b/demos/coreMQTT/mqtt_demo_mutual_auth.c
@@ -781,7 +781,7 @@ static BaseType_t prvConnectToServerWithBackoffRetries( NetworkContext_t * pxNet
 
     xStatus = ( xBackoffStatus == pdPASS ) ? pdPASS : pdFAIL;
 
-    return xStatus;
+    return ( xNetworkStatus == TRANSPORT_SOCKET_STATUS_SUCCESS ) ? pdPASS : pdFAIL;
 }
 /*-----------------------------------------------------------*/
 
--- a/demos/coreMQTT/mqtt_demo_mutual_auth.c
+++ b/demos/coreMQTT/mqtt_demo_mutual_auth.c
@@ -723,7 +723,6 @@ static BaseType_t prvConnectToServerWithBackoffRetries( NetworkContext_t * pxNet
     ServerInfo_t xServerInfo = { 0 };
 
     SocketsConfig_t xSocketsConfig = { 0 };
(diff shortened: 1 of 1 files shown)


demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -779,6 +779,8 @@ static BaseType_t prvConnectToServerWithBackoffRetries( NetworkContext_t * pxNet
         }
     } while( ( xNetworkStatus != TRANSPORT_SOCKET_STATUS_SUCCESS ) && ( xBackoffStatus == pdPASS ) );
 
+    xStatus = ( xBackoffStatus == pdPASS ) ? pdPASS : pdFAIL;
```
yourslab, line 782: Can we just return `xBackoffStatus` and remove `xStatus` altogether?

yourslab: Oh I think the issue is that `xBackoffStatus` is using `pdFALSE` instead of `pdFAIL`. Although, I suppose they do map to the same most value for the most part.

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -779,6 +779,8 @@ static BaseType_t prvConnectToServerWithBackoffRetries( NetworkContext_t * pxNet
...
+
     return xStatus;
```
yourslab, line 784: Although, I think we care more about `xNetworkStatus` instead
```suggestion
    return ( xNetworkStatus == TRANSPORT_SOCKET_STATUS_SUCCESS ) ? pdPASS : pdFAIL;
```

demos/coreMQTT/mqtt_demo_mutual_auth.c
lundinc2: xStatus removed 👍

yourslab (approved):

aggarg (approved):

PR #2994, opened 2021-02-16
Update corePKCS11 and PKCS #11 PAL interfaces with latest changes

lundinc2:
Update corePKCS11 and PKCS #11 PAL interfaces with latest changes

Description
-----------
The default implementation of PKCS11_PAL_DestroyObject has been removed, so each port needs to be updated with an implementation of PKCS11_PAL_DestroyObject

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/vendors/cypress/boards/CYW943907AEVAL1F/ports/pkcs11/core_pkcs11_pal.c
+++ b/vendors/cypress/boards/CYW943907AEVAL1F/ports/pkcs11/core_pkcs11_pal.c
@@ -405,11 +405,18 @@ CK_RV PKCS11_PAL_DestroyObject( CK_OBJECT_HANDLE xHandle )
 
     prvHandleToLabel( &pcLabel, xHandle );
 
-    xLabel.type = CKA_LABEL;
-    xLabel.pValue = pcLabel;
-    xLabel.ulValueLen = strlen( pcLabel );
+    if( pcLabel != NULL )
+    {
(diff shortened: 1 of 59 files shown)


mingyue86010 (approved):

vendors/cypress/boards/CYW943907AEVAL1F/ports/pkcs11/core_pkcs11_pal.c
```diff
@@ -357,3 +360,95 @@ void PKCS11_PAL_GetObjectValueCleanup( CK_BYTE_PTR pucData,
...
+    }   
+    else
+    {
+        xResult = CKR_OBJECT_HANDLE_INVALID;
+    }   
+
+    if( xResult == CKR_OK )
+    {
+        /* Some ports return a pointer to memory for which using memset directly won't work. */
+        pxZeroedData = pvPortMalloc( ulObjectLength );
```
dan4thewin, line 424: Why not use the stack for this?

lundinc2: Objects can get up to ~1200 bytes for eg. RSA private key. I'm afraid it would blow the stack, and generally these operations don't happen often, so may be some wasted memory there.

dan4thewin (approved):

PR #3006, opened 2021-02-23
Possible bug-fix for aws_demo_logging.c

htibosch:
Description
-----------
As I am mostly working on FreeRTOS+TCP these days, I rarely look at the PR's in the [amazon-freertos](https://github.com/aws/amazon-freertos) repo.

But as today I was repairing a possible bug in aws_demo_logging.c, I saw [Ming Yue's](@mingyue86010) recent [PR #2995](https://github.com/aws/amazon-freertos/pull/2995) for this module.

Ming Yue, the problem was not the uninitialised local variable `xLength2`, but the missing else clause of `if( xArgs != NULL )`:
~~~c
     if( xArgs != NULL )
     {
         xLength2 = vsnprintf( cPrintString + xLength,
                               dlMAX_PRINT_STRING_LENGTH - xLength,
                               pcFormat,
                               xArgs );
     }
+    else
+    {
+        xLength2 = snprintf( cPrintString + xLength,
+                             dlMAX_PRINT_STRING_LENGTH - xLength,
+                             "%s",
+                             pcFormat );
+    }
~~~
In earlier versions, the else clause was present.
Here is a normal use case in which 'xArgs' is set to NULL:
~~~
void vLoggingPrint( const char * pcFormat )
{
    prvLoggingPrintf( LOG_NONE, NULL, 0, pdFALSE, pcFormat, NULL );
}
~~~
Now my possible bug-fix: when checking the free space with `uxStreamBufferGetSpace()`, it must happen within the critical section, thus like this:
~~~
    SetThreadPriority( GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
    {
    
        /* How much space is in the buffer? */
        xLength2 = uxStreamBufferGetSpace( xLogStreamBuffer );
    
        /* There must be enough space to write both the string and the length of
         * the string. */
        if( xLength2 >= ( xLength + sizeof( xLength ) ) )
        {
            uxStreamBufferAdd( xLogStreamBuffer,
                               0,
                               ( const uint8_t * ) &( xLength ),
                               sizeof( xLength ) );
            uxStreamBufferAdd( xLogStreamBuffer,
                               0,
                               ( const uint8_t * ) cOutputString,
                               xLength );
        }
        else
        {
            /* Log line will be dropped, bad luck. */
        }
    }
    SetThreadPriority( GetCurrentThread(), iOriginalPriority );
~~~

Otherwise a race condition is possible in which a higher-priority task comes in between and fills the stream buffer. This may cause the logging to stop working properly.
Hein

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
@@ -119,6 +119,8 @@ static void prvCreatePrintSocket( void * pvParameter1,
  * A Windows thread will finally call printf() and fflush().
  */
 static void prvLoggingPrintf( uint8_t usLoggingLevel,
+                              const char * pcFile,
+                              size_t fileLineNo,
                               BaseType_t xFormatted,
                               const char * pcFormat,
                               va_list xArgs );
@@ -361,7 +363,7 @@ void vLoggingPrintfWithFileAndLine( const char * pcFile,
     va_list args;
 
     va_start( args, pcFormat );
-    prvLoggingPrintf( LOG_NONE, pcFile, fileLineNo, pcFormat, args );
+    prvLoggingPrintf( LOG_NONE, pcFile, fileLineNo, pdTRUE, pcFormat, args );
 
     va_end( args );
 }
@@ -597,11 +599,11 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
             SetThreadPriority( GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
             {
                 /* How much space is in the buffer? */
-                xLength2 = uxStreamBufferGetSpace( xLogStreamBuffer );
+                size_t uxSpace = uxStreamBufferGetSpace( xLogStreamBuffer );
 
                 /* There must be enough space to write both the string and the length of
                  * the string. */
-                if( xLength2 >= ( xLength + sizeof( xLength ) ) )
+                if( uxSpace >= ( xLength + sizeof( xLength ) ) )
                 {
                     uxStreamBufferAdd( xLogStreamBuffer,
                                        0,
--- a/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
@@ -379,7 +379,7 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
     char cOutputString[ dlMAX_PRINT_STRING_LENGTH ];
     char * pcSource, * pcTarget, * pcBegin;
     size_t xLength, rc;
-    size_t xLength2 = 0;
+    size_t xLength2;
     static BaseType_t xMessageNumber = 0;
     uint32_t ulIPAddress;
     const char * pcTaskName;
--- a/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
@@ -332,6 +332,8 @@ void vLoggingPrintfInfo( const char * pcFormat,
 
     va_start( args, pcFormat );
     prvLoggingPrintf( LOG_INFO, NULL, 0, pdTRUE, pcFormat, args );
+
+    va_end( args );
 }
 
 /*-----------------------------------------------------------*/
@@ -346,6 +348,7 @@ void vLoggingPrintfDebug( const char * pcFormat,
 
     va_end( args );
 }
+
 /*-----------------------------------------------------------*/
 
 void vLoggingPrintfWithFileAndLine( const char * pcFile,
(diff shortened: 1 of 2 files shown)


vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -577,31 +587,37 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
...
-                                   0,
-                                   ( const uint8_t * ) &( xLength ),
-                                   sizeof( xLength ) );
-                uxStreamBufferAdd( xLogStreamBuffer,
-                                   0,
-                                   ( const uint8_t * ) cOutputString,
-                                   xLength );
-                SetThreadPriority( GetCurrentThread(), iOriginalPriority );
+                /* How much space is in the buffer? */
+                xLength2 = uxStreamBufferGetSpace( xLogStreamBuffer );
```
AniruddhaKanhere, line 600: Might be better to use a different variable for this instead of `xLength2`. Something like `xStreamBufferSpace`.

vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -577,31 +587,37 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
...
-             * the string. */
-            if( xLength2 >= ( xLength + sizeof( xLength ) ) )
+            /* First write in the length of the data, then write in the data
+             * itself.  Raising the thread priority is used as a critical section
+             * as there are potentially multiple writers.  The stream buffer is
+             * only thread safe when there is a single writer (likewise for
+             * reading from the buffer). */
+            xCurrentTask = GetCurrentThread();
+            iOriginalPriority = GetThreadPriority( xCurrentTask );
+            SetThreadPriority( GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
```
AniruddhaKanhere, line 597: As you have said, why not use a FreeRTOS mutex here? 
As far as I know, all the tasks accessing this section will be FreeRTOS tasks which can be handled safely by a mutex. My concern is that raising the priority might not be enough to remove race conditions altogether.

AniruddhaKanhere (changes requested):

vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -478,6 +478,13 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
                                   pcFormat,
                                   xArgs );
         }
+        else
```
mingyue86010, line 481: @htibosch Your change of add the "else" looks correct. Thank you for pointing out the using scenario of 
```
void vLoggingPrint( const char * pcFormat )
{
    prvLoggingPrintf( LOG_NONE, NULL, 0, pdFALSE, pcFormat, NULL );
}
```
The previous version did not address the case when args = NULL and only print out a string.

vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
htibosch: A FreeRTOS mutex can only be used by a FreeRTOS task. If a Windows thread calls an API for a FreeRTOS mutex, the kernel will assume that the task identified by `pxCurrentTCB` is waiting for the mutex.
When a FreeRTOS task gets blocked in any of the Windows API's, it may happen that a second FreeRTOS task becomes the active one, causing a big chaos.
The safest option seems to me to play with the current Windows thread priority.

vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
htibosch: You're right, it looks as if I wanted to economise on the use of local variables! Now changed to `uxSpace`.

vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
htibosch: No problem.
This function already has a [long history](https://github.com/aws/amazon-freertos/pull/285#issuecomment-461647595) :-)

vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -376,7 +379,7 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
     char cOutputString[ dlMAX_PRINT_STRING_LENGTH ];
     char * pcSource, * pcTarget, * pcBegin;
     size_t xLength, rc;
-    size_t xLength2 = 0;
+    size_t xLength2;
```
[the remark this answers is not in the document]
htibosch: Well seen, thanks. The function header was also incorrect, I changed it now:
~~~c
 static void prvLoggingPrintf( uint8_t usLoggingLevel,
+                              const char * pcFile,
+                              size_t fileLineNo,
                               BaseType_t xFormatted,
                               const char * pcFormat,
                               va_list xArgs );
~~~

htibosch (commented):
Thanks for your quick reviews

lundinc2 (approved):

AniruddhaKanhere (approved):

n9wxu (approved):

PR #3017, opened 2021-02-27
Better locking for Windows logging

htibosch:
Description
-----------
When adding text to the logging stream buffer, a mutex is needed. But as both Windows threads, as well as FreeRTOS tasks may send logging, using the W32 mutex is not a good option.

The current method will increase the thread priority to `TIME_CRITICAL`. @RichardBarry commented that setting the thread priority might not become effective immediately.
I tested this, and **indeed** the stream buffers get corrupted when there are both FreeRTOS tasks and Windows threads sending frequent logging.

After a lot of testing, I found that the "interlocked" functions work best:
~~~c
void vBitOpsAcquire( LONG volatile *plValue )
{
    /* Return the value of bit-0 before setting it, as an atomic operation. */
    while( _interlockedbittestandset( plValue, 0U ) == 1 )
    {
        /* It would be good to yield here, but that would not
         * work for a Windows thread. */
    }
}

void vBitOpsRelease( LONG volatile *plValue )
{
    /* Return the value of bit-0 before clearing it, as an atomic operation. */
    BOOLEAN rc = _interlockedbittestandreset( plValue, 0U );

    /* If the bit was not set, this is a fatal error. */
    configASSERT( rc != 0 );
}
~~~

I tested this locking with 2 FreeRTOS tasks which have a IDLE priority. Also I had a Windows thread sending constant logging.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/microchip/boards/ecc608a_plus_winsim/aws_demos/application_code/aws_demo_logging.c
@@ -125,6 +125,13 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
                               const char * pcFormat,
                               va_list xArgs );
 
+/* A simple locking mechanism to protect access to the logging
+stream buffer. */
+static void prvBitOpsAcquire( LONG volatile *plValue );
+
+/* Release the lock. */
+static void prvBitOpsRelease( LONG volatile *plValue );
+
 /*-----------------------------------------------------------*/
 
 /* Windows event used to wake the Win32 thread which performs any logging that
@@ -230,6 +237,7 @@ void vLoggingInit( BaseType_t xLogToStdout,
                     NULL );
 
                 /* Use the cores that are not used by the FreeRTOS tasks. */
+                configASSERT( Win32Thread != NULL );
                 SetThreadAffinityMask( Win32Thread, ~0x01u );
                 SetThreadPriorityBoost( Win32Thread, TRUE );
                 SetThreadPriority( Win32Thread, THREAD_PRIORITY_IDLE );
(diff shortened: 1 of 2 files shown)


htibosch:
There is something strange: I recently submitted [PR #3006](https://github.com/aws/amazon-freertos/pull/3006), which was merged, but the same changes appear in this PR, as though the merge was not successful.

htibosch:
As the project "ecc608a_plus_winsim" has a copy of "aws_demo_logging.c", I updated that as well. See

`vendors/microchip/boards/ecc608a_plus_winsim/aws_demos\application_code`

I tested the changes again, using 4 FreeRTOS tasks and 4 Windows thread. They all produce a continuous stream of logging:
~~~
    22064 4219 [xxx] Task   0: 1004
    22065 4219 [xxx] Thread 0: 4509
    22066 4219 [xxx] Thread 1: 4485
    22067 4219 [xxx] Task   0: 1005
    22068 4219 [xxx] Thread 1: 4486
    22069 4219 [xxx] Task   0: 1006
    22070 4219 [xxx] Task   0: 1007
    22071 4219 [xxx] Thread 3: 4480
    22072 4219 [xxx] Thread 1: 4487
    22073 4219 [xxx] Thread 1: 4488
    22074 4219 [xxx] Thread 2: 4529
    22075 4219 [xxx] Task   0: 1008
    22076 4220 [xxx] Task   1: 1003
    22077 4220 [xxx] Thread 0: 4510
~~~

The threads get more attention than the tasks because they can run on any core, whereas FreeRTOS tasks have an affinity mask of 0x01.

When I removed the locking mechanism, the stream buffer would get corrupted.
Also, when I restored the old locking by raising thread priority, corruption would occur.

In `prvLoggingFlushBuffer()`, I added a few asserts that check if the stream buffer is corrupted. It is nicer to see an assert rather than a Windows exception.

n9wxu (approved):

AniruddhaKanhere (approved):

htibosch:
"Build Checks" says: Build finished. 132 tests run, 0 skipped, 0 failed.
And yet it gets a image ?

AniruddhaKanhere:
> "Build Checks" says: Build finished. 132 tests run, 0 skipped, 0 failed.
> And yet it gets a image ?

Yes, that means that some tests did not run. It is a known issue and we are actively working on fixing it.
Apologies for the trouble. For now, we shall re-trigger the checks and track them.

PR #3024, opened 2021-03-03
CY8CKIT-064S0S2-4343W: Fix aws_demo

RaymondNgun:
CY8CKIT-064S0S2-4343W: Fix aws_demo

Description
-----------
aws_demo is no longer connecting to AP. It looks like a merge error
resulted in the removal of the needed function call.

Signed-off-by: Raymond Ngun <raymond.ngun@infineon.com>

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/cypress/boards/CY8CKIT_064S0S2_4343W/aws_demos/application_code/main.c
+++ b/vendors/cypress/boards/CY8CKIT_064S0S2_4343W/aws_demos/application_code/main.c
@@ -383,7 +383,8 @@ void prvWifiConnect( void )
 
     if( xWifiStatus == eWiFiSuccess )
     {
-        configPRINTF( ( "Wi-Fi Connected to AP %s. Creating tasks which use network...\r\n", clientcredentialWIFI_SSID) );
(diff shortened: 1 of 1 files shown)


mingyue86010 (approved):

mingyue86010:
Providing some more data point here:

The error is not caused by missing WIFI_ConnectAP(). Actually we do not require the demo to explictly calls WIFI_ConnectAP(), since the Network Manager will handle that. AwsIotNetworkManager_EnableNetwork() should enable the network and that is called from demo runner. The reason that why the demo failed is because calling WIFI_GetIPInfo( &xIpConfig ) at line 393 without wifi connected which causes a kernel assert failure. If you remove this line you will get the demo runs.

mingyue86010:
However, although explictly calling WIFI_ConnectAP() is not required but currently all the board demo does this in this way. It definetly is redundent and need to be changed and imporved. But for now I think we could just accept this PR and let the demo runs.

dan4thewin (approved):

PR #3029, opened 2021-03-04
Fix indefinite wait issue in IotNetworkAfr_Destroy()

aggarw13:
### Problem

As part of the change in PR #2986, the `IotNetworkAfr_Close` function was updated to wait for the receive task to be exited and clear the `_FLAG_RECEIVE_TASK_EXITED` flag as part of the `xEventGroupWaitBits()` call. However, that affects the `IotNetworkAfr_Destroy` function to wait indefinitely as it also contains logic to block for the receive task to exit. 

### Solution

As the `IotNetworkAfr_Destroy` function is supposed to be called ONLY _after_ the connection is closed (as mentioned [here](https://github.com/aws/amazon-freertos/blob/201912.00/libraries/abstractions/platform/include/platform/iot_network.h\#L269-L270)), the receive task should have already exited (because of call to `IotNetworkAfr_Close`) before `IotNetworkAfr_Destroy` function is called, and thus, the wait for receive task to be exited is not required. 

This PR fixes the issue by removing the wait logic in the `IotNetorkAfr_Destroy` function and also makes hygiene improvements in the receive task and `IotNetworkAfr_Close` functions logic.

### DIFF
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -180,6 +180,10 @@ static void _networkReceiveTask( void * pArgument )
 
         pNetworkConnection->bufferedByteValid = true;
 
+        /* The network receive task is created ONLY when the receive callback is set. Thus, assert
+         * check that the callback is valid. */
+        configASSERT( pNetworkConnection->receiveCallback != NULL );
+
         /* Invoke the network callback. */
         pNetworkConnection->receiveCallback( pNetworkConnection,
                                              pNetworkConnection->pReceiveContext );
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -68,18 +68,18 @@
 /**
  * @brief The event group bit to set when a connection's socket is shut down.
  */
-#define _FLAG_SHUTDOWN                ( 1 )
+#define _FLAG_SHUTDOWN                             ( 1 )
 
 /**
  * @brief The event group bit to set when a connection's receive task exits.
  */
-#define _FLAG_RECEIVE_TASK_EXITED     ( 2 )
+#define _FLAG_RECEIVE_TASK_EXITED                  ( 2 )
 
 /**
  * @brief The event group bit to set when the connection is destroyed from the
  * receive task.
  */
-#define _FLAG_CONNECTION_DESTROYED    ( 4 )
+#define _FLAG_RECEIVE_TASK_CONNECTION_DESTROYED    ( 4 )
 
 /*-----------------------------------------------------------*/
 
@@ -92,7 +92,7 @@ typedef struct _networkConnection
     IotNetworkReceiveCallback_t receiveCallback; /**< @brief Network receive callback, if any. */
     void * pReceiveContext;                      /**< @brief The context for the receive callback. */
     bool bufferedByteValid;                      /**< @brief Used to determine if the buffered byte is valid. */
-    uint8_t bufferedByte;                        /**< @brief A single byte buffered from a receive, since AFR Secure Sockets does not have poll(). */
+    uint8_t bufferedByte;                        /**< @brief A single byte buffered from a receive, since FreeRTOS Secure Sockets does not have poll(). */
 } _networkConnection_t;
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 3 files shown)


libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -679,18 +685,22 @@ IotNetworkError_t IotNetworkAfr_Destroy( void * pConnection )
...
-                                     _FLAG_CONNECTION_DESTROYED );
+                                     _FLAG_RECEIVE_TASK_CONNECTION_DESTROYED );
     }
     else
     {
-        /* If a receive task was created, wait for it to exit. */
-        if( pNetworkConnection->receiveTask != NULL )
+        /* As this function should be called ONLY called after the connection is closed,
+         * the receive task should have already exited. */
+        if( pNetworkConnection->receiveCallback != NULL )
```
yourslab, line 694: What is the reason why we check the `receiveCallback` is not `NULL` here?

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -643,7 +648,8 @@ IotNetworkError_t IotNetworkAfr_Close( void * pConnection )
     ( void ) xEventGroupSetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ),
                                  _FLAG_SHUTDOWN );
 
-    if( pNetworkConnection->receiveTask != NULL )
+    /* If this function is not called from the receive task, wait for the receive task to exit. */
+    if( ( pNetworkConnection->receiveTask != NULL ) && ( xTaskGetCurrentTaskHandle() == pNetworkConnection->receiveTask ) )
```
yourslab, line 652: Do we want to enter this block if this function is called from the receive task? The comment seems to indicate otherwise.

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -189,11 +189,16 @@ static void _networkReceiveTask( void * pArgument )
...
+        /* Break out of receive task loop if connection is closed or destroyed. */
+        if( ( connectionFlags & _FLAG_RECEIVE_TASK_CONNECTION_DESTROYED ) == _FLAG_RECEIVE_TASK_CONNECTION_DESTROYED )
         {
             destroyConnection = true;
             break;
         }
+        else if( ( connectionFlags & _FLAG_SHUTDOWN ) == _FLAG_SHUTDOWN )
+        {
+            break;
+        }
```
yourslab, line 201: This block seems redundant? Don't we already do the same thing inside the `do...while` or is there something I'm missing?

libraries/abstractions/platform/freertos/iot_network_freertos.c
aggarw13: It is possible that the `receiveCallback` is not set by the application, in which case the network receive task will not be spawned by this network abstraction implementation. Refer [here](https://github.com/aws/amazon-freertos/blob/201912.00/libraries/abstractions/platform/freertos/iot_network_freertos.c#L441-L471) for the function that sets the receive callback.

libraries/abstractions/platform/freertos/iot_network_freertos.c
aggarw13: This is a different case to detect the calling of `IotNetworkAfr_Close` (or specifically, the `.close` member) function from within the receive callback.
The earlier logic within the do-while loop detects the connection closure if it happens _before_ the receive callback is invoked.

libraries/abstractions/platform/freertos/iot_network_freertos.c
aggarw13: Good catch, we don't want to enter this block! Have fixed it.

libraries/abstractions/platform/freertos/iot_network_freertos.c
yourslab: If the application sets `receiveCallback` to `NULL`, won't there be a null-pointer dereference when `_networkReceiveTask` is called? Should the same check also be added there as well?

libraries/abstractions/platform/freertos/iot_network_freertos.c
yourslab: Okay, that makes sense. Need to handle both cases.

libraries/abstractions/platform/freertos/iot_network_freertos.c
aggarw13: The `_networkReceiveTask` is created only when the `receiveCallback` callback is non-NULL. We can add an assert check in the receive task for the validity of the receive callback

yourslab (approved):

n9wxu (approved):

PR #3032, opened 2021-03-08
NimBLE AFR: Fix long read response in case of smaller MTU value.

prasad-alatkar:
NimBLE AFR: Fix long read response in case of smaller MTU value

Description
-----------

- NimBLE host handles offset internally, so application does not need to worry about sending the data in (MTU - 1) fragments as few other stack expect the application to do so.
- However, this NimBLE stack feature can cause issue with few vendor agnostic applications where they do the fragmentation (e.g. FFS).
- The solution proposed in this PR is to track the offset virtual offset. This way the AFR layer maintains the `virtual_offset` in case of read long characteristic response. 
- This way ESP32 NimBLE AFR porting layer will now support the applications' behavior of handling the data fragmentation depending upon MTU value by themselves. 

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -54,8 +54,8 @@ static SemaphoreHandle_t xSem;
 static bool semInited;
 bool xSemLock = 0;
 uint16_t gattOffset = 0;
-static bool long_read_flag;
-static size_t virtual_offset;/* Maintain offset for long read ops */
+static bool long_read_flag = false;
+static size_t virtual_offset = 0;/* Maintain offset for long read ops */
 
 void prvGattGetSemaphore()
 {
@@ -763,13 +763,13 @@ BTStatus_t prvBTSendResponse( uint16_t usConnId,
             if ( pxResponse->xAttrValue.xLen >= ( mtu_value - 1 ) )
             {
                 /* Long read or blob request if data received is of ( MTU - 1 ) size and more */
-                long_read_flag = 1;
+                long_read_flag = true;
                 prvPrepareLongReadResponse( dst_buf, pxResponse->xAttrValue.pucValue, mtu_value - 1 );
             }
             else
             {
                 /* Last long read request if data received is < ( MTU - 1 ) size */
-                long_read_flag = 0;
+                long_read_flag = false;
                 prvPrepareLongReadResponse( dst_buf, pxResponse->xAttrValue.pucValue, pxResponse->xAttrValue.xLen );
             }
 
--- a/vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
+++ b/vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
@@ -45,6 +45,8 @@
 
 #define BLE_GATT_DSC_CLT_CFG_UUID128 0xFB, 0x34, 0x9B, 0x5F, 0x80, 0x00, 0x00, 0x80, 0x00, 0x10, 0x00, 0x00, 0x02, 0x29, 0x00, 0x00
 
+#define BLE_MAX_ATTR_LEN 512
+
 static struct ble_gatt_svc_def espServices[ MAX_SERVICES + 1 ];
 static BTService_t * afrServices[ MAX_SERVICES ];
 static uint16_t serviceCnt = 0;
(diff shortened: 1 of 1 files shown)


prasad-alatkar:
@ravibhagavandas PTAL.

prasad-alatkar:
CC: @dhrishi @mahavirj

ravibhagavandas (dismissed):

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -726,6 +728,19 @@ static bool prvValidGattRequest()
     return false;
 }
 
+static void prvPrepareLongReadResponse( uint8_t * dst, uint8_t * pucValue, size_t xLen )
+{
+    if ( ( virtual_offset + xLen ) <= 512 )
```
dan4thewin, line 733: Could we make the constant into a macro with a doc string?  Is 512 unalterable, or should it be a configuration value?  Should we pass in the size of dst and perhaps assert rather than logging a message?  I'm concerned that this function can fail, but returns nothing, so the application proceeds as if it had succeeded.

prasad-alatkar: @dan4thewin Thank you for reviewing the PR. I have added assert as per your suggestion. I have also added macro for 512 as per your suggestion. To answer your query on 512 unalterable, yes the value is maximum GATT attribute size so it is not alterable.

ravibhagavandas (dismissed):

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -45,13 +45,17 @@
...
+#define BLE_MAX_ATTR_LEN 512
+
 static struct ble_gatt_svc_def espServices[ MAX_SERVICES + 1 ];
 static BTService_t * afrServices[ MAX_SERVICES ];
 static uint16_t serviceCnt = 0;
 static SemaphoreHandle_t xSem;
 static bool semInited;
 bool xSemLock = 0;
 uint16_t gattOffset = 0;
+static bool long_read_flag;
```
gkwicker, line 57: The variables on lines 57&58 should be explicitly initialized, e.g. `static bool long_read_flag = false;`

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -734,18 +752,40 @@ BTStatus_t prvBTSendResponse( uint16_t usConnId,
...
 
     if( prvValidGattRequest() )
     {
         if( ctxt && ( ( ctxt->op == BLE_GATT_ACCESS_OP_READ_CHR ) || ( ctxt->op == BLE_GATT_ACCESS_OP_READ_DSC ) ) )
         {
+            ESP_LOGD(TAG, "Send response callback; long_read_flag = %d, xLen received = %d, offset = %d", long_read_flag, pxResponse->xAttrValue.xLen, virtual_offset);
+            if ( pxResponse->xAttrValue.xLen >= ( mtu_value - 1 ) )
+            {
+                /* Long read or blob request if data received is of ( MTU - 1 ) size and more */
+                long_read_flag = 1;
```
gkwicker, line 766: `long_read_flag` is of type `bool`, so shouldn't we do `long_read_flag = true;` here?

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
```diff
@@ -734,18 +752,40 @@ BTStatus_t prvBTSendResponse( uint16_t usConnId,
...
+                prvPrepareLongReadResponse( dst_buf, pxResponse->xAttrValue.pucValue, mtu_value - 1 );
+            }
+            else
+            {
+                /* Last long read request if data received is < ( MTU - 1 ) size */
+                long_read_flag = 0;
```
gkwicker, line 772: See previous comment, `long_read_flag = false;`

gkwicker (changes requested):
See comments

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gatt_server.c
prasad-alatkar: Okay, I did not explicitly initialized them as they are static variables, will update.

gkwicker (approved):

Issue #3075, opened 2021-03-30
_httpReconnect always fails


Sherryzhang2:
**Describe the bug**
In the OTA process, I choice the HTTPS to download the image data. But it always fails in _httpReconnect.
Th reason is that in IotHttpsClient_Connect, it disconnects the connection by calling IotHttpsClient_Disconnect. IotHttpsClient_Disconnect  firstly calls  _networkDisconnect to disconnect the connection in which the _FLAG_RECEIVE_TASK_EXITED flag is cleared and the _FLAG_SHUTDOWN flag is set. Then IotHttpsClient_Disconnect  calls _networkDestroy to destroy the connect in which it asserts that connection flag _FLAG_RECEIVE_TASK_EXITED bit is set.

Proposed fix, in IotNetworkAfr_Destroy, the connection flag _FLAG_SHUTDOWN should be used in the assert instead of the _FLAG_RECEIVE_TASK_EXITED flag which has already be cleared in  IotNetworkAfr_Close.

Sherryzhang2:
I think the root cause of https://github.com/aws/amazon-freertos/issues/3073 is the same as this issue.

muneebahmed10:
Hi @Sherryzhang2,

Thank you for bringing this to our attention. I've created #3082 to fix this. Can you pull that change and confirm that the issue is resolved?

Thank you

muneebahmed10:
Hi @Sherryzhang2,

Thank you again for reporting this bug. I will close this issue. Please open another issue if you have any other questions or bug reports.

Thank you

Sherryzhang2:
@muneebahmed10 Thanks for your prompt fix.

PR #3082, opened 2021-03-31
Fix assert in IotNetworkAfr_Destroy

muneebahmed10:
Description
-----------

Reported in #3075. #2986 updated `IotNetworkAfr_Close` to [clear](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/platform/freertos/iot_network_freertos.c#L661) the `_FLAG_RECEIVE_TASK_EXITED` flag, and #3029 updated `IotNetworkAfr_Destroy` to remove a check for that flag, as it would otherwise wait for it indefinitely. However, it added an [assert](https://github.com/aws/amazon-freertos/blob/master/libraries/abstractions/platform/freertos/iot_network_freertos.c#L703) for the same flag that was cleared, resulting in the failure in #3075. This ~removes that assert so that it will not trigger when calling `IotNetworkAfr_Destroy`~ changes `IotNetworkAfr_Close` to not clear the flag.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -693,20 +693,6 @@ IotNetworkError_t IotNetworkAfr_Destroy( void * pConnection )
     }
     else
     {
-        /* As this function should be called ONLY called after the connection is closed,
-         * the receive task should have already exited. */
-        if( pNetworkConnection->receiveCallback != NULL )
-        {
-            EventBits_t connectionFlags;
-            connectionFlags = xEventGroupGetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ) );
-
-            configASSERT( ( connectionFlags & _FLAG_RECEIVE_TASK_EXITED ) == _FLAG_RECEIVE_TASK_EXITED );
-
-            /* Suppress compiler warning of unused connectionFlags variable when
-             * configASSERT() is disabled. */
-            ( void ) connectionFlags;
-        }
-
         _destroyConnection( pNetworkConnection );
     }
 
--- a/libraries/abstractions/platform/freertos/iot_network_freertos.c
+++ b/libraries/abstractions/platform/freertos/iot_network_freertos.c
@@ -657,10 +657,10 @@ IotNetworkError_t IotNetworkAfr_Close( void * pConnection )
     {
         /* Wait for the network receive task to exit so that the socket can be shutdown safely
          * without causing the socket to block forever if there are pending reads or writes
-         * from other tasks. */
+         * from other tasks. Do not clear the flag as IotNetworkAfr_Destroy will do so. */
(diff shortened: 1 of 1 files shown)


libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -693,20 +693,6 @@ IotNetworkError_t IotNetworkAfr_Destroy( void * pConnection )
     }
     else
     {
-        /* As this function should be called ONLY called after the connection is closed,
-         * the receive task should have already exited. */
-        if( pNetworkConnection->receiveCallback != NULL )
```
aggarw13, line 698: It will be valuable to retain this check for ensuring that the receive task has been deleted before the socket resources are cleared.
Instead of removing this assert check, can you update the `IotNetworkAfr_Close` function to _not_ clear the `_FLAG_RECEIVE_TASK_EXITED` flag?

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -657,10 +657,10 @@ IotNetworkError_t IotNetworkAfr_Close( void * pConnection )
     {
         /* Wait for the network receive task to exit so that the socket can be shutdown safely
          * without causing the socket to block forever if there are pending reads or writes
-         * from other tasks. */
+         * from other tasks. Do not clear the flag as IotNetworkAfr_Destroy will do so. */
         ( void ) xEventGroupWaitBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ),
                                       _FLAG_RECEIVE_TASK_EXITED,
-                                      pdTRUE,
+                                      pdFALSE,
```
aggarw13, line 663: It would be good to add a comment about the purpose of not clearing the flag here.

libraries/abstractions/platform/freertos/iot_network_freertos.c
```diff
@@ -698,7 +698,13 @@ IotNetworkError_t IotNetworkAfr_Destroy( void * pConnection )
         if( pNetworkConnection->receiveCallback != NULL )
         {
             EventBits_t connectionFlags;
-            connectionFlags = xEventGroupGetBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ) );
+
+            /* Clear the flag indicating the receive task has exited. */
+            connectionFlags = xEventGroupWaitBits( ( EventGroupHandle_t ) &( pNetworkConnection->connectionFlags ),
```
aggarw13, line 703: The bit should be cleared after the assert check of the `_FLAG_RECEIVE_TASK_EXITED` flag

muneebahmed10: Removed the wait

libraries/abstractions/platform/freertos/iot_network_freertos.c
muneebahmed10: Updated

aggarw13 (approved):

dan4thewin (approved):

Issue #3073, opened 2021-03-27
IotMqtt_Disconnect sometimes not returning but leaving task blocked forever


lightblu:
Hi,

I just tried updating from the 202007.00 release to 202012.00. 

We have integrated MQTT libraries to do MQTT on top of our own SOCKETS_ interface implementation.
We are still using the MQTT V2.x.x APIs, which seems to be now the "MQTT Compatibility Layer" wrapper to the new coreMQTT library (as I understand it from https://github.com/aws/amazon-freertos/blob/master/libraries/c_sdk/standard/mqtt/README.md - it is quite taxing that this is now the third bigger API move within 2 years after already going from MQTT V1 to V2 - is there any stability on the horizon with the LTS right now? :/ )

Before with the 202007.00 all MQTT functionality was stable, however with the 202012.00 updae we sometimes (but not always) get the situation that IotMqtt_Disconnect does not return but just leaves the calling task hanging forever. Here is an example log, "cloud" is the task using the IotMqtt_x API functions:

    2021-03-26 22:23:38 [TRACE] (   cloud): CLOSE MQTT id=1616797339  (our message, before publishing a last message)

    2021-03-26 22:23:38 [TRACE] (   cloud): [INFO ][CMQTT][90021] (MQTT connection 0x200162d0) MQTT PUBLISH operation queued.
    2021-03-26 22:23:38 [TRACE] (   cloud): [INFO ][CMQTT][90021] (MQTT connection 0x200162d0, PUBLISH operation 0x20016388) Waiting for operation completion.
    2021-03-26 22:23:38 [TRACE] (   cloud): [INFO ][CMQTT][90462] (MQTT connection 0x200162d0, PUBLISH operation 0x20016388) Wait complete with result SUCCESS.
    2021-03-26 22:23:38 [TRACE] (   cloud): calling IotMqtt_Disconnect() 
    2021-03-26 22:23:38 [TRACE] (   cloud): [INFO ][CMQTT][90463] (MQTT connection 0x200162d0) Disconnecting connection.
    2021-03-26 22:23:39 [TRACE] (   cloud): SOCKETS_Shutdown[0] 2
    2021-03-26 22:23:39 [TRACE] (iot_thre): [ERROR][CMQTT][90860] (MQTT connection 0x200162d0) Failed to receive PINGRESP within 12000 ms.
    2021-03-26 22:23:39 [TRACE] (iot_thre): SOCKETS_Shutdown[0] 2
[...]

PR #3100, opened 2021-04-15
Check for NULL before freeing subscriptions

muneebahmed10:
Description
-----------

Reported in #3073. #3072 fixed a memory leak by freeing allocated buffers when removing subscriptions. However, for the case of `IotMqtt_RemoveAllMatches` for a `NULL` matching function, all subscriptions are removed in a loop regardless of whether or not they exist. This is fine for dynamic allocation, as `free( NULL )` does nothing, but when using static memory, `IotStaticMemory_ReturnInUse` tries to [dereference](https://github.com/aws/amazon-freertos/blob/master/libraries/c_sdk/standard/common/iot_static_memory_common.c#L126) the pointer without checking first for NULL.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/c_sdk/standard/mqtt/src/iot_mqtt_subscription_container.c
+++ b/libraries/c_sdk/standard/mqtt/src/iot_mqtt_subscription_container.c
@@ -272,8 +272,12 @@ bool IotMqtt_RemoveSubscription( _mqttSubscription_t * pSubscriptionArray,
         pSubscriptionArray[ deleteIndex ].topicFilterLength = 0;
 
         /* Free memory allocated for topic filter. */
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (approved):

aggarg (approved):

lightblu:
Hi,
thanks for tackling/improving this. When I recently discussed  the issue with a colleague, we came to basically that

> This is fine for dynamic allocation, as free( NULL ) does nothing, but when using static memory, IotStaticMemory_ReturnInUse tries to dereference the pointer

So while I believe and agree that preventing null-pointer-"frees" is a good idea, I wondered if `IotStaticMemory_ReturnInUse` shouldn't also just stick more to the general free-behavior of just ignoring `IotStaticMemory_ReturnInUse(NULL,`

PR #3103, opened 2021-04-15
Add a delay in demos after MQTT_ProcessLoop.

yngki:
Added a short delay in demos after MQTT_ProcessLoop to give CPU cycles to other tasks.

Checklist:
----------

- [ ] My code is Linted.


xuelix (approved):

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -1191,6 +1191,8 @@ static MQTTStatus_t prvWaitForPacket( MQTTContext_t * pxMQTTContext,
         /* Event callback will set #usPacketTypeReceived when receiving appropriate packet. This
          * will wait for at most mqttexamplePROCESS_LOOP_TIMEOUT_MS. */
         xMQTTStatus = MQTT_ProcessLoop( pxMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+
+        vTaskDelay( mqttexampleDELAY_BETWEEN_PUBLISHES_TICKS );
```
aggarw13, line 1195: What issue does adding the delay cause in CI? It will be valuable to add the a comment for the reasoning

demos/coreMQTT/mqtt_demo_mutual_auth.c
```diff
@@ -1191,6 +1191,8 @@ static MQTTStatus_t prvWaitForPacket( MQTTContext_t * pxMQTTContext,
         /* Event callback will set #usPacketTypeReceived when receiving appropriate packet. This
          * will wait for at most mqttexamplePROCESS_LOOP_TIMEOUT_MS. */
         xMQTTStatus = MQTT_ProcessLoop( pxMQTTContext, mqttexamplePROCESS_LOOP_TIMEOUT_MS );
+
```
gkwicker, line 1194: These demos are often used as a starting point when developing a new application.  Why does a `vTaskDelay()` need to be here?  Would all applications using this library follow this pattern, or are we adding this due to a special case in one particular application?

gkwicker (commented):
See comment

n9wxu:
The fix suggested does not resolve the delay caused by processloop when the processloop timeout is non-zero.

PR #3127, opened 2021-04-28
[Mqtt shim] Memory leak fix for freeing buffers when serializing MQTT packets fails

gshvang:
Several wrappers in the MQTT shim allocate buffers to hold serialized packets, but do not free them on error. This PR contains the fix for freeing those buffers on error.

### DIFF
--- a/libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
+++ b/libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
@@ -23,11 +23,11 @@
  * https://www.FreeRTOS.org
  */
 
- /**
-  * @file iot_mqtt_serializer_deserializer_wrapper.c
-  * @brief Implements serializer and deserializer wrapper functions for the shim.
-  */
-  /* The config header is always included first. */
+/**
+ * @file iot_mqtt_serializer_deserializer_wrapper.c
+ * @brief Implements serializer and deserializer wrapper functions for the shim.
+ */
+/* The config header is always included first. */
 #include "iot_config.h"
 
 /* Standard includes. */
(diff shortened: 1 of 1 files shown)


libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -395,9 +422,6 @@ IotMqttError_t _IotMqtt_subscribeSerializeWrapper( const IotMqttSubscription_t *
         *pPacketIdentifier = packetId;
     }
 
-    /* Free allocated memory as the packet was serialized. */
-    IotMqtt_FreeMessage( subscriptionList );
```
muneebahmed10, line 399: It looks like this introduces another memory leak by not freeing the subscription list on success? I believe restoring this line and deleting lines 385 and 412 would alway free the subscription list

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -481,9 +520,6 @@ IotMqttError_t _IotMqtt_unsubscribeSerializeWrapper( const IotMqttSubscription_t
         *pPacketIdentifier = packetId;
     }
 
-    /* Free allocated memory as the packet was serialized. */
-    IotMqtt_FreeMessage( unsubscriptionList );
```
muneebahmed10, line 485: Same comment as above

libraries/c_sdk/standard/mqtt/src/iot_mqtt_serializer_deserializer_wrapper.c
```diff
@@ -396,7 +416,7 @@ IotMqttError_t _IotMqtt_subscribeSerializeWrapper( const IotMqttSubscription_t *
     }
 
     /* Free allocated memory as the packet was serialized. */
-    IotMqtt_FreeMessage( subscriptionList );
+    IotMqtt_FreeMessage(subscriptionList);
```
muneebahmed10, line 419: Uncrustify. Are you using Visual Studio? I'd recommend against that for editing code and using a text editor like VS Code instead.

muneebahmed10 (approved):

aggarg (approved):

PR #3134, opened 2021-04-30
Update MQTT Agent submodule and add interface

muneebahmed10:
Description
-----------

Updates the coreMQTT-Agent submodule and adds the files for the FreeRTOS interface.

Checklist:
----------

- [ ] My code is Linted.


libraries/core_mqtt_agent_demo_dependencies.cmake
```diff
@@ -0,0 +1,132 @@
...
+# core_mqtt_agent_demo_dependencies module.
+# Metadata module used for the MQTT Agent library in the FreeRTOS console.
+# It represents a collection of module dependencies required
+# by the coreMQTT Agent demos.
+# This module enables the FreeRTOS console experience of enabling
+# library dependencies when selecting the MQTT Agent library, so that the
+# coreMQTT Agent demos can be downloaded.
+afr_module(NAME core_mqtt_agent_demo_dependencies )
+
+afr_set_lib_metadata(ID "core_mqtt_agent_demo_dependencies")
```
aggarw13, line 56: This metadata configuration and `core_mqtt_agent_dependencies` CMake target are only required if coreMQTT Agent is slated for display on FreeRTOS console.

aggarw13: If we want to display the option of downloading the coreMQTT Agent demo for every board that supports a coreMQTT demo, then, we can just utilize the `core_mqtt_agent_demo_dependencies` target for displaying the MQTT library on FreeRTOS console. Thus, we will need to remove the [`core_mqtt_demo_dependencies`](https://github.com/aws/amazon-freertos/blob/202012.00/libraries/core_mqtt_demo_dependencies.cmake#L45-L128) target

demos/coreMQTT_Agent/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+# FreeRTOS coreMQTT demos
+afr_demo_module(core_mqtt_agent)
+
+# Set the default demo that is enabled in the downloaded package.
+afr_set_demo_metadata(ID "CORE_MQTT_CONNECTION_SHARING_DEMO")
+afr_set_demo_metadata(DESCRIPTION "Examples that demonstrate the MQTT library")
+afr_set_demo_metadata(DISPLAY_NAME "coreMQTT Demos")
```
aggarw13, line 7: The metadata information for the coreMQTT Agent demos should be separate from the coreMQTT demos as they represented by 2 different demo targets now
```suggestion
afr_set_demo_metadata(DESCRIPTION "Examples that demonstrate the coreMQTT Agent library")
afr_set_demo_metadata(DISPLAY_NAME "coreMQTT Agent Demo")
```

aggarw13:
For the failing builds, you need to remove references to the path of the old `mqtt_demo_connection_sharing` from project files of boards

libraries/abstractions/mqtt_agent/mqtt_agent_interface.cmake
```diff
@@ -0,0 +1,36 @@
+# MQTT Agent Interface
+afr_module(NAME mqtt_agent_interface INTERNAL)
+
+set(src_dir "${CMAKE_CURRENT_LIST_DIR}")
+
+set(include_dir "${CMAKE_CURRENT_LIST_DIR}/include")
+
+# Include filepaths for source and include.
+include( ${CMAKE_CURRENT_LIST_DIR}/transport_interface.cmake )
```
aggarw13, line 9: Inclusion of `transport_interface.cmake` shouldn't be required

demos/coreMQTT_Agent/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+afr_module_include_dirs(
+    ${AFR_CURRENT_MODULE}
+    INTERFACE
+       "${AFR_DEMOS_DIR}/include"
+)
+
+afr_module_dependencies(
+    ${AFR_CURRENT_MODULE}
+    INTERFACE
```
aggarw13, line 31: There needs to be a dependency on `AFR::mqtt_agent_interface` module for the interface implementation files

aggarw13: Similar dependency needs to be added to the `core_mqtt_agent_demo_dependencies` target as well

libraries/abstractions/mqtt_agent/mqtt_agent_interface.cmake
```diff
@@ -0,0 +1,36 @@
+# MQTT Agent Interface
+afr_module(NAME mqtt_agent_interface INTERNAL)
```
aggarw13, line 2: As the name of this module (`mqtt_agent_interface`) does not match its parent folder name (i.e. `mqtt_agent`), this CMake files needs to be explicitly added to the metadata list with the following CMake function call.
```
# Add cmake files of module to metadata.
afr_module_cmake_files(${AFR_CURRENT_MODULE} 
    ${CMAKE_CURRENT_LIST_DIR}/mqtt_agent_interface.cmake
)
```

Alternatively, you can rename the parent folder to match the CMake module name to have this file automatically included in the CMake metadata list

libraries/abstractions/mqtt_agent/freertos_agent_message.c
```diff
@@ -0,0 +1,74 @@
+/*
+ * FreeRTOS V202011.00
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
aggarw13, line 3: Are these abstraction files under `mqtt_agent` folder exactly the same as those in `FreeRTOS/FreeRTOS` [here](https://github.com/FreeRTOS/FreeRTOS/tree/202104.00/FreeRTOS-Plus/Demo/Common/coreMQTT_Agent_Interface)?

libraries/abstractions/mqtt_agent/mqtt_agent_interface.cmake
muneebahmed10: Isn't that call already present at lines 8-11? I think I'll just keep it as is

libraries/abstractions/mqtt_agent/freertos_agent_message.c
muneebahmed10: Not exactly. They were taken from the demos repo https://github.com/FreeRTOS/coreMQTT-Agent-Demos/tree/main/lib/FreeRTOS/mqtt-agent-interface. Besides the headers the code is the same

demos/coreMQTT_Agent/CMakeLists.txt
muneebahmed10: Added

libraries/abstractions/mqtt_agent/mqtt_agent_interface.cmake
muneebahmed10: Removed

demos/coreMQTT_Agent/CMakeLists.txt
muneebahmed10: I removed this for now. I'll add it back once we port the actual demo

libraries/abstractions/mqtt_agent/mqtt_agent_interface.cmake
aggarw13: Oops, I didn't notice it.

aggarw13 (approved):

aggarg (approved):

PR #3146, opened 2021-05-10
Add MQTT Agent demo

muneebahmed10:
Description
-----------

Remove old coreMQTT connection sharing demo and replace with MQTT agent demo

Checklist:
----------

- [ ] My code is Linted.


demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+/* Exponential backoff retry include. */
+#include "backoff_algorithm.h"
+
+/* Include PKCS11 helpers header. */
+#include "pkcs11_helpers.h"
+
+/* Subscription manager header include. */
+#include "subscription_manager.h"
+
+
```
aggarw13, line 85:
```suggestion

```

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "task.h"
+
+/* Demo Specific configs. */
+#include "mqtt_demo_connection_sharing_config.h"
```
aggarw13, line 64: Should this config file be changed to something like `mqtt_agent_demo_config.h` to make the connection of the file with this demo more obvious?

vendors/cypress/boards/CY8CKIT_064S0S2_4343W/aws_demos/config_files/mqtt_demo_connection_sharing_config.h
```diff
@@ -92,6 +92,13 @@
  * #define democonfigMQTT_BROKER_PORT    ( insert here. )
  */
 
+/**
+ * @brief The number of simple subscribe-publish tasks to create for the demo
+ */
+#define democonfigNUM_SIMPLE_SUB_PUB_TASKS_TO_CREATE       1
+#define democonfigSIMPLE_SUB_PUB_TASK_STACK_SIZE           ( configMINIMAL_STACK_SIZE )
+
```
aggarw13, line 100: Should there also be greyed out config documentation for `MQTT_AGENT_NETWORK_BUFFER_SIZE` and `MQTT_AGENT_COMMAND_QUEUE_LENGTH` macros as is the case for the connection setting macros?

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+        }
+    #endif
+
+
+    /* Create an instance of the MQTT agent task. */
+    xTaskCreate( prvMQTTAgentTask,
+                 "MQTT Agent",
+                 configMINIMAL_STACK_SIZE,
+                 NULL,
+                 tskIDLE_PRIORITY + 1,
```
aggarw13, line 980: Can you add a comment about why the agent task has higher priority than the application tasks?

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+            LogError( ( "All %d iterations failed", democonfigMQTT_MAX_DEMO_COUNT ) );
+        }
+    }
+
+    return ret;
+}
+
+/*-----------------------------------------------------------*/
+
+static MQTTStatus_t prvMQTTInit( void )
```
aggarw13, line 477: (Nit) Maybe, this is more fitting of a name?
```suggestion
static MQTTStatus_t prvMQTTAgentInit( void )
```

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+            xNetworkResult = prvSocketDisconnect( &xNetworkContext );
+        }
+        else if( xMQTTStatus == MQTTSuccess )
+        {
+            /* MQTTAgent_Terminate() was called, but MQTT was not disconnected. */
+            xConnectStatus = MQTT_Disconnect( &( xGlobalMqttAgentContext.mqttContext ) );
+            xNetworkResult = prvSocketDisconnect( &xNetworkContext );
+            break;
+        }
+        /* Error. */
```
aggarw13, line 895: What kind of errors does this represent for which connection is retried and the agent resumed?

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+                                        MQTTPublishInfo_t * pxPublishInfo )
+{
+    bool xPublishHandled = false;
+    char cOriginalChar, * pcLocation;
+
+    ( void ) packetId;
+
+    /* Fan out the incoming publishes to the callbacks registered using
+     * subscription manager. */
+    xPublishHandled = handleIncomingPublishes( ( SubscriptionElement_t * ) pMqttAgentContext->pIncomingCallbackContext,
```
aggarw13, line 845: Would have been more readable if the name of the function has the `MQTTSubscriptionManager_` prefix in it to clearly mention that this publish callback delegates the incoming publish to the subscription manager
```suggestion
    xPublishHandled = handleIncomingPublishes( ( SubscriptionElement_t * ) pMqttAgentContext->pIncomingCallbackContext,
```

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+
+static MQTTAgentMessageContext_t xCommandQueue;
+
+static struct DemoParams taskParameters[ democonfigNUM_SIMPLE_SUB_PUB_TASKS_TO_CREATE ];
+
+/**
+ * @brief The global array of subscription elements.
+ *
+ * @note No thread safety is required to this array, since the updates the array
+ * elements are done only from one task at a time. The subscription manager
```
aggarw13, line 420:
```suggestion
 * @note No thread safety is required to this array, since updates to the array
 * elements are done only from one task at a time. The subscription manager
```

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+ *    demo_config.h.  For example, if demo_config.h contains the following
+ *    setting:
+ *
+ *    #define democonfigNUM_SIMPLE_SUB_PUB_TASKS_TO_CREATE 3
+ *
+ *    then the initial task will create three instances of the task
+ *    implemented in simple_sub_pub_demo.c.  See the comments at the top
+ *    of that file for more information.
+ *
+ * 3) After creating the demo tasks the initial task could create the MQTT
```
aggarw13, line 45: Is this paragraph valid about the initial task assuming the role of the agent task? This demo calls `xTaskCreate` on the `prvMqttAgentTask` function, thereby creating a separate task for the agent

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+    uint32_t ulTaskNumber = pxParams->ulTaskNumber;
+    MQTTQoS_t xQoS;
+    TickType_t xTicksToDelay;
+    MQTTAgentCommandInfo_t xCommandParams = { 0 };
+    char * pcTopicBuffer = topicBuf[ ulTaskNumber ];
+    uint32_t numSuccesses = 0U;
+
+    /* Have different tasks use different QoS.  0 and 1.  2 can also be used
+     * if supported by the broker. */
+    xQoS = ( MQTTQoS_t ) ( ulTaskNumber % 2UL );
```
aggarw13, line 456: Should the `2UL` be a configurable macro to allow the demo to be tested with Qos2 if the broker supports it?

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+     * is used as the command context and is sent back to this task as a notification
+     * in the callback that executed upon receipt of the subscription acknowledgment.
+     * That way this task can match an acknowledgment to a subscription. */
+    xTaskNotifyStateClear( NULL );
+    taskENTER_CRITICAL();
+    {
+        ulNextSubscribeMessageID++;
+        ulSubscribeMessageID = ulNextSubscribeMessageID;
+    }
+    taskEXIT_CRITICAL();
```
aggarw13, line 376: Comment about why this is in the Critical section will be valuable

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+               pcTopicFilter,
+               ( int ) ulSubscribeMessageID ) );
+
+    do
+    {
+        /* TODO: prvIncomingPublish as publish callback. */
+        xCommandAdded = MQTTAgent_Subscribe( &xGlobalMqttAgentContext,
+                                             &xSubscribeArgs,
+                                             &xCommandParams );
+    } while( xCommandAdded != MQTTSuccess );
```
aggarw13, line 411: Why is `MQTTAgent_Subscribe` call in a loop? Does this mean that enqueuing of the subsription command will be blocking until it succeeds? 
Can this cause an overfilling of the agent queue in the failure case that a broker has suddenly severed an MQTT connection, and the re-connection attempts in the main task are failing but in the meanwhile multiple demo tasks are stuck in adding subscription commands because the queue is full?

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+
+    /* Wait for acks to the subscribe message - this is optional but done here
+     * so the code below can check the notification sent by the callback matches
+     * the ulNextSubscribeMessageID value set in the context above. */
+    xCommandAcknowledged = prvWaitForCommandAcknowledgment( NULL );
```
aggarw13, line 416: Why has `NULL` been passed to the function if the goal is the match the ulNextSubscribeMessaeID value?

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+
+    /* Check both ways the status was passed back just for demonstration
+     * purposes. */
+    if( ( xCommandAcknowledged != pdTRUE ) ||
+        ( xApplicationDefinedContext.xReturnStatus != MQTTSuccess ) )
+    {
+        LogInfo( ( "Error or timed out waiting for ack to subscribe message topic %s",
```
aggarw13, line 423:
```suggestion
        LogWarn( ( "Error or timed out waiting for ack to subscribe message topic %s",
```

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+    static char cTerminatedString[ mqttexampleSTRING_BUFFER_LENGTH ];
+
+    ( void ) pvIncomingPublishCallbackContext;
+
+    /* Create a message that contains the incoming MQTT payload to the logger,
+     * terminating the string first. */
+    if( pxPublishInfo->payloadLength < mqttexampleSTRING_BUFFER_LENGTH )
+    {
+        memcpy( ( void * ) cTerminatedString, pxPublishInfo->pPayload, pxPublishInfo->payloadLength );
+        cTerminatedString[ pxPublishInfo->payloadLength ] = 0x00;
```
aggarw13, line 341: NULL termination isn't required when the string can be logged as `%.s*`

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "queue.h"
+#include "task.h"
+
+/* Demo Specific configs. */
+#include "mqtt_demo_connection_sharing_config.h"
+
+/* MQTT library includes. */
+#include "core_mqtt.h"
```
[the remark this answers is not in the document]
muneebahmed10: I don't think so

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: I didn't want to change it right now and have some build break, but it needs to be done eventually

vendors/cypress/boards/CY8CKIT_064S0S2_4343W/aws_demos/config_files/mqtt_demo_connection_sharing_config.h
muneebahmed10: Sure

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+        ret = prvConnectAndCreateDemoTasks( NULL );
+
+        if( ret == EXIT_SUCCESS )
+        {
+            LogInfo( ( "Demo iteration %lu successful.", ulDemoCount ) );
+            break;
+        }
+        else if( ulDemoCount < ( democonfigMQTT_MAX_DEMO_COUNT - 1 ) )
+        {
+            LogWarn( ( "Demo iteration %lu failed. Retrying...", ulDemoCount ) );
```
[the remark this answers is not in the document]
muneebahmed10: OK

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: Sure

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: Any error. Any status code that's not `MQTTSuccess`.

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: I think the `MQTTSubscriptionManager_` prefix sort of implies it's a separate library instead of part of the demo, so that might be confusing. As it is now, the subscription manager code is the same across all the repositories this demo is in, and I don't want to change it in only this location.

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+    uint16_t usNumSubscriptions = 0U;
+
+    /* These variables need to stay in scope until command completes. */
+    static MQTTAgentSubscribeArgs_t xSubArgs = { 0 };
+    static MQTTSubscribeInfo_t xSubInfo[ SUBSCRIPTION_MANAGER_MAX_SUBSCRIPTIONS ] = { 0 };
+    static MQTTAgentCommandInfo_t xCommandParams = { 0 };
+
+    /* Loop through each subscription in the subscription list and add a subscribe
+     * command to the command queue. */
+    for( ulIndex = 0U; ulIndex < SUBSCRIPTION_MANAGER_MAX_SUBSCRIPTIONS; ulIndex++ )
```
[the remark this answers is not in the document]
muneebahmed10: Yeah sure

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -0,0 +1,1025 @@
...
+/**
+ * @brief The global array of subscription elements.
+ *
+ * @note No thread safety is required to this array, since the updates the array
+ * elements are done only from one task at a time. The subscription manager
+ * implementation expects that the array of the subscription elements used for
+ * storing subscriptions to be initialized to 0. As this is a global array, it
+ * will be initialized to 0 by default.
+ */
+SubscriptionElement_t xGlobalSubscriptionList[ SUBSCRIPTION_MANAGER_MAX_SUBSCRIPTIONS ];
```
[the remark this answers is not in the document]
muneebahmed10: Yes, it's only accessed by the agent task through the command completion callbacks.

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: Not any more

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+     * attempts to post to it time out.  The queue will not become full if the
+     * priority of the MQTT agent task is higher than the priority of the task
+     * calling this function. */
+    LogInfo( ( "Sending subscribe request to agent for topic filter: %s with id %d",
+               pcTopicFilter,
+               ( int ) ulSubscribeMessageID ) );
+
+    do
+    {
+        /* TODO: prvIncomingPublish as publish callback. */
```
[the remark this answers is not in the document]
muneebahmed10: Not any more.

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+
+/* Standard includes. */
+#include <string.h>
+#include <stdio.h>
+#include <assert.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+#include "queue.h"
```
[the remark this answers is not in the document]
muneebahmed10: I don't think so

demos/coreMQTT_Agent/simple_sub_pub_demo.c
muneebahmed10: `MQTTAgent_Subscribe` will return success if the command is added to the queue. This will keep retrying until the command is added once. Yes, it's possible for the queue to become full here in the case when both the connection has failed and there are more tasks than the size of the queue.

The queue shouldn't really ever become full when the agent task is running since it has a higher priority then the subscribe-publish tasks.

demos/coreMQTT_Agent/simple_sub_pub_demo.c
muneebahmed10: There's already a comment block above this that explains that each subscribe is given a unique identifier for matching purposes

demos/coreMQTT_Agent/simple_sub_pub_demo.c
muneebahmed10: This may have been since it's not necessary for this specific demo since there's only one subscribe per task, but I updated this

demos/coreMQTT_Agent/simple_sub_pub_demo.c
```diff
@@ -0,0 +1,560 @@
...
+        memcpy( ( void * ) cTerminatedString, pxPublishInfo->pPayload, pxPublishInfo->payloadLength );
+        cTerminatedString[ pxPublishInfo->payloadLength ] = 0x00;
+    }
+    else
+    {
+        memcpy( ( void * ) cTerminatedString, pxPublishInfo->pPayload, mqttexampleSTRING_BUFFER_LENGTH );
+        cTerminatedString[ mqttexampleSTRING_BUFFER_LENGTH - 1 ] = 0x00;
+    }
+
+    LogInfo( ( "Received incoming publish message %s", cTerminatedString ) );
```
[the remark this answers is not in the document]
muneebahmed10: This is what it looks like on the console:
```
121 25364 [SubPub1] [INFO] Sending publish request to agent with message "Publisher1 publishing message 9" on topic "/filter/Publisher1"
122 25364 [SubPub1] [INFO] Task Publisher1 waiting for publish 9 to complete.
123 26365 [MQTT Agent] [INFO] Received incoming publish message Publisher1 publishing message 9
124 26365 [SubPub1] [INFO] Publish 9 completed successfully.
```

Including the topic doesn't give you any information that isn't already in the payload (i.e. that it's from Publisher1). And since these buffers aren't NULL terminated, I think adding another buffer just to print the topic uses up stack space unnecessarily.

demos/coreMQTT_Agent/simple_sub_pub_demo.c
aggarw13: What's the value of having blocking nature for the subscribe operation then if it leaves a door open for the task to be permanently blocked if the size of the queue is smaller than the number of tasks..
Should a single large blocking time be used instead of blocking for ever, and treat the unsuccessful queueing of operation as a demo error?

demos/coreMQTT_Agent/simple_sub_pub_demo.c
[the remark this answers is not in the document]
aggarw13: My main reason for suggesting logging of the incoming publish topic is to make it visually matching in the logs that the incoming publish topic matches the outgoing one.

demos/coreMQTT_Agent/mqtt_agent_task.c
```diff
@@ -956,44 +963,47 @@ static int prvConnectAndCreateDemoTasks( void * pvParameters )
     /* Set the pParams member of the network context with desired transport. */
     xNetworkContext.pParams = &secureSocketsTransportParams;
 
+    /* Initialize the MQTT context with the buffer and transport interface. */
+    xMQTTStatus = prvMQTTAgentInit();
```
aggarw13, line 967: Should there be an assert for the return value?

demos/coreMQTT_Agent/simple_sub_pub_demo.c
aggarw13: The comment does not explicitly mention that the critical section is used for protecting access to the global variable,

demos/coreMQTT_Agent/mqtt_agent_task.c
muneebahmed10: I think `if` checks are preferred over asserts here

demos/coreMQTT_Agent/simple_sub_pub_demo.c
muneebahmed10: I'll just remove the loop in that case

aggarw13 (approved):

dan4thewin (approved):

PR #3164, opened 2021-05-20
Minor fixes to C SDK HTTPS Client

iocoder:
This is Ramsès from Silicon Labs Inc. We were unable to build HTTPS Client because of the following issues:

- The source code of HTTPS Client includes a "C file" from the testing code, namely the access APIs C file ("iot_test_access_https_client.c"). Unfortunately this causes problems with our toolchain and automation tools, and it also doesn't align with our coding standards as we don't allow the inclusion of C files from other C files. This PR converts the file into a header.

- Missing simicolon after a Unity macro causes problems with the most recent version of unity. This PR adds the missing semicolon.

Thanks!


abhidixi11:
Hello Ramsès,

You don't have to compile the test code. You can set IOT_BUILD_TESTS to 0  and .c file will not be included. Our tests will not work if you change .c to .h. 

Have you tried our new coreHTTP library ? It does not have this problem.  Also please note, https library will not be available in main in a few days. ( but be still available though tag and old releases)
Thanks.

yourslab:
Hi @iocoder. Unfortunately, we won't be able to accept these changes because it will break our current test suite. If you'd like to compile the library without test code, you can set `IOT_BUILD_TESTS` to `0`.

PR #3184, opened 2021-06-01
[AFR] GreenGrass Discovery Demo Updated using coreMQTT Library.

gshvang:
This PR contains the updated Greengrass discovery demo using the coreMQTT library.

### DIFF
--- a/demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
+++ b/demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
@@ -23,15 +23,15 @@
  * http://www.FreeRTOS.org
  */
 
-/**
- * @file aws_greengrass_discovery_demo.c
- * @brief A simple Greengrass discovery example.
- *
- * A simple example that perform discovery of the greengrass core device.
- * The JSON file is retrieved.
- */
+ /**
+  * @file aws_greengrass_discovery_demo.c
+  * @brief A simple Greengrass discovery example.
+  *
+  * A simple example that perform discovery of the greengrass core device.
+  * The JSON file is retrieved.
+  */
 
-/* The config header is always included first. */
+  /* The config header is always included first. */
 #include "iot_config.h"
 
 /* Standard includes. */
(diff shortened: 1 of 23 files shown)


demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+    ulTimeMs = ( uint32_t ) ( ulTimeMs - ulGlobalEntryTimeMs );
 
-        /* Call the MQTT blocking PUBLISH function. */
-        xMqttStatus = IotMqtt_TimedPublish( mqttConnection,
-                                            &xPublishInfo,
-                                            0,
-                                            _maxCommandTimeMs );
+    return ulTimeMs;
+}
+static BaseType_t prvBackoffForRetry( BackoffAlgorithmContext_t * pxRetryParams )
```
yourslab, line 225: It looks like a good chunk of the functions here can just reference the code in https://github.com/aws/amazon-freertos/tree/main/demos/common/mqtt_demo_helpers

It might require slight modifications to the CMake but referencing common functions from one place will make it easier to make bug fixes in the future

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -223,23 +498,63 @@ static int _discoverGreengrassCore( const IotNetworkInterface_t * pNetworkInterf
...
         {
-            _sendMessageToGGC( mqttConnection );
+            _sendMessageToGGC( &xMQTTContext );
 
             IotLogInfo( "Disconnecting from broker." );
 
             /* Call MQTT v2's DISCONNECT function. */
-            IotMqtt_Disconnect( mqttConnection, 0 );
-            mqttConnection = IOT_MQTT_CONNECTION_INITIALIZER;
+            MQTT_Disconnect( &xMQTTContext );
```
yourslab, line 535: Might be a good idea to log a message if MQTT_Disconnect returned an error status

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+static void _sendMessageToGGC( MQTTContext_t * pxMQTTContext )
+{
+    BaseType_t xDemoStatus = pdFAIL;
+
+    /* Call the MQTT PUBLISH function. */
+    xDemoStatus = prvMQTTPublishToTopic( pxMQTTContext );
+
+    if( xDemoStatus != pdPASS )
+    {
+        IotLogError( "mqtt_client - Failure to publish." );
```
yourslab, line 470: Preferably we just use LogError function instead of IotLogError. Is there any reason why you decided to use IotLogError?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+    BaseType_t xStatus = pdPASS;
+    uint32_t ulMessageCounter;
+    char cBuffer[ ggdDEMO_MAX_MQTT_MSG_SIZE ];
+    const char * pcTopic = ggdDEMO_MQTT_MSG_TOPIC;
+
+    /* Some fields are not used by this demo so start with everything at 0. */
+    ( void ) memset( ( void * ) &xMQTTPublishInfo, 0x00, sizeof( xMQTTPublishInfo ) );
+
+    /* This demo uses QoS1. */
+    xMQTTPublishInfo.qos = MQTTQoS0;
```
yourslab, line 430: It says uses QoS1 but qos is set to 0?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+
+    if( xResult != MQTTSuccess )
+    {
+        IotLogError( ( "Failed to establish MQTT connection: Server=%s, MQTTStatus=%s",
+                       clientcredentialMQTT_BROKER_ENDPOINT, MQTT_Status_strerror( xResult ) ) );
+    }
+    else
+    {
+        /* Successfully established and MQTT connection with the broker. */
+        IotLogInfo( ( "An MQTT connection is established with %s.", clientcredentialMQTT_BROKER_ENDPOINT ) );
```
muneebahmed10, line 327: Isn't the MQTT connection established with the greengrass core and not the iot core broker?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+     * attempts are reached.
+     */
+    do
     {
-        if( connectAttempt > 0 )
+        /* Establish a TLS session with the MQTT broker. This example connects to
+         * the MQTT broker as specified in clientcredentialMQTT_BROKER_ENDPOINT and
+         * clientcredentialMQTT_BROKER_PORT at the top of this file. */
+        IotLogInfo( "Creating a TLS connection to %s:%u.",
+                    clientcredentialMQTT_BROKER_ENDPOINT, clientcredentialMQTT_BROKER_PORT );
```
muneebahmed10, line 378: The server config indicates it's creating a connection to `pxHostAddressData->pcHostAddress` instead. Also, shouldn't this be using `LogInfo` instead of `IotLogInfo`?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -223,23 +498,63 @@ static int _discoverGreengrassCore( const IotNetworkInterface_t * pNetworkInterf
...
+        }
+
+        if( xDemoStatus == pdPASS )
         {
-            _sendMessageToGGC( mqttConnection );
+            _sendMessageToGGC( &xMQTTContext );
 
             IotLogInfo( "Disconnecting from broker." );
 
             /* Call MQTT v2's DISCONNECT function. */
```
muneebahmed10, line 534: This comment seems out of date

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,129 +145,344 @@ typedef struct
...
+
+/** @brief Static buffer used to hold MQTT messages being sent and received. */
+static MQTTFixedBuffer_t xBuffer =
+{
+    ucSharedBuffer,
+    NETWORK_BUFFER_SIZE
+};
+
+static void _sendMessageToGGC( MQTTContext_t * pxMQTTContext );
+static int _discoverGreengrassCore();
```
muneebahmed10, line 191: If these are being rewritten anyway, we should remove the leading underscores

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -58,18 +58,78 @@
...
+/**
+ * @brief The base back-off delay (in milliseconds) to use for network operation retry
+ * attempts.
+ */
+#define RETRY_BACKOFF_BASE_MS                        ( 500U )
+
+/**
+ * @brief Timeout for receiving CONNACK packet in milliseconds.
+ */
+#define mqttexampleCONNACK_RECV_TIMEOUT_MS           ( 1000U )
```
muneebahmed10, line 127: The other macros either use `ggdDEMO_` or just omit a prefix, so that should probably be done here too since this isn't an MQTT demo

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -223,23 +498,63 @@ static int _discoverGreengrassCore( const IotNetworkInterface_t * pNetworkInterf
...
+             * demo execution, to clean up the system resources that it may have consumed. */
+            if( xIsConnectionEstablished == pdTRUE )
+            {
+                /* Close the network connection.  */
+                xNetworkStatus = SecureSocketsTransport_Disconnect( &xNetworkContext );
+
+                if( xNetworkStatus != TRANSPORT_SOCKET_STATUS_SUCCESS )
+                {
+                    xDemoStatus = pdFAIL;
+                    status = EXIT_FAILURE;
```
yourslab, line 547: Is there a need to have both a status and xDemoStatus? Doesn't seem like both are needed.

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
gshvang: Yeah the connection is with Greengrass Core, I will update the message.

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
gshvang: As discussed we can have it in the separate PR.

vendors/pc/boards/windows/aws_demos/config_files/aws_ggd_config.h
```diff
@@ -32,15 +32,39 @@
...
+ * 2. Define the LIBRARY_LOG_NAME and LIBRARY_LOG_LEVEL macros depending on
+ * the logging configuration for DEMO.
+ * 3. Include the header file "logging_stack.h", if logging is enabled for DEMO.
+ */
+
+#include "logging_levels.h"
+
+/* Logging configuration for the Demo. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME    "MQTT_MutualAuth_Demo"
```
muneebahmed10, line 50: Should be updated

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
muneebahmed10: Publish and disconnect at least can use the common code here

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,350 @@ typedef struct
...
+    xMQTTPublishInfo.pTopicName = pcTopic;
+    xMQTTPublishInfo.topicNameLength = ( uint16_t ) strlen( pcTopic );
+
+    for( ulMessageCounter = 0; ulMessageCounter < ( uint32_t ) ggdDEMO_MAX_MQTT_MESSAGES; ulMessageCounter++ )
+    {
+        xMQTTPublishInfo.pPayload = ( const void * ) cBuffer;
+        xMQTTPublishInfo.payloadLength = ( uint32_t ) sprintf( cBuffer, ggdDEMO_MQTT_MSG_DISCOVERY, ( long unsigned int ) ulMessageCounter ); /*lint !e586 sprintf can be used for specific demo. */
+
+        /* Get a unique packet id. */
+        usPublishPacketIdentifier = MQTT_GetPacketId( pxMQTTContext );
```
muneebahmed10, line 439: QoS 0 publishes don't use packet IDs

vendors/pc/boards/windows/aws_demos/config_files/aws_ggd_config.h
```diff
@@ -32,15 +32,39 @@
 #ifndef _AWS_GGD_CONFIG_H_
 #define _AWS_GGD_CONFIG_H_
 
+/**************************************************/
```
muneebahmed10, line 35: This should probably be changed for all the boards and not just Windows, right?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
gshvang: Logic in Disconnect helper function is based on some static variables which are set in mqttConnection helper function. So it won't work it in this demo. The publish helper function publishes QOS1 messages whereas in ggd demo we are showing QOS0, so it will unnecessarily complicate the demo. May be I can include QOS1 changes in the PR where I will be using the demo helper functions.

vendors/pc/boards/windows/aws_demos/config_files/aws_ggd_config.h
gshvang: Yeah

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+};
+
+static void sendMessageToGGC( MQTTContext_t * pxMQTTContext );
+static int discoverGreengrassCore();
+static void prvEventCallback( MQTTContext_t * pxMQTTContext,
+                              MQTTPacketInfo_t * pxPacketInfo,
+                              MQTTDeserializedInfo_t * pxDeserializedInfo );
+static uint32_t prvGetTimeMs( void );
+static BaseType_t prvBackoffForRetry( BackoffAlgorithmContext_t * pxRetryParams );
+static BaseType_t prvCreateMQTTConnectionWithBroker( GGD_HostAddressData_t * pxHostAddressData,
```
muneebahmed10, line 194: Document these functions

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+                                                     MQTTContext_t * pxMQTTContext,
+                                                     NetworkContext_t * pxNetworkContext );
 
 /*-----------------------------------------------------------*/
+static void prvEventCallback( MQTTContext_t * pxMQTTContext,
+                              MQTTPacketInfo_t * pxPacketInfo,
+                              MQTTDeserializedInfo_t * pxDeserializedInfo )
+{
+    return;
```
muneebahmed10, line 203: `( void )` the parameters if they're not used

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+    else
+    {
+        LogError( ( "Unable to retry operation with broker: Random number generation failed" ) );
+    }
+
+    return xReturnStatus;
 }
 
-/*-----------------------------------------------------------*/
+static BaseType_t prvCreateMQTTConnectionWithBroker( GGD_HostAddressData_t * pxHostAddressData,
```
muneebahmed10, line 267: Is this parameter only used for logging? Can the log just be moved to whatever functions calls this one?

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+    xSocketsConfig.recvTimeoutMs = TRANSPORT_SEND_RECV_TIMEOUT_MS;
+
+    /* Initialize reconnect attempts and interval. */
+    BackoffAlgorithm_InitializeParams( &xReconnectParams,
+                                       RETRY_BACKOFF_BASE_MS,
+                                       RETRY_MAX_BACKOFF_DELAY_MS,
+                                       RETRY_MAX_ATTEMPTS );
+
+    /* Attempt to connect to MQTT broker. If connection fails, retry after
+     * a timeout. Timeout value will exponentially increase till maximum
```
muneebahmed10, line 367:
```suggestion
     * a timeout. Timeout value will exponentially increase until maximum
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+     * attempts are reached.
+     */
+    do
     {
-        if( connectAttempt > 0 )
+        /* Establish a TLS session with the MQTT broker. This example connects to
+         * the MQTT broker as specified in pxHostAddressData->pcHostAddress and
```
muneebahmed10, line 373: We can be more explicit here for more helpful comments
```suggestion
         * the Greengrass Core as specified in pxHostAddressData->pcHostAddress and
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
-    xCredentials.pAlpnProtos = NULL;
-
-    /* Set the server info. */
+    ServerInfo_t xServerInfo = { 0 };
+
+    SocketsConfig_t xSocketsConfig = { 0 };
+    TransportSocketStatus_t xNetworkStatus = TRANSPORT_SOCKET_STATUS_SUCCESS;
+    BackoffAlgorithmContext_t xReconnectParams;
+    BaseType_t xBackoffStatus = pdFALSE;
+    BaseType_t xMqttStatus = pdFALSE;
```
muneebahmed10, line 342: Nit: If you're using these variables in a pass/fail scenario, initializing them as such would be more consistent
```suggestion
    BaseType_t xBackoffStatus = pdFAIL;
    BaseType_t xMqttStatus = pdFAIL;
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+                if( xNetworkStatus != TRANSPORT_SOCKET_STATUS_SUCCESS )
+                {
+                    LogError( ( "SecureSocketsTransport_Disconnect() failed to close the network connection. "
+                                "StatusCode=%d.", ( int ) xNetworkStatus ) );
+                }
+            }
         }
+    } while ( ( xNetworkStatus != TRANSPORT_SOCKET_STATUS_SUCCESS || xMqttStatus != pdPASS ) && ( xBackoffStatus == pdPASS ) );
+
+    return ( xNetworkStatus == TRANSPORT_SOCKET_STATUS_SUCCESS && xMqttStatus == pdPASS ) ? pdPASS : pdFAIL;
```
muneebahmed10, line 412: The first clause is unnecessary since the second implies the former, i.e. `xNetworkStatus == TRANSPORT_SOCKET_STATUS_SUCCESS` whenever `xMqttStatus == pdPASS`.
```suggestion
    } while ( ( xMqttStatus != pdPASS ) && ( xBackoffStatus == pdPASS ) );

    return xMqttStatus;
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+ * between the current time and the global entry time. This will reduce the chances
+ * of overflow for the 32 bit unsigned integer used for holding the timestamp.
+ */
+static uint32_t ulGlobalEntryTimeMs;
+
+/**
+ * @brief Packet Identifier generated when Publish request was sent to the broker;
+ * it is used to match received Publish ACK to the transmitted Publish packet.
+ */
+static uint16_t usPublishPacketIdentifier;
```
muneebahmed10, line 165: Why have this for QoS 0?
```suggestion
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+        xMQTTPublishInfo.pPayload = ( const void * ) cBuffer;
+        xMQTTPublishInfo.payloadLength = ( uint32_t ) sprintf( cBuffer, ggdDEMO_MQTT_MSG_DISCOVERY, ( long unsigned int ) ulMessageCounter ); /*lint !e586 sprintf can be used for specific demo. */
 
-        IotLogInfo( "Attempting to establish MQTT connection to Greengrass." );
-        xMqttStatus = IotMqtt_Connect( &xNetworkInfo,
-                                       &xMqttConnectInfo,
-                                       _maxCommandTimeMs,
-                                       pMqttConnection );
+        /* Send PUBLISH packet. Packet ID is not used for a QoS1 publish. */
+        xResult = MQTT_Publish( pxMQTTContext, &xMQTTPublishInfo, usPublishPacketIdentifier );
```
muneebahmed10, line 439: This isn't QoS 1. Also there's no point in having the variable if it's not used
```suggestion
        /* Send PUBLISH packet. Packet ID is not used for a QoS0 publish. */
        xResult = MQTT_Publish( pxMQTTContext, &xMQTTPublishInfo, 0 );
```

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,347 @@ typedef struct
...
+    return xStatus;
 }
 
-/*-----------------------------------------------------------*/
+static void sendMessageToGGC( MQTTContext_t * pxMQTTContext )
+{
+    BaseType_t xDemoStatus = pdFAIL;
+
+    /* Call the MQTT PUBLISH function. */
+    xDemoStatus = prvMQTTPublishToTopic( pxMQTTContext );
```
muneebahmed10, line 461: Why not just rename `prvMQTTPublishToTopic` to `sendMessageToGGC`? This function as it is doesn't do anything except call another function.

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -222,24 +492,68 @@ static int _discoverGreengrassCore( const IotNetworkInterface_t * pNetworkInterf
...
+            xMQTTStatus = MQTT_Disconnect( &xMQTTContext );
+
+            if( xMQTTStatus != MQTTSuccess )
+            {
+                LogError( ( "Failed to DISCONNECT MQTT connection: MQTTStatus=%s", MQTT_Status_strerror( xMQTTStatus ) ) );
+            }
+
+            /* We will always close the network connection, even if an error may have occurred during
+             * demo execution, to clean up the system resources that it may have consumed. */
+            if( xIsConnectionEstablished == pdTRUE )
```
muneebahmed10, line 539: This statement is always true if this line is reached

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -222,24 +492,68 @@ static int _discoverGreengrassCore( const IotNetworkInterface_t * pNetworkInterf
...
+            }
+
+            LogInfo( ( "Disconnected from the broker." ) );
         }
         else
         {
-            IotLogError( "Could not connect to the Broker. Error %s.",
-                         IotMqtt_strerror( mqttStatus ) );
+            LogError( ( "Could not connect to the Broker." ) );
             status = EXIT_FAILURE;
```
muneebahmed10, line 557: This looks like a duplicate since lines 519-520 do the same thing?

muneebahmed10 (commented):
Won't [CMakeLists.txt](https://github.com/aws/amazon-freertos/blob/main/demos/greengrass_connectivity/CMakeLists.txt#L17) need to be updated to use coreMQTT as well?

gshvang:
> Won't [CMakeLists.txt](https://github.com/aws/amazon-freertos/blob/main/demos/greengrass_connectivity/CMakeLists.txt#L17) need to be updated to use coreMQTT as well?

Updated

yourslab:
Since CMakeLists are updated, won't we have to generate all the project files again? In which case, shouldn't we just use the helper functions where possible?

gshvang:
> Since CMakeLists are updated, won't we have to generate all the project files again? In which case, shouldn't we just use the helper functions where possible?

As we have discussed, some modifications needs to be made to the helper functions as well to use them. Planning to make it part of separate PR.

muneebahmed10 (commented):
> Since CMakeLists are updated, won't we have to generate all the project files again? In which case, shouldn't we just use the helper functions where possible?

Just ran the update script and the project files didn't change, likely since the coreMQTT files were included in the projects anyway from the other demos

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -85,134 +142,404 @@ typedef struct
     char * pcTopic;                     /**< Topic to subscribe and publish to. */
 } GGDUserData_t;
 
+/**
+ * @brief Stub for Network Context.
```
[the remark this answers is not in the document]
gshvang: Updated.

gshvang:
> > Since CMakeLists are updated, won't we have to generate all the project files again? In which case, shouldn't we just use the helper functions where possible?
> 
> Just ran the update script and the project files didn't change, likely since the coreMQTT files were included in the projects anyway from the other demos

Yeah, ble and coremqtt demos already include coreMQTT dependencies as well.

muneebahmed10 (approved):

yourslab (approved):

demos/greengrass_connectivity/aws_greengrass_discovery_demo.c
```diff
@@ -71,6 +89,45 @@
...
+
+/**
+ * @brief Milliseconds per FreeRTOS tick.
+ */
+#define MILLISECONDS_PER_TICK                  ( MILLISECONDS_PER_SECOND / configTICK_RATE_HZ )
+
+/**
+ * @brief The maximum number of retries for network operation with server.
+ */
+#define RETRY_MAX_ATTEMPTS                     ( 5U )
```
gkwicker, line 107: Why is 5 the right number?

vendors/cypress/boards/CY8CKIT_064S0S2_4343W/aws_demos/config_files/aws_ggd_config.h
```diff
@@ -1,6 +1,6 @@
 /*
- * Amazon FreeRTOS V1.4.8
- * Copyright (C) 2019 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ * FreeRTOS V1.4.8
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
gkwicker, line 3: Should this be 2021?

gkwicker (approved):
Approved but see comments

PR #3192, opened 2021-06-03
Add Espressif port for OTAv3

divekarshubham:
Add Espressing port for OTAv3

Description
-----------

* Adding back ota_pal.c to implement the Platform Interface functions required for OTAv3 with changes to function signatures and updated data types.
* Adding esp helper functions
* Adding configuration files and updating cmake to build components for OTA

Checklist:
----------

- [x] My code is Linted.


mahavirj:
@divekarshubham @pvyawaha 

- What is the baseline that you used for this port? Is it some older version before v4.2 integration PR?
- We had substantial amount of cleanups. E.g. if you see file https://github.com/aws/amazon-freertos/blob/30e14afb403d78732388461f14f1d4195e72e4ad/vendors/espressif/boards/ports/ota/aws_esp_ota_ops.c, it has very minimal additional code. But if we see same file in this PR then it is adding back lot of changes which are not really required.

May you please rebase your changes on top of latest ESP OTA PAL? That way we can understand exact differences and it will be easier to review and maintain as well.

divekarshubham:
> @divekarshubham @pvyawaha
> 
>     * What is the baseline that you used for this port? Is it some older version before v4.2 integration PR?
> 
>     * We had substantial amount of cleanups. E.g. if you see file https://github.com/aws/amazon-freertos/blob/30e14afb403d78732388461f14f1d4195e72e4ad/vendors/espressif/boards/ports/ota/aws_esp_ota_ops.c, it has very minimal additional code. But if we see same file in this PR then it is adding back lot of changes which are not really required.
> 
> 
> May you please rebase your changes on top of latest ESP OTA PAL? That way we can understand exact differences and it will be easier to review and maintain as well.

Thanks for pointing this out, I have pulled in the latest changes.

vendors/espressif/boards/esp32/aws_demos/config_files/core_mqtt_agent_config.h
```diff
@@ -0,0 +1,71 @@
...
+ * will be.
+ */
+#define MQTT_AGENT_MAX_SUBSCRIPTION_FILTER_LENGTH ( 100 )
+
+/**
+ * @brief Dimensions the buffer used to serialise and deserialise MQTT packets.
+ * @note Specified in bytes.  Must be large enough to hold the maximum
+ * anticipated MQTT payload.
+ */
+#define MQTT_AGENT_NETWORK_BUFFER_SIZE    ( 5000 )
```
pvyawaha, line 53: With default block size of 4K , this will not be enough for OTA over HTTP demo

vendors/espressif/boards/esp32/aws_demos/config_files/ota_demo_config.h
```diff
@@ -0,0 +1,67 @@
...
+#endif
+
+/**
+ * @brief Build version of the firmware.
+ *
+ *        This is used in the OTA demo to set the appFirmwareVersion variable
+ *        that is declared in the ota_appversion32.h file in the OTA library.
+ */
+#ifndef APP_VERSION_BUILD
+    #define APP_VERSION_BUILD    0
```
pvyawaha, line 64: We are giving default version in all documents 0.9.2 why changing to 0.9.0 ?

divekarshubham: Reverted to 0.9.2 in https://github.com/aws/amazon-freertos/pull/3192/commits/a485c919afafb552fcbc037c6d5dd26c19b85c2d

vendors/espressif/boards/esp32/aws_demos/config_files/core_mqtt_agent_config.h
divekarshubham: From End-to-End testing this buffer size seems sufficient for the OTA-HTTP demo.

pvyawaha (dismissed):

n9wxu (approved):

pvyawaha (approved):

PR #3237, opened 2021-06-15
Add missing .h files to the OTA CMake target

yanjos-dev:
Add missing .h files to the OTA CMake target

Description
-----------
The OTA CMake target is defined based on various CMake variables in the OTA library. These variables are missing some of the .h header files. This does not cause any issues with the build process, but it does cause issues generating metadata that is ingested by some tools. This PR adds the missing header files directly to the CMake target for the OTA library to resolve this.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/ota_demo_dependencies.cmake
+++ b/libraries/ota_demo_dependencies.cmake
@@ -23,9 +23,18 @@ afr_module_sources(
     ${AFR_CURRENT_MODULE}
     PRIVATE
         ${OTA_SOURCES}
-        ${OTA_OS_FREERTOS_SOURCES} 
+        # Include missing OTA source .h files that are required to generate correct metadata.
+        "${CMAKE_CURRENT_LIST_DIR}/ota_for_aws/source/include/ota_appversion32.h"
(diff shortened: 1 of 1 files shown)


libraries/ota_demo_dependencies.cmake
```diff
@@ -1,31 +1,36 @@
...
-              LIST_DIRECTORIES false
-              ${ota_public_include_dir}/*.h )
-    list(APPEND OTA_HEADER_FILES ${ota_public_include_header_files})
-endforeach()
-# Remove the hardcoded coreJSON and TinyCBOR dependancies and add 
-# them manually later to enable them being used by other applications.
+
+# Remove the coreJSON and TinyCBOR files from the OTA_SOURCES and
+# OTA_INCLUDE_PRIVATE_DIR variables. This is so that the OTA target can depend
+# on the local copy of the libraries instead of the ones in the OTA repository.
```
divekarshubham, line 8:
```suggestion
# on the libraries part of amazon-freertos instead of the ones nested in the OTA repository.
```

divekarshubham (approved):

dan4thewin (approved):

PR #3246, opened 2021-06-18
Fix Renesas OTA PAL error return statements

yanjos-dev:
Fix Renesas OTA PAL return statements

Description
-----------
The Renesas OTA PAL were returning a sub error instead of the combined error in multiple of the functions. This meant that the functions were effectively returning the correct error, but shifted over. This change updates the return statements to return a complete error instead of just the sub error. This change also uncrustifies the code, so it will look like a large change.

To review this change, please view the specific commits that are unrelated to the uncrustify commit.

Note: After this PR, all of the OTA PAL qualification tests are passing.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/vendors/renesas/boards/rx65n-rsk/ports/ota_pal_for_aws/ota_pal.c
+++ b/vendors/renesas/boards/rx65n-rsk/ports/ota_pal_for_aws/ota_pal.c
@@ -99,120 +99,121 @@ static uint8_t * otaPal_ReadAndAssumeCertificate( const uint8_t * const pucCertN
 
 /*------------------------------------------ firmware update configuration (start) --------------------------------------------*/
 /* R_FLASH_Write() arguments: specify "low address" and process to "high address" */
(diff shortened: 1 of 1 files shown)


divekarshubham (commented):
`otaPal_CheckFileSignature` has inconsistent returns
This should return all `OtaPalMainStatus_t` and the function signature should change.
Else you can return `OtaPalStatus_t` and use the following to extract main err:
`eResult = OTA_PAL_MAIN_ERR( otaPal_CheckFileSignature( pFileContext ));`

NIT: there is a macro `R_OTA_ERR_NONE`, should we change it to reflect latest pal statuses?

yanjos-dev:
Thanks for noticing that, I have corrected that function. We discussed offline, but for others reference, the R_OTA_ERR_NONE error is the platform specific error. It's the sub error owned by the vendor so I'll leave that alone for now.

divekarshubham (dismissed):

aggarg (dismissed):

divekarshubham (approved):

cobusve (approved):

PR #3257, opened 2021-06-21
Esp32s2 SameSocketDifferentTasks fix

AniruddhaKanhere:
Description
-----------
The test `AFQP_SECURE_SOCKETS_Threadsafe_SameSocketDifferentTasks` expects the return value of the recv function to be always greater than 0. The secure sockets API sometimes does return a 0 value which should not be treated as an error.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
+++ b/libraries/abstractions/secure_sockets/test/iot_test_tcp.c
@@ -2745,10 +2745,13 @@ static void prvSOCKETS_Threadsafe_SameSocketDifferentTasks( Server_t xConn )
                 configPRINTF( ( "Received %d bytes. Total %d\r\n", xReturned, xTotalReceived ) );
                 TEST_ASSERT_GREATER_OR_EQUAL_MESSAGE( 0, xReturned, "Error occurred receiving large message" );
 
-                /* Data was received. */
-                TEST_ASSERT_EQUAL_MEMORY( &cTransmittedString[ xTotalReceived ], pcReceivedString, xReturned );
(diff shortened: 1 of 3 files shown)


alfred2g (dismissed):

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
```diff
@@ -2742,12 +2742,15 @@ static void prvSOCKETS_Threadsafe_SameSocketDifferentTasks( Server_t xConn )
             {
                 xReturned = SOCKETS_Recv( ( Socket_t ) xSocket, ( char * ) pcReceivedString, xRecvLen, 0 );
 
-                TEST_ASSERT_GREATER_THAN_MESSAGE( 0, xReturned, "Error occurred receiving large message" );
+                TEST_ASSERT_GREATER_OR_EQUAL_MESSAGE( 0, xReturned, "Error occurred receiving large message" );
```
gkwicker, line 2745: If `xReturned < 0` it would be good to log the value of `xReturned` (you might be able to use `tcptestFAILUREPRINTF()` for this) before this line to aid in debugging – I

gkwicker (changes requested):
See comment

libraries/abstractions/secure_sockets/test/iot_test_tcp.c
AniruddhaKanhere: Yes, thank you.
I added the requested line. :)

gkwicker (approved):

PR #3267, opened 2021-06-23
Marvell: fix WiFi reconnect issue

AniruddhaKanhere:
Description
-----------
The PR https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/62 modified the DHCP state machine to accept two parameters to avoid race conditions.
In the case of Marvell, the old code was not modified. To try and reset DHCP state machine, it sent an event to IP-task with `eInitialWait` as the payload. But, this does not work with the new code since DHCP state machine is at some other state and the request to switch to `eInitialWait` state is ignored.

This PR fixes that. Instead of sending an event to IP-task, we should directly reset the DHCP state machine by passing a `pdTRUE` as a param to the `vDHCPProcess`.

Link to the test [here](https://amazon-freertos-ci.corp.amazon.com/job/freertos_refinteg/job/custom_job_pipeline/491).

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/marvell/WMSDK/mw320/sdk/src/platform/net/freertos-plus-tcp/net.c
+++ b/vendors/marvell/WMSDK/mw320/sdk/src/platform/net/freertos-plus-tcp/net.c
@@ -142,7 +142,9 @@ int net_configure_address(struct wlan_ip_config *addr, void *intrfc_handle)
 		wlan_wlcmgr_send_msg(WIFI_EVENT_NET_STA_ADDR_CONFIG,
 			        WIFI_EVENT_REASON_SUCCESS, NULL);
 		*ipLOCAL_IP_ADDRESS_POINTER = 0x00UL;
-		/* The network is not up until DHCP has completed. */
+		/* The network is not up until DHCP has completed. Reset the
+		 * DHCP state machine to start the process of acquiring an IP
+		 * address. */
 		vDHCPProcess( pdTRUE, eInitialWait );
 
 		while (*ipLOCAL_IP_ADDRESS_POINTER == 0)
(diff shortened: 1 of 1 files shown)


vendors/marvell/WMSDK/mw320/sdk/src/platform/net/freertos-plus-tcp/net.c
```diff
@@ -141,8 +142,10 @@ int net_configure_address(struct wlan_ip_config *addr, void *intrfc_handle)
 		wlan_wlcmgr_send_msg(WIFI_EVENT_NET_STA_ADDR_CONFIG,
 			        WIFI_EVENT_REASON_SUCCESS, NULL);
 		*ipLOCAL_IP_ADDRESS_POINTER = 0x00UL;
-		/* The network is not up until DHCP has completed. */
-		xSendEventToIPTask( eDHCPEvent );
+		/* The network is not up until DHCP has completed. Reset the
+		 * DHCP state machine to start the process of acquiring an IP
+		 * address. */
+		vDHCPProcess( pdTRUE, eInitialWait );
```
aggarw13, line 148: A comment about the significance of the `pdTRUE` parameter may be valuable.

aggarg (dismissed):

vendors/marvell/WMSDK/mw320/sdk/src/platform/net/freertos-plus-tcp/net.c
AniruddhaKanhere: Done.

aggarg (approved):

abhidixi11 (approved):

PR #3270, opened 2021-06-24
Overflow bug in windows aws_demo_logging.c

pvyawaha:
Fix bug which was causing memory corruption in aws_demo_logging.c 

Description
-----------

The vsnprintf returns the number of characters that would have been written if n had been sufficiently large and not actual number of characters . This was causing overflow and memory corruption.

This is blocking issue for OTA release.

Checklist:
----------

- [ ] My code is Linted.

### DIFF
--- a/vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
@@ -537,7 +537,7 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
         }
 
         /* Check if the buffer has room for "\r\n" and make it if not. */
-        if (xLength > ( dlMAX_PRINT_STRING_LENGTH - 3 ) )
+        if ( xLength > ( dlMAX_PRINT_STRING_LENGTH - 3 ) )
         {
             xLength -= 3;
         }
--- a/vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
+++ b/vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
@@ -527,13 +527,27 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
         }
         else if ( xLength2 < ( dlMAX_PRINT_STRING_LENGTH - xLength ) )
         {
-            /* Buffer required is larger than allocated, not complete log is in buffer. */
-            xLength += ( dlMAX_PRINT_STRING_LENGTH - xLength );
+            /* Complete log is in buffer. */
+            xLength += xLength2;
         }
         else
         {
-            /* Complete log is in buffer. */
-            xLength += xLength2;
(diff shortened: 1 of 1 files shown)


vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -525,8 +525,22 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
...
+        }
+        else
+        {
+            /* Buffer required is larger than allocated, not complete log is in buffer. */
+            xLength = dlMAX_PRINT_STRING_LENGTH ;
+        }
 
-        xLength += xLength2;
+        /* Check if the buffer has room for "\r\n" and make it if not. */
+        if (xLength > ( dlMAX_PRINT_STRING_LENGTH - 3 ) )
```
aggarw13, line 540: (Nit) formatting seems off. Has uncrustify been run on this file?
```suggestion
        if ( xLength > ( dlMAX_PRINT_STRING_LENGTH - 3 ) )
```

vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
```diff
@@ -525,8 +525,22 @@ static void prvLoggingPrintf( uint8_t usLoggingLevel,
...
+        {
+            xLength -= 3;
```
aggarw13, line 542: As a safety measure to prevent underflow if the `dlMAX_PRINT_STRING_LENGTH` is less than 3 characters in length (which is unlikely), the check for `dlMAX_PRINT_STRING_LENGTH` OR `xLength` >= 3 can be added

AniruddhaKanhere, line 542: Should this be `xLength = dlMAX_PRINT_STRING_LENGTH - 3;`? Considering that xLength can be much bigger.

divekarshubham (dismissed):

vendors/pc/boards/windows/aws_demos/application_code/aws_demo_logging.c
pvyawaha: I dont think this is required and it will not fail here first, if its not called from task then the string itself is 4 bytes and some other prefixes before the actual log, I can put a comment that minimum 128 is required , what do you think ?

pvyawaha: The prefix  practically can not be that long but its possible , particularly if someone reduces the buffer. Thats good point.

pvyawaha: Thats good point, although practically it should not happen but its possible.

divekarshubham (approved):

AniruddhaKanhere (approved):
@pvyawaha, thank you for the changes.

n9wxu (approved):

PR #3394, opened 2021-09-28
Pr esp common io uart fixes

dpfrey:
Espressif common io UART fixes

Description
-----------

Checklist:
----------
- [x*] I have tested my changes. No regression in existing tests.
- [ ] My code is Linted.

\* I wrote some code that uses the espressif common io UART, but no automated tests were run.

### DIFF
--- a/vendors/espressif/boards/ports/common_io/iot_uart.c
+++ b/vendors/espressif/boards/ports/common_io/iot_uart.c
@@ -301,7 +301,6 @@ int32_t iot_uart_ioctl(IotUARTHandle_t const pxUartPeripheral, IotUARTIoctlReque
             case eUartSetConfig : {
                 int32_t uart_port_num = uart_ctx->uart_port_num;
                 IotUARTConfig_t *iot_uart_config = (IotUARTConfig_t *) pvBuffer;
-                memcpy(&uart_ctx->iot_uart_conf, iot_uart_config, sizeof(IotUARTConfig_t));
                 uart_config_t uart_config = {0};
                 uart_config.baud_rate = iot_uart_config->ulBaudrate;
                 if (iot_uart_config->ucWordlength == 5) {
@@ -352,7 +351,13 @@ int32_t iot_uart_ioctl(IotUARTHandle_t const pxUartPeripheral, IotUARTIoctlReque
                 ret |= iot_uart_driver_install(uart_port_num, uart_config);
                 //Create a callback function to handle UART event from ISR
                 ret |= uart_register_callback_with_isr(uart_port_num, uart_event_cb, (void *)iot_uart_handler);
-                return (ret == ESP_OK) ? IOT_UART_SUCCESS : IOT_UART_INVALID_VALUE;
+
+                if (ret == ESP_OK) {
+                    memcpy(&uart_ctx->iot_uart_conf, iot_uart_config, sizeof(IotUARTConfig_t));
+                    return IOT_UART_SUCCESS;
+                } else {
+                    return IOT_UART_INVALID_VALUE;
+                }
             }
             case eUartGetConfig : {
                 IotUARTConfig_t *iot_uart_config = (IotUARTConfig_t *) pvBuffer;
--- a/vendors/espressif/boards/ports/common_io/iot_uart.c
+++ b/vendors/espressif/boards/ports/common_io/iot_uart.c
@@ -302,7 +302,7 @@ int32_t iot_uart_ioctl(IotUARTHandle_t const pxUartPeripheral, IotUARTIoctlReque
                 int32_t uart_port_num = uart_ctx->uart_port_num;
                 IotUARTConfig_t *iot_uart_config = (IotUARTConfig_t *) pvBuffer;
                 memcpy(&uart_ctx->iot_uart_conf, iot_uart_config, sizeof(IotUARTConfig_t));
-                uart_config_t uart_config;
+                uart_config_t uart_config = {0};
                 uart_config.baud_rate = iot_uart_config->ulBaudrate;
                 uart_config.data_bits = iot_uart_config->ucWordlength;
                 if (iot_uart_config->ucFlowControl == true) {
--- a/vendors/espressif/boards/ports/common_io/iot_uart.c
+++ b/vendors/espressif/boards/ports/common_io/iot_uart.c
@@ -318,7 +318,9 @@ int32_t iot_uart_ioctl(IotUARTHandle_t const pxUartPeripheral, IotUARTIoctlReque
 
                 if (iot_uart_config->ucFlowControl == true) {
                     uart_config.flow_ctrl = UART_HW_FLOWCTRL_CTS_RTS;
-                    uart_config.rx_flow_ctrl_thresh = UART_FIFO_LEN;
+                    uart_config.rx_flow_ctrl_thresh = (UART_FIFO_LEN - 1);
+                } else {
+                    uart_config.flow_ctrl = UART_HW_FLOWCTRL_DISABLE;
                 }
                 switch (iot_uart_config->xParity) {
                     case eUartParityNone:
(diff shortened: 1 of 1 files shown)


shubhamkulkarni97:
Hi @dpfrey, Thanks for the PR.

Could you describe the issue you were facing without these changes?
It would be great if we could reproduce the issue and verify the fix at our end.

dpfrey:
> Could you describe the issue you were facing without these changes? It would be great if we could reproduce the issue and verify the fix at our end.

I have some application code that looks like this:
```
    IotUARTConfig_t uart_config;
    int32_t status;

    s_console_uart = iot_uart_open(UART_NUM_0);
    configASSERT(s_console_uart);

    status = iot_uart_ioctl(s_console_uart, eUartGetConfig, &uart_config);
    configASSERT(status == IOT_UART_SUCCESS);

    uart_config.ulBaudrate = 115200;
    uart_config.ucWordlength = 8;
    uart_config.xParity = eUartParityNone;
    uart_config.xStopbits = eUartStopBitsOne;
    uart_config.ucFlowControl = false;

    status = iot_uart_ioctl(s_console_uart, eUartSetConfig, &uart_config);
    configASSERT(status == IOT_UART_SUCCESS);
```

When I ran the software, I saw this on my console:
```
I (321E (328) uart: uart_param_config(654): hw_flowctrl mode error
```
This message was generated from `vendors/espressif/esp-idf/components/driver/uart.c` in the `uart_param_config` function. The problem was that if flow control is disabled, `uart_config.flow_ctrl` is not set and the value of that field is thus undefined.

This is fixed by: 5a780d2bfd53bc254fa0d44817a532b9f81b3408 and d44ac548c53f8c28c18972eba8316721af7400e7

For a little while, my main application was setting `uart_config.ucFlowControl = true` which was actually incorrect for my board, but I noticed `"rx flow thresh error"` on my console. Again, this message was coming from `uart.c`. The reason for this message is described in detail in the commit message for `d44ac548c53f8c28c18972eba8316721af7400e7`.

I had fixed the issues above, but my uart was filled with garbage. I then realized that the word length handling was incorrect.  See 593b718dd30a0d12224fb2f1b40cf688b1937f7b for the fix.

There is one more fundamental issue that is *not* addressed by this PR.  The `eUartGetConfig`, modify, `eUartSetConfig` pattern doesn't really work the way that it should because the `eUartGetConfig` ioctl gives back a cached value that has never been set before. This value is from `uart_ctx->iot_uart_conf`.  Notice that this value will have never been written unless the `eUartSetConfig` ioctl had been called previously.

shubhamkulkarni97:
@dpfrey We are observing test failures and crashes with these changes. We will check for the issues and get back.

shubhamkulkarni97:
@dpfrey Can you apply attached patch on your branch? It should fix all the issues in common-io tests.

Please run `git am common_io_uart_fixes.patch` to apply the patch.

[common_io_uart_fixes.patch.zip](https://github.com/aws/amazon-freertos/files/7283267/common_io_uart_fixes.patch.zip)

dpfrey:
@shubhamkulkarni97: I am confused by this patch.  In `vendors/espressif/boards/ports/common_io/iot_uart.c` in the `iot_uart_ioctl` function for the `eUartSetConfig` case, you return `IOT_UART_INVALID_VALUE` in the case that the user requested flow control. Why did you do that?

Also, even if you don't want to support flow control, the logic is still subtly wrong. It should be like this because the field is a `uint8_t` (not a `bool`):
```
if (if (iot_uart_config->ucFlowControl != 0) {
    return IOT_UART_INVALID_VALUE;
} else {
    uart_config.flow_ctrl = UART_HW_FLOWCTRL_DISABLE;
}
```
The version in the patch treats any value other than `1` as a request to disable flow control, but any value other than `0` or `1` is invalid.

shubhamkulkarni97:
> I am confused by this patch. In vendors/espressif/boards/ports/common_io/iot_uart.c in the iot_uart_ioctl function for the eUartSetConfig case, you return IOT_UART_INVALID_VALUE in the case that the user requested flow control. Why did you do that?

Flow control is not supported on ESP devkits. This is a hardware limitation as CTS and RTS pins are not connected.
Application hangs when flow control is enabled.

> Also, even if you don't want to support flow control, the logic is still subtly wrong. It should be like this because the field is a uint8_t (not a bool)

I agree with this comment, you may add this change to your branch

dpfrey:
@shubhamkulkarni97: It is true that on my ESP32 dev kit (Espressif pico kit v4.1), the RTS and CTS pins of the onboard CP2102N USB-to-UART adapter chip are not connected to the ESP32. However, I believe that common_io UART is attempting to be a general UART abstraction for all uses on the platform, not just for the console port of the dev kit.

As an example, I set this up:
|    | UART1 | UART2 |
| --- | --- | --- |
| TXD | GPIO4 | GPIO12 |
| RXD | GPIO2 | GPIO14 |
| RTS | GPIO15 | GPIO27 |
| CTS | GPIO13 | GPIO33 |

I modified `struct esp_uart_pin_config` in `vendors/espressif/boards/ports/common_io/include/iot_board_gpio.h` to have a spot for an RTS and CTS pin specification and I modified `ESP_UART_PIN_MAP` to match my table above. I also had to make a change to `vendors/espressif/boards/ports/common_io/iot_uart.c` to get `iot_uart_driver_install` to actually use the RTS and CTS pins I had specified.

Once I set this up, I started a timer and every few seconds, I called `iot_uart_write_async(s_uart_1, ...)` and then `iot_uart_read_async(s_uart2, ...)` and I found that I was able to send and receive the data as expected.

It seems a bit clumsy to have to modify `ESP_UART_PIN_MAP` in `vendors/espressif/boards/ports/common_io/include/iot_board_gpio.h` in the FreeRTOS tree based on my specific board setup and application. Is there another way that this could be done to give applications control over choosing which pins they use for their UARTs?

shubhamkulkarni97:
Hi @dpfrey Sorry for delayed response.

> It is true that on my ESP32 dev kit (Espressif pico kit v4.1), the RTS and CTS pins of the onboard CP2102N USB-to-UART adapter chip are not connected to the ESP32. However, I believe that common_io UART is attempting to be a general UART abstraction for all uses on the platform, not just for the console port of the dev kit.

I agree with you that common_io UART should support flow control as it is abstraction for all use cases. Most probably the flow control specific issue is related to wrong use of common-io API.

> I modified struct esp_uart_pin_config in vendors/espressif/boards/ports/common_io/include/iot_board_gpio.h to have a spot for an RTS and CTS pin specification and I modified ESP_UART_PIN_MAP to match my table above. I also had to make a change to vendors/espressif/boards/ports/common_io/iot_uart.c to get iot_uart_driver_install to actually use the RTS and CTS pins I had specified
> It seems a bit clumsy to have to modify ESP_UART_PIN_MAP in vendors/espressif/boards/ports/common_io/include/iot_board_gpio.h in the FreeRTOS tree based on my specific board setup and application. Is there another way that this could be done to give applications control over choosing which pins they use for their UARTs?

This change is specific to the common_io architecture design. I think AFR technical team might be in a better position to comment on pin mappings for RTS and CTS.

> Based on my understanding, GPIO_NUM_20 doesn't correspond to a pin that is actually routed outside the module.

Thanks for catching this. I think this pin was changed when adding common_io support for ESP32-S2. We will take a look and update pin map to support ESP32 and ESP32-S2.


@aggarw13 @mingyue86010 Could you please help to answer above questions related to supporting CTS and RTS in common-io pin map?

Also, https://github.com/aws/amazon-freertos/commit/441e02e3776833b436c9d69859d28a01bd29fa84 commit by @aggarw13 adds iot_uart_init() call in app_main. This functions initialises UART_0 in flow control mode. Could you please elaborate the use case of this function and requirement to enable flow control? This function call causes issues because it re-initializes UART_0 which is initialised as IDF console by default.

AniruddhaKanhere:
Hello @dpfrey, Thanks a lot for reporting and bringing this to our attention.
Thanks @shubhamkulkarni97 for incorporating the changes that @dpfrey suggested into your PR.

Now, since the PR by @shubhamkulkarni97 is merged into mainline, I think we can close this PR.

Thanks again - I'll close this PR now.

Issue #1646, opened 2019-12-18
SOCKETS_SO_WAKEUP_CALLBACK memory leak in secure_sockets


dougbenedict:
**Describe the bug**
The SOCKETS_SO_WAKEUP_CALLBACK processing for lwip creates a background thread which performs a select() on the socket and invokes the callback routine when there is activity on the socket.

Setting the SOCKETS_SO_WAKEUP_CALLBACK with a NULL callback parameter calls the routine  prvRxSelectClear() which is empty except for a /* TODO */ comment. This orphans the select thread rather than deleting it and cleaning up. Open and close enough sockets using the SOCKETS_SO_WAKEUP_CALLBACK option and the kernel heap will run out of memory.

In the freertos_plus_tcp iot_secure_sockets.c implementation, the SOCKETS_SO_WAKEUP_CALLBACK option is not handled at all.

**System information**
- Operating System Windows
- Discovered in 201910.00 but I see the code is the same in the current master
- Project Custom Application

**Expected behavior**
The spawned select task should be deleted and the resources reclaimed when the SOCKETS_SO_WAKEUP_CALLBACK option is called with a NULL callback. And the lwip and freertos_plus_tcp implementations should be consistent.

**To reproduce**

Using the lwip network stack.

1. Create a TCP socket using SOCKETS_socket():
socket = SOCKETS_Socket(SOCKETS_AF_INET, SOCKETS_SOCK_STREAM, SOCKETS_IPPROTO_TCP);
2. Install an activity callback using the SOCKETS_SO_WAKEUP_CALLBACK option:
 SOCKETS_SetSockOpt(socket, 0, SOCKETS_SO_WAKEUP_CALLBACK, &tcp_socket_callback, sizeof(&cy_tcp_socket_callback));
3. Use the TCP socket normally.
4. When done with the TCP socket, remove the callback:
  SOCKETS_SetSockOpt(socket, 0, SOCKETS_SO_WAKEUP_CALLBACK, NULL, 0);
5. Close the socket:
  SOCKETS_Close(socket);

The select task is now orphaned and the resources are not recovered.

htibosch:
I felt like responding to this post, until I saw that it is using lwIP.

However, the macro `SOCKETS_SO_WAKEUP_CALLBACK` seems to borrowed from FreeRTOS+TCP, in which it is defined with the same value:

~~~c
    #define FREERTOS_SO_WAKEUP_CALLBACK	( 17 )
~~~

[...]

PR #3406, opened 2021-10-15
Add method to remove callback from secure sockets implimentation

AniruddhaKanhere:
Description
-----------
In reference to https://github.com/aws/amazon-freertos/issues/1646, this PR fixes the callback removal issue which could cause an orphan task to be created.
Cleaning a callback requires the spawned task to be terminated before everything can be cleaned up. This requires two way communication/synchronization. Thus, event groups were used to allow both tasks to communicate with each other of the actions required.
Also, the `lwip_select` function was made to block for 10 seconds instead of infinite time - this would introduce overheads but was necessary. The time interval of 10 seconds can be changed by modifying the provided macro. 

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
+++ b/libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
@@ -73,13 +73,13 @@
 
 /*-----------------------------------------------------------*/
 
-#define SS_STATUS_CONNECTED    ( 1 )
-#define SS_STATUS_SECURED      ( 2 )
+#define SS_STATUS_CONNECTED               ( 1 )
+#define SS_STATUS_SECURED                 ( 2 )
 
 #define SECURE_SOCKETS_SELECT_WAIT_SEC    ( 10 )
 
 #define SOCKETS_START_DELETION            ( 0x01 )
-#define SOCKETS_COMPLETE_DELETION            ( 0x02 )
+#define SOCKETS_COMPLETE_DELETION         ( 0x02 )
 
 /*
  * secure socket context.
@@ -102,7 +102,7 @@ typedef struct _ss_ctx_t
     int recv_flag;
 
     TaskHandle_t rx_handle;
-    void ( * rx_callback )( Socket_t pxSocket );
+    void ( *rx_callback )( Socket_t pxSocket );
     EventGroupHandle_t rx_EventGroup;
 
     bool enforce_tls;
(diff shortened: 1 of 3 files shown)


libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -319,6 +339,12 @@ static void prvRxSelectSet( ss_ctx_t * ctx,
     ctx->rx_callback = ( void ( * )( Socket_t ) )pvOptionValue;
 
     prvIncrementRefCount( ctx );
+
+    ctx->rx_EventGroup = NULL;
```
dan4thewin, line 343: Does the next line make the last unnecessary?

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
```diff
@@ -336,7 +362,28 @@ static void prvRxSelectSet( ss_ctx_t * ctx,
...
+    /* Wait for the task to delete itself. */
+    if( xEventGroupWaitBits( ctx->rx_EventGroup, SOCKETS_COMPLETE_DELETION, pdTRUE, pdTRUE, portMAX_DELAY ) == SOCKETS_COMPLETE_DELETION )
+    {
+        /* The task 'vTaskRxSelect' has started deleting itself. */
+    }
+    else
+    {
+        /* Control should never reach here as this task should wait on the event group bits indefinately. */
+        configASSERT( pdTRUE );
+    }
```
dan4thewin, line 377: I think this is an assumption.  I think indefinite blocking requires INCLUDE_vTaskSuspend.  If a customer configures that to be 0, then the function could return due to timeout.  Perhaps wrap this in a while loop?

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
AniruddhaKanhere: Yes, I missed that. Thanks, fixed it.

libraries/abstractions/secure_sockets/lwip/iot_secure_sockets.c
AniruddhaKanhere: Ah yes! I overlooked that. I have changed that to a while. Thanks for catching that Dan.

AniruddhaKanhere (commented):
Thanks for the review Dan. I have applied your suggestions.

dan4thewin (approved):

Issue #3417, opened 2021-10-28
[BUG] <BLE advertising name can't be increased to more than 5 characters >


sumit612:
**System information**
- Hardware board: [ Custom Board ESP32 WROOM]
- Operating System: [ Windows 10 ]
- Code version: 202012.00
- Project/Demo: [ CONFIG_OTA_UPDATE_DEMO_ENABLED with custom functionalities added ]

**Describe the bug**
We were trying to increase BLE advertising name length. We tried the below process to increase the name lenth, but it can't be increased to more the 5 characters.
      1. change IOT_BLE_DEVICE_SHORT_LOCAL_NAME_SIZE  more than 5(e.g 6) in "iot_ble_config_defaults.h"
      2. update IOT_BLE_DEVICE_COMPLETE_LOCAL_NAME according to the name length(e.g "123456")
	  
If we try to increase more than 5 characters the OTA demo stops and it's log screenshot attached below.

![5character_issue](https://user-images.githubusercontent.com/11428929/139211027-54de3bce-de53-4168-8c71-0bfcfeee17c1.PNG)

Is there a way to increase BLE advertising name to more than 5 characters?

dachalco:
Hi @sumit612 

Removed the bug label as it's a valid failure per the [BT/BLE v4.2 spec](https://www.bluetooth.com/specifications/specs/core-specification-4-2/) which our BT/BLE software is based on. For the MQTT BLE demo, and I believe for the BLE GATT DEMO too, our advertisement data looks like this when _shortened local name_ is set to 6:
```
$2 = {
  flags = 6 '\006', 
  uuids16 = 0x0, 
  num_uuids16 = 0 '\000', 
  uuids16_is_complete = 0, 
  uuids32 = 0x0, 
  num_uuids32 = 0 '\000', 
  uuids32_is_complete = 0, 
  uuids128 = 0x3ffb3d70, 
  num_uuids128 = 1 '\001', 
  uuids128_is_complete = 1, 
  name = 0x3ffbdc40 <ble_svc_gap_name> "123456", 
  name_len = 6 '\006', 
  name_is_complete = 0, 
  tx_pwr_lvl = -128 '\200', 
  tx_pwr_lvl_is_present = 1, 
  slave_itvl_range = 0x0, 
  svc_data_uuid16 = 0x0, 
  svc_data_uuid16_len = 0 '\000', 
  public_tgt_addr = 0x0, 
  num_public_tgt_addrs = 0 '\000', 
[...]

PR #3420, opened 2021-11-03
Add logging to indicate when exceeding max ADV PDU data size

dachalco:
Description
-----------
Add logging when user attempts setting BT/BLE advertising data over the max ADV PDU data size, for our stack (BT/BLE v4.2). This mitigates issues like [this](https://github.com/aws/amazon-freertos/issues/3417). 

Also increases default length for esp32 ble demos, so that if fully displays default device name in advertisement, instead of truncating it as it was.

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/libraries/ble/src/iot_ble_gap.c
+++ b/libraries/ble/src/iot_ble_gap.c
@@ -652,6 +652,7 @@ BTStatus_t IotBle_On( void )
             if( status == eBTStatusSuccess )
             {
                 status = prvSetAdvData( &_scanRespParams );
+
                 if( status == eBTStatusNoMem )
                 {
                     IotLogError( "Exceeded max advertisement PDU data size. Please reduce advertised data." );
--- a/libraries/ble/src/iot_ble_gap.c
+++ b/libraries/ble/src/iot_ble_gap.c
@@ -652,6 +652,14 @@ BTStatus_t IotBle_On( void )
             if( status == eBTStatusSuccess )
             {
                 status = prvSetAdvData( &_scanRespParams );
+                if( status == eBTStatusNoMem )
+                {
+                    IotLogError( "Exceeded max advertisement PDU data size. Please reduce advertised data." );
+                }
+            }
+            else if( status == eBTStatusNoMem )
+            {
+                IotLogError( "Exceeded max advertisement PDU data size. Please reduce advertised data." );
             }
         }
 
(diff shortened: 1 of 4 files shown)


paulbartell (approved):

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gap.c
```diff
@@ -743,9 +743,14 @@ BTStatus_t prvBTSetAdvData( uint8_t ucAdapterIf,
...
-    if( xESPStatus != 0 )
+    switch( xESPStatus )
     {
-        xStatus = eBTStatusFail;
+        case BLE_HS_EMSGSIZE:
+            xStatus = eBTStatusNoMem;
+            break;
+
+        default:
+            xStatus = eBTStatusFail;
```
paulbartell, line 753: missing a break statement here.

paulbartell (changes requested):

vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_internals.h
```diff
@@ -140,6 +140,7 @@ static inline const BTStatus_t BTNRFError( const ret_code_t xErrCode )
             break;
 
         case NRF_ERROR_NO_MEM:
```
paulbartell, line 142: Suggest adding a comment stating that this fall-though is intentional.

vendors/espressif/boards/ports/ble/nimble/iot_ble_hal_gap.c
dachalco: it's the last case, so break is extraneous

vendors/nordic/boards/nrf52840-dk/ports/ble/iot_ble_hal_internals.h
dachalco: Comment may seem redundant, alternative is to catch in its own case, but they both map to same BT API error status.

libraries/ble/src/iot_ble_gap.c
```diff
@@ -652,6 +652,15 @@ BTStatus_t IotBle_On( void )
             if( status == eBTStatusSuccess )
             {
                 status = prvSetAdvData( &_scanRespParams );
+
```
ravibhagavandas, line 655: The log may seem misleading here as `eBTStatusNoMem` can be emitted due to other reasons as well. Could you update the log accordingly.

PR #3473, opened 2022-03-25
Update freertos_kernel submodule to align with TF-M v1.5.0

xinyu-tfm:
Signed-off-by: Xinyu Zhang <xinyu.zhang@arm.com>
Change-Id: I7c55be2dc68386eb05d62b311b457b0a1ac6c24f

Description
-----------
The [PR ](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/469)to update FreeRTOS-Kernel to integrate with TF-M v1.5.0 has been merged. This PR is to have the submodule updated. 


paulbartell (approved):

kstribrnAmzn:
Looks like this submodule update is causing some build checks to fail. Looking into why now.

kstribrnAmzn:
Looks like this submodule update is causing some build checks to fail. Looking into why now.

kstribrnAmzn:
This PR brings in code starting from September 2021. It's the other changes (unfortunately) which are causing build failures. We'll need to fix the build failures before merging in.

xinyu-tfm:
Thanks very much for your investigation! 😄

PR #3528, opened 2022-08-30
Support for dynamic credential switching

KazukiMochizuki926:
Description
-----------
Supports dynamic credential switching.
Fleet Provisioning uses the claim certificate and private key to connect to AWS the first time, and then uses the device-specific claim certificate and private key to connect to AWS next time.
However, the current FreeRTOS library cannot support Fleet Provisioning because it uses fixed credential information.
Therefore, the FreeRTOS library source code has been updated to allow dynamic credential switching.
The IDT test items FullSecureSockets, FullTLS, and FulMQTT have passed.

idt_1
idt_2
idt_3

Checklist:
----------

- [x] My code is Linted.

### DIFF
--- a/demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
+++ b/demos/common/mqtt_demo_helpers/mqtt_demo_helpers.c
@@ -62,6 +62,9 @@
 /* Include the secure sockets implementation of the transport interface. */
 #include "transport_secure_sockets.h"
 
+/* Include header for client certificates and client private key label. */
+#include "core_pkcs11_config.h"
+
 /*-----------------------------------------------------------*/
 
 /**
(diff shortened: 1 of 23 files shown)


jasonpcarroll:
Hi @KazukiMochizuki926,
Thank you for your contribution! I've started the CI checks and will review.
Best.
Jason Carroll

cobusve:
Hey all, I took a closer look a this one and we need to take a bit of change of direction here. I think making all these modules dependent on the list of credentials leaks this change into too many places. 

I think we should change this to pass the credentials into all the modules and have the list in a single place, that will not require all of these modules to be updated with this dependency which will cause all kinds of problems later on.

Also can we please move this effort to the FreeRTOS/FreeRTOS location where the new code is located?

